PR #283, opened 2016-01-22
Use SoapAction value to derive if it is a WSTrust 1.3 or 2005 endpoint.

kpanwar:
Use SoapAction value to derive if it is a WSTrust 1.3 or 2005 endpoint.

### DIFF
--- a/src/ADAL.Common/MexParser.cs
+++ b/src/ADAL.Common/MexParser.cs
@@ -43,7 +43,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     {
         public WsTrustVersion Version { get; set; }
 
-        public string Id { get; set; }   
+        public string Id { get; set; }
 
         public UserAuthType AuthType { get; set; }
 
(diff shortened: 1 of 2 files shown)


src/ADAL.Common/MexParser.cs
```diff
@@ -96,18 +98,25 @@ internal static WsTrustAddress ExtractWsTrustAddressFromMex(XDocument mexDocumen
                 SetPolicyEndpointAddresses(mexDocument, bindings);
                 Random random = new Random();
                 //try ws-trust 1.3 first
-                policy = policies.Values.Where(p => p.Url != null && p.AuthType == userAuthType && p.Version == WsTrustVersion.WsTrust13).OrderBy(p => random.Next()).FirstOrDefault() ??
-                         policies.Values.Where(p => p.Url != null && p.AuthType == userAuthType).OrderBy(p => random.Next()).FirstOrDefault();
+                policy =
+                    policies.Values.Where(
```
RPangrle: You've got two statements here chained by the '??' operator. Can you put each into its own variable on its own line (or lines) and then have a policy = a ?? b statement instead? I think that'd make this a bit easier to follow.

RPangrle:
:shipit:

PR #304, opened 2016-03-02
Fixes a couple doc bugs for UserCredential and UserAssertion

psignoret:


psignoret:
.

src/ADAL.NET/UserCredential.cs
```diff
@@ -35,7 +35,7 @@ public UserCredential(string userName, string password)
         }
 
         /// <summary>
-        /// Constructor to create credential with client id and secret
+        /// Constructor to create credential with with username and password
```
kpanwar: double with.

psignoret:
Fixed, thanks.

psignoret:
Fixed, thanks.

Issue #321, opened 2016-03-23
Add logging implementation for iOS/Android - v3


kpanwar:
Add logging implementation for iOS/Android - v3

PR #328, opened 2016-03-24
Add logging and callback support for all platforms

kpanwar:
Add logging and callback support for all platforms per Issue #321

### DIFF
--- a/src/ADAL.PCL/IAdalLogCallback.cs
+++ b/src/ADAL.PCL/IAdalLogCallback.cs
@@ -1,0 +1,56 @@
+﻿using System;
+using System.Threading;
+
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory
+{
+    /// <summary>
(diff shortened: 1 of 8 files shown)


src/ADAL.PCL/IAdalLogCallback.cs
```diff
@@ -0,0 +1,56 @@
...
+        Warning,
+        Error
+    }
+
+    /// <summary>
+    /// Callback for capturing ADAL logs to custom logging schemes.
+    /// </summary>
+    public interface IAdalLogCallback
+    {
+        void Log(LogLevel level, string message);
```
RPangrle: Seeing how this is a new API we should have something for PII vs non-PII

kpanwar: This is adal v3. We can discuss Pii for msal in slack.

RPangrle:
:shipit:

Issue #307, opened 2016-03-07
Multi threaded cache load issue v2


kpanwar:
Multi threaded cache load issue v2

PR #333, opened 2016-03-26
Fix cache race condition in v2

abhishek58g:
Commit for issue #307


tests/Test.ADAL.Common/TokenCacheTests.cs
```diff
@@ -665,5 +665,33 @@ public static AuthenticationResult GenerateRandomCacheValue(int maxFieldSize, st
                     UserInfo = new UserInfo { UniqueId = uniqueId, DisplayableId = displayableId }
                 };
         }
+
+        public static void ParallelStorePositiveTest(byte[] oldcache)
+        {
+            for (int i = 0; i < 100; i++)
```
kpanwar: no need for 100 times.

abhishek58g: Removed the loop

kpanwar:
you should put lock around all the methods. This will reduce bugs in long run.

abhishek58g:
Added locks for all methods except private ones

kpanwar:
I think you should consider putting a lock around all methods.

abhishek58g:
Added locks around all the methods

PR #336, opened 2016-03-29
Fix v3 build issue in jenkins

kpanwar:
Fix v3 build issue in jenkins. This required reverting to Profile7
because we cannot support certificate based cryptography anyways.


weijjia:
:shipit:

Issue #329, opened 2016-03-24
clean up documentation and headers for v3


kpanwar:
clean up documentation and headers for v3

kpanwar:
change the header to be MIT license. If the header does not exist then add it to the .cs file.

kpanwar:
@abhishek58g close this issue if you are done with the work.

abhishek58g:
Closed it.

PR #342, opened 2016-03-30
Code Cleaning :- Adding Appropriate headers #329

abhishek58g:
Code Cleaning :- Adding Appropriate headers #329


kpanwar:
tests/TestApp/XFormsApp.iOS/SecondPageRenderer.cs is missing -------------------- on the top of the header.. I am sure other files do too. Can you check and confirm?

PR #344, opened 2016-04-01
Add username/password flow for Desktop only

kpanwar:
Add username/password flow for Desktop only

### DIFF
--- a/src/ADAL.PCL.Desktop/ADAL.PCL.Desktop.csproj
+++ b/src/ADAL.PCL.Desktop/ADAL.PCL.Desktop.csproj
@@ -58,7 +58,7 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
-	<Compile Include="..\ADAL.Common\AuthenticationContextIntegratedAuthExtensions.cs">
+    <Compile Include="..\ADAL.Common\AuthenticationContextIntegratedAuthExtensions.cs">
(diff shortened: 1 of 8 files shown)


weijjia:
:shipit:

forivall:
Is there any information on why this was removed? is there any way to provide completely native fields when for the login, or are we now forced to use a webview when logging in on Xamarin?

kpanwar:
Because native apps should not be collecting user credentials anyways. It is a poor practice and the only reason it exists in desktop is for legacy support at the moment.

kpanwar:
I would highly recommend you to use webview in xamarin.

forivall:
Use of a webview collects user credentials as well, no matter what, you're going to need to trust the native application. It's trivial to collect user credentials off of a webview. Preventing headless login without release notes explaining how to migrate only encourages even less secure workarounds or reliance on older versions. (Which is exactly what I've been forced to do)

kpanwar:
Just because it is trivial it does not mean that poor practices should be served on a platter.

forivall:
Then serve good practices on a silver platter

kpanwar:
We do serve good practices. For example, we use a webview that cannot be overridden, allow HTTPS only URLs and provide signed binaries, to name a few. Malicious apps can always do whatever they want to steal credentials but the library itself tries to protect against that. :)

If you look at the next generation library MSAL that speaks to newer v2 endpoint (serving aad and msa authentication from a single source), we don't even support username password flow. I hope this information helps you make right decision for your apps.

Issue #345, opened 2016-04-05
Fix iOS ExpireOn claim conversion from Broker - V3


kpanwar:
Xamarin iOS ignored the expires_on claim returned from the broker and used UTC.Now + expires_in to compute expiration time. This resulted in current time as expiration time for the access token.

kpanwar:
fixed.

PR #346, opened 2016-04-05
Fix bug to consume expires_on claim from iOS broker for Issue #345

kpanwar:
Xamarin iOS ignored the expires_on claim returned from the broker and used UTC.Now + expires_in to compute expiration time. This resulted in current time as expiration time for the access token.


kpanwar:
Xamarin iOS ignored the expires_on claim returned from the broker and used UTC.Now + expires_in to compute expiration time. This resulted in current time as expiration time for the access token.

weijjia:
:shipit:

PR #347, opened 2016-04-07
fix for android and documentation update

kpanwar:
fix for android and documentation update


src/ADAL.PCL.iOS/PlatformParameters.cs
```diff
@@ -44,11 +44,20 @@ private PlatformParameters()
...
         }
 
+        /// <summary>
+        /// Additional parameters used in acquiring user's authorization
+        /// </summary>
+        /// <param name="callerViewController">UIViewController instance</param>
+        /// <param name="skipBroker">skips calling to broker if broker is present. true, by default</param>
         public PlatformParameters(UIViewController callerViewController, bool skipBroker):this(callerViewController)
         {
             SkipBroker = skipBroker;
```
weijjia: non related, but adal already change it useBroker, and the default value is false.

PR #349, opened 2016-04-07
Add support for AcquireToken with client id, client secret, and user credential

espentveit:
In order to use PowerBI via Azure using non-interactive authentication a new AquireToken function signature was needed.

Fixes the error: 'AADSTS90014: The request body must contain the following parameter: 'client_secret or client_assertion' when trying to authenticate against a PowerBI web application.


espentveit:
Issue pull request against servicing branch instead

kpanwar:
If you are a web app then you should NOT be collecting user credentials to acquire tokens on the service side. This is a not a secure and prescribed way of authentication and support can be revoked any time on the service.

PR #350, opened 2016-04-07
Add support for AcquireToken with client id, client secret, and user credential

espentveit:
In order to use PowerBI via Azure using non-interactive authentication a new AquireToken function signature was needed.

Fixes the error: 'AADSTS90014: The request body must contain the following parameter: 'client_secret or client_assertion' when trying to authenticate against a PowerBI web application.

### DIFF
--- a/src/ADAL.Common/AcquireTokenNonInteractiveHandler.cs
+++ b/src/ADAL.Common/AcquireTokenNonInteractiveHandler.cs
@@ -25,6 +25,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     internal class AcquireTokenNonInteractiveHandler : AcquireTokenHandlerBase
     {
         private readonly UserCredential userCredential;
+        private readonly string clientSecret;
         private UserAssertion userAssertion;
         
         public AcquireTokenNonInteractiveHandler(Authenticator authenticator, TokenCache tokenCache, string resource, string clientId, UserCredential userCredential, bool callSync)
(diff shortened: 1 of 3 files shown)


kpanwar:
If you are a web app then you should NOT be collecting user credentials to acquire tokens on the service side. This is a not a secure and prescribed way of authentication and support can be revoked any time on the service. This pull request will not be accepted.

espentveit:
I understand the concern. This was a use case for getting a listing from the PowerBI API using a console app, without having to provide interactive auth every time (scheduled task). 
First I tried adding an native app through the PowerBI 'Add application' web site, but that didn't seem to work. Afterwards I tried adding it as an web application, and that did work. Details below. 

No collection of user credentials intended.

**Native app**
A native application was created through the PowerBI web site using an admin user.

The application did not show up in Azure AD Applications in Azure portal, but the PowerBI 'Add Application' web site did provide a clientId, so I tried using that with the user that was used to add the application.

The function AcquireToken(resource, clientId, userCredential) gave this error:
"Microsoft.IdentityModel.Clients.ActiveDirectory.AdalServiceException:
AADSTS65001: The user or administrator has not consented to use the application with ID 'appID'. Send an interactive authorization request for this user and resource."

Seemed like a interactive session was needed. Downloading and logging in as admin using the clientID provided from the PowerBI site via the example app at https://github.com/Microsoft/PowerBI-CSharp/tree/master/samples/webforms it would request the rights needed. The rights were granted by the user. Still didn't work in a non-interactive fashion (gave same error as above), and it seemed like the rights granted wasn't persisted.

**Web application**
Adding an an web application using the PowerBI Add Application interface worked, and it showed up in AD Applications in the Azure portal.

Tried using clientId and clientSecret. 
AcquireToken(resource, clientCredential)
The AcquireToken returned a token, but the API gave errors when I tried using it to consume the API [3].

I then thought maybe the clientId with username and password would work.

Tried getting a token by using the user that added the application:
The function:
AcquireToken(resource, clientId, userCredential)
Returned:
AADSTS90014: The request body must contain the following parameter: 'client_secret or client_assertion'

Which made me look for an overload of the function that allowed client_secret to be added, since the authentication endpoint said it was missing. There was none, so I implemented one based on that the first reply on a stackoverflow post [1] seemed to indicate it would work.

Google searches seemed to indicate that other have this problem as well. 

See first comment where this is proposed as a solution:
[1] http://stackoverflow.com/questions/32341877/cant-get-client-credentials-access-token-to-authorize-power-bi

[2] http://www.it1me.com/it-answers?id=33922731&ttl=Get+accessToken+from+power+BI+without+GUI+authentication

See follow up comments from JustLogic regarding a user probably being needed to consume PowerBI API (a appID/appSecret pair alone would not provide access to listings).

[3] http://stackoverflow.com/questions/35692974/how-to-connect-to-power-bi-api-using-non-interactive-authentication

kpanwar:
As discussed before, this is not a scenario that the library supports or plans on supporting. AAD Service will turn off this surface area as well.

PR #355, opened 2016-04-13
Allow WebUi to be overriden

crmann1:
Allow a consumer of ADAL v3 to register their own WebUI.


kpanwar:
We do not allow the developer to pass in their own webUI. All the classes are kept internal by design.

kpanwar:
Leave a comment

Issue #358, opened 2016-04-14
apply ADAL v3 code analysis and warning as error build updates


kpanwar:
ADAL v3 code analysis and warning as error updates

PR #361, opened 2016-04-15
build updates and pkeyauth cleanup

kpanwar:
build updates and pkeyauth cleanup for #358 

This PR contains turning on warnings as errors, using code analysis based on SDL 7.0 ruleset, Fixes for the new errors. There is also a bug fix for pkeyauth.


src/ADAL.PCL.iOS/AdalInitializer.cs
```diff
@@ -33,7 +33,10 @@
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 {
-    public static class AdalInitializer
+    /// <summary>
+    ///
```
weijjia: where is the summary?

weijjia:
:shipit:

PR #370, opened 2016-04-19
Update authority for Germany

brandwe:


kpanwar:
can you send a PR from a branch within this repo instead of your fork? there should only be 1 file changed instead of 128 files

PR #377, opened 2016-04-26
ADAL v2 :- CredScan CleanUp

abhishek58g:
ADAL v2 :- CredScan CleanUp


kpanwar:
why is the PFX file being deleted? it should be in the exclusion list. right?

PR #378, opened 2016-04-26
Adal v3 :- CredScan Clean up

abhishek58g:
Adal v3 :- CredScan Clean up


kpanwar:
Why are you deleting this file? shouldn't it be in the exclusion list for adal v3 build?

PR #389, opened 2016-05-05
Check assertion Hash when there are multiple tokens found - v3

kpanwar:
For the server side scenarios, ADAL should check user assertion Hash (if provided) when there are multiple tokens found because the developer did not supply upn. #384


RPangrle:
:shipit:

Issue #385, opened 2016-05-04
ADAL fails to initialize on Android when fill linking is enabled


mhutch:
When using ADAL with Xamarin.Android, if full linking is enabled, then ADAL fails with the following exception:

System.ArgumentNullException: Value cannot be null.
Parameter name: type
  at System.Activator.CreateInstance (System.Type type, Boolean nonPublic) [0x00006] in <filename unknown>:0 
  at System.Activator.CreateInstance (System.Type type) [0x00000] in <filename unknown>:0 
  at Microsoft.IdentityModel.Clients.ActiveDirectory.PlatformPlugin.InitializeByAssemblyDynamicLinking () [0x00011] in <filename unknown>:0 
  at Microsoft.IdentityModel.Clients.ActiveDirectory.PlatformPlugin..cctor () [0x00011] in <filename unknown>:0 

The problem appears to be that ADAL is dynamically loading types using reflection, and the linker does not recognize this pattern so removes them.

Linker attributes could be used to to ensure the dynamically loaded types are preserved: https://developer.xamarin.com/guides/android/advanced_topics/linking/

See also https://bugzilla.xamarin.com/show_bug.cgi?id=40811#c2

kpanwar:
fixed.

marty-wang:
which version contains the fix? because I still see the very issue in v3.12.0.

kpanwar:
I just tested and the library works as expected. Can you try it in a new sample project? I am not sure what your setup is like.

PR #391, opened 2016-05-05
Add Preseve tag for dynamically loaded classes #385

kpanwar:
When using ADAL with Xamarin.Android, if full linking is enabled, then
ADAL fails with the following exception:

System.ArgumentNullException: Value cannot be null.
Parameter name: type
at System.Activator.CreateInstance (System.Type type, Boolean nonPublic)
[0x00006] in :0
at System.Activator.CreateInstance (System.Type type) [0x00000] in :0
at
Microsoft.IdentityModel.Clients.ActiveDirectory.PlatformPlugin.InitializeByAssemblyDynamicLinking
() [0x00011] in :0
at Microsoft.IdentityModel.Clients.ActiveDirectory.PlatformPlugin..cctor
() [0x00011] in :0

The problem appears to be that ADAL is dynamically loading types using
reflection, and the linker does not recognize this pattern so removes
them.

Linker attributes could be used to to ensure the dynamically loaded
types are preserved:
https://developer.xamarin.com/guides/android/advanced_topics/linking/

See also https://bugzilla.xamarin.com/show_bug.cgi?id=40811#c2

### DIFF
--- a/src/ADAL.PCL.Android/BrokerHelper.cs
+++ b/src/ADAL.PCL.Android/BrokerHelper.cs
@@ -36,6 +36,7 @@ using Java.IO;
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 {
+    [Android.Runtime.Preserve(AllMembers = true)]
     class BrokerHelper : IBrokerHelper
     {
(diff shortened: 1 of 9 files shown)


weijjia:
:shipit:

Issue #392, opened 2016-05-05
Enable resource owner grant flow for ADFS in v2


kpanwar:
Enable resource owner grant flow for ADFS in v2

kpanwar:
fixed.

PR #393, opened 2016-05-05
Enable resource owner grant flow for ADFS in v2 #392

kpanwar:
Enable resource owner grant flow for ADFS in v2 #392


weijjia:
:shipit:

Issue #387, opened 2016-05-05
ADAL should send haschrome=1 QP for all interactive requests


kpanwar:
ADAL should send haschrome=1 QP for all interactive requests. This will hide the back button the UI and developers will stop getting access_denied error when it was actually user_cancel.

kpanwar:
fixed.

PR #396, opened 2016-05-06
v2 - add haschrome=1 QP to all interactive requests #387

kpanwar:
ADAL should send haschrome=1 QP for all interactive requests. This will
hide the back button the UI and developers will stop getting
access_denied error when it was actually user_cancel.


weijjia:
:shipit:

Issue #402, opened 2016-05-09
enable refresh_token grant for ADFS in v2


kpanwar:
enable refresh_token grant for ADFS in ADAL v2

kpanwar:
fixed.

kpanwar:
fixed

PR #403, opened 2016-05-09
enable refresh_token grant for ADFS in v2. #402

kpanwar:
enable refresh_token grant for ADFS in v2. Remove the check where ADAL failed when authority was not ADFS.


weijjia:
:shipit:

Issue #405, opened 2016-05-10
Fix array resize issue during Url encoding - v2


kpanwar:
Fix array resize issue during Url encoding where Array.Resize call was not being provided new size of the array and developer got index out of bound exception.

kpanwar:
fixed

PR #407, opened 2016-05-10
Fix array resize issue during Url encoding where Array.Resize call wa…

kpanwar:
…s not being provided new size of the array and developer got index out of bound exception. CopyTo function was also being called at the wrong place. the code should resize the array (if needed) and then copy to. #405


weijjia:
:shipit:

Issue #406, opened 2016-05-10
Fix array resize issue during Url encoding - v3


kpanwar:
Fix array resize issue during Url encoding where Array.Resize call was not being provided new size of the array and developer got index out of bound exception.

PR #408, opened 2016-05-11
Fix array resize issue during Url encoding

kpanwar:
Fix array resize issue during Url encoding where Array.Resize call was
not being provided new size of the array and developer got index out of
bound exception. #406


weijjia:
:shipit:

Issue #412, opened 2016-05-16
DictionaryRequestParameters bug in recent versions?


smourier:
In ADAL older versions (Experimental), the DictionaryRequestParameters used to override the ToString() method. Now it doesn't any more which causes this code to build an incorrect auth url:

            var url = authContext.GetAuthorizationRequestUrlAsync(scope, clientId, etc..) or
            var result = authContext.AcquireTokenAsync(scope, clientId, etc..)

The build url now has this form:
https://login.microsoftonline.com/common/oauth2/authorize?Microsoft.IdentityModel.Clients.ActiveDirectory.DictionaryRequestParameters

which is of course 100% invalid.

kpanwar:
Fixed the issue in 3.10.305161347

PR #414, opened 2016-05-16
Added missing ToString() in GetAuthorizationUrl

kpanwar:
Added missing ToString() that caused regressed in GetAuthorizationUrl #412


weijjia:
:shipit:

PR #418, opened 2016-05-20
correct handling of id_token if profile_info is not available

Amberg:
Correct handling for missing profile_info
https://azure.microsoft.com/en-us/documentation/articles/active-directory-v2-preview-oidc-changes/


kpanwar:
@Amberg - this branch is EOL and we do not support it. You should consider switching to MSAL, our next gen library that will be supported and will GA. It is available at  https://www.nuget.org/packages/Microsoft.Identity.Client

PR #425, opened 2016-05-23
Removing Creds from sts

abhishek58g:
Removing Creds from sts


kpanwar:
PR are sent to Convergence-v4. not release

PR #430, opened 2016-05-23
Sijun errmessage

lovemaths:


src/ADAL.PCL/Constants.cs
```diff
@@ -388,7 +388,7 @@ internal static class AdalErrorMessage
             +
             "1. The PromptBehavior.Never flag was passed, but the constraint could not be honored, because user interaction was required. "
             +
-            "2. An error occurred during a silent web authentication that prevented the http authentication flow from completing in a short enough time frame";
+            "2. An error occurred during a silent web authentication, which blocks the http authentication flow from completing";
```
kpanwar: this should be past tense, we also need to let the developer know what it was tried for a short time frame. I think the original message is good enough.

lovemaths: OK. Just changed it back to the original one.

PR #442, opened 2016-06-01
Xamarin.Mac backend (dev branch)

mhutch:
Adds a Xamarin.Mac backend and test app for it.


src/ADAL.PCL.Mac/AuthenticationAgentNSWindowController.cs
```diff
@@ -0,0 +1,284 @@
...
+        {
+            if (request == null)
+            {
+                WebView.DecideUse(decisionToken);
+                return;
+            }
+
+            string requestUrlString = request.Url.ToString();
+
+            if (requestUrlString.StartsWith(BrokerConstants.BrowserExtPrefix, StringComparison.OrdinalIgnoreCase))
```
kpanwar: launching browser in Mac for device enrollment is not  a supported scenario. native objc library returns error with "server is redirecting us to browser, this behavior is not defined on Mac OS X yet" message

src/ADAL.PCL.Mac/PlatformParameters.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// Skips calling to broker if broker is present. false, by default
+        /// </summary>
+        public bool UseBroker { get; set; }
+
+        /// <summary>
+        /// Shows the prompt in a modal dialog.
+        /// </summary>
+        public bool UseModalDialog { get; set; }
```
kpanwar: native library does not provide an option to use modal dialog. AFAIK, it forces modal by default.

src/ADAL.PCL.Mac/PlatformParameters.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// Caller NSWindow
+        /// </summary>
+        public NSWindow CallerWindow { get; private set; }
+
+        /// <summary>
+        /// Skips calling to broker if broker is present. false, by default
+        /// </summary>
+        public bool UseBroker { get; set; }
```
kpanwar: Broker is not supported at all. We should not expose this flag.

src/ADAL.PCL.Mac/PlatformParameters.cs
```diff
@@ -0,0 +1,75 @@
...
+//------------------------------------------------------------------------------
+
+using AppKit;
+
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory
+{
+    /// <summary>
+    /// Additional parameters used in acquiring user's authorization
+    /// </summary>
+    public class PlatformParameters : IPlatformParameters
```
kpanwar: Just as an FYI, we would also need to support PromptBehavior enumeration in Mac. I don't expect you to do this work. I will add it once your PR is merged and will cut a release to include the same.

src/ADAL.PCL.iOS/BrokerKeyHelper.cs
```diff
@@ -73,7 +73,12 @@ internal static byte[] GetRawBrokerKey()
                     ValueData = byteData
                 };
 
-                SecStatusCode code = SecKeyChain.Add(record);
+                var result = SecKeyChain.Add(record);
```
kpanwar: we don't need brokerkeyhelper class here at all.

src/ADAL.PCL.iOS/Properties/AssemblyInfo.cs
```diff
@@ -39,7 +39,4 @@
 // The following GUID is for the ID of the typelib if this project is exposed to COM
 [assembly: Guid("ff47962a-d498-4c63-b7e9-4db3653ad7e0")]
 
-// Assembly version information is in file ADAL.Common\CommonAssemblyInfo.cs
-
-[assembly: AssemblyDelaySign(true)]
-[assembly: AssemblyKeyFile(@"..\..\build\35MSSharedLib1024.snk")]
```
kpanwar: we require delay signing for our libraries. any specific reason this was removed?

src/ADAL.PCL.iOS/TokenCachePlugin.cs
```diff
@@ -33,6 +33,12 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 {
     internal class TokenCachePlugin : ITokenCachePlugin
     {
+#if MAC
```
kpanwar: in native library we do not persist the cache in keychain when using Mac. the persistence is completely owned by the developer by implemeting the delegates

src/ADAL.PCL/WsTrustRequest.cs
```diff
@@ -78,16 +78,6 @@ internal static class WsTrustRequest
             }
 
             StringBuilder messageBuilder = BuildMessage(DefaultAppliesTo, wsTrustAddress, credential);
-            string soapAction = XmlNamespace.Issue.ToString();
```
kpanwar: why was this code removed? We need this for integrated auth support.

src/ADAL.PCL.iOS/BrokerKeyHelper.cs
mhutch: This isn't used in the Mac backend. It was causing a warn-as-error build failure in the iOS backend.

src/ADAL.PCL.iOS/Properties/AssemblyInfo.cs
mhutch: This was failing when building on Mac. I moved it to the csproj instead.

src/ADAL.PCL/WsTrustRequest.cs
mhutch: It was dead code, causing warning-as-error build failures.

src/ADAL.PCL.Mac/PlatformParameters.cs
mhutch: From what I can see reading the source of the obj-c version, the dialog isn't actually modal, more like a singleton. This seems a little odd.

src/ADAL.PCL.Mac/PlatformParameters.cs
mhutch: Oh, my mistake. I assumed it was an interface member.

src/ADAL.PCL.iOS/TokenCachePlugin.cs
mhutch: Is there a reason for this or was it simply never implemented?

src/ADAL.PCL.Mac/AuthenticationAgentNSWindowController.cs
mhutch: Fixed!

kpanwar:
:shipit:

PR #451, opened 2016-06-09
Fix to allow hardware based keys to perform authentication.

pscho-msft:
We should only up-level software keys using the CAPI PROV_RSA_FULL and
PROV_RSA_SCHANNEL providers and never up-level hardware based keys.


brentschmaltz:
:shipit:

Issue #455, opened 2016-06-14
Add Serviceability attribute to assemblies for desktop .NET to enable servicing libraries


polita:
In .NET 4.5.1, the CLR team added a feature for serviceability that allows library owners to service a library installed locally to an app without having to GAC it. They use something called a breadcrumb list to keep track of where the assembly is located, and if the assembly is serviced through MU, all occurrences in the breadcrumb list get updated.

To make this work, it looks like you need an assembly attribute for serviceability:
[assembly: AssemblyMetadata(“Serviceable”, “True”)]

PR #459, opened 2016-06-17
Add Serviceability attribute to .NET desktop assemblies so we can ser…

polita:
…vice them

@kpanwar @abhishek58g Please take a look. I applied the attribute to the assemblies that run on desktop only. #455


kpanwar:
thanks @polita!

Issue #460, opened 2016-06-20
TokenCache not working properly for UserCredential scenario


AndreaCuneo:
Hi,

the token cache doesn't take into account the password for the caching of the token. This leads to some unfortunate scenario where a malicious user can get access to a token only knowing the username if the user has acquired a token in the last hour (more with refresh).

To reproduce:

```
    var clientId = "app-id";
    var resourceId = "resource-id";
    var authority = "https://login.windows.net/" + "app-tenant";
    var correctCredentials = new UserCredential("username", "right-password");
    var wrongCredentials = new UserCredential("username", "wrong-password");
    var ctx = new AuthenticationContext(authority);
    var res1 = ctx.AcquireToken(resourceId, clientId, correctCredentials);
    var res2 = ctx.AcquireToken(resourceId, clientId, wrongCredentials);
```

Expected result:
- the second call to AcquireToken should fail

Actual result:
- the second call return a valid token

edit: I'm using V2 but looking at the code the problem is still present.

AndreaCuneo:
This has also another subtle problem were the cache, caches the "failed to acquire token" response when the wrong credentials are used first.

To reproduce:

```
    var clientId = "app-id";
    var resourceId = "resource-id";
    var authority = "https://login.windows.net/" + "app-tenant";
    var correctCredentials = new UserCredential("username", "right-password");
    var wrongCredentials = new UserCredential("username", "wrong-password");
    var ctx = new AuthenticationContext(authority);
    var res1 = ctx.AcquireToken(resourceId, clientId, wrongCredentials);
    var res2 = ctx.AcquireToken(resourceId, clientId, correctCredentials);
```

Expected result:
- the second call to AcquireToken should succeed

Actual result:
- the second call return an exception

wouterroos:
I can confirm that this is an issue. I'm currently able to circumvent this issue by passing a `new TokenCache()` argument when creating the `AuthenticationToken`.

kpanwar:
this is by design. We always look into cache for the tokens. you should consider using web view where you have options to force re-auth or you can clear or pass a new token cache instance. on a side note, can you provide the cache dump that actually shows caching of "failed to acquire token"?

wouterroos:
After doing some testing, I found that my scenario slightly defers from the one @AndreaCuneo mentioned. Consider the following test:

    [TestMethod]
    public async Task CredentialTest()
    {
        var clientId = "app-id";
[...]

PR #463, opened 2016-06-24
set displayableId as the username for cache lookup

kpanwar:
per #460, set displayableId as the username for cache lookup. This will
make sure that when user2 creds are passed to the API, it does not
return the token for user1.


weijjia:
:shipit:

PR #464, opened 2016-06-25
Support for prompt behavior in Xamarin.Mac

kpanwar:
Support for prompt behavior in Xamarin.Mac. PromptBehavior.Never is not supported as we don't do hidden browser flows.


weijjia:
:shipit:

PR #465, opened 2016-06-25
Add ADAL.PCL.Mac45 project

bratsche:
This should be the same as the existing Mac project except
that it uses Xamarin.Mac .NET 4.5 Framework as its target
instead of Xamarin.Mac Mobile Framework.


kpanwar:
Why is there a need for .net 45 when mobile framework is the recommended way for unified API?

bratsche:
We want to use this for Xamarin Studio and it's using the .NET 4.5 target

kpanwar:
It seems like .net framework is more encompassing with minimal drawbacks. We would like to support only one instead of multiple targets. We can update existing project to target .net 45 instead. Thoughts?

bratsche:
Sure, if that's what you want to do I'll submit a separate PR.

PR #467, opened 2016-06-25
Fix the readme to point to your samples and documentation if they exist.

JimDaly:
The link (https://github.com/AzureADSamples) returns 404, so your samples and documentation are not discoverable. I found this repo (https://github.com/nhwilly/AzureADSamples) which suggests that content has been migrated to (https://github.com/azure-samples?query=active-directory) but this doesn't seem to include any documentation. 

So if the link is dead and documentation is not discoverable, you might as well remove the whole paragraph. But I'd actually prefer that you update it with accurate information.


kpanwar:
We will update it with correct link instead.

PR #468, opened 2016-06-27
Chrmann/fix adal cache lookup2

crmann1:
Leave a comment


kpanwar:
made different fix to address the issue cleanly as per PR #477 .

PR #476, opened 2016-07-01
verify certificate chain for broker app

kpanwar:
In Android API 15 and under, adal could communicate with a fake broker app where 
the attacker could take the certificate from a valid broker app and create a
fake certificate chain since the lower sdk do not perform certificate
chain validation.

The fix will create a certificate chain based on the returned signatures
from the os, and validate the certificate chain.


weijjia:
:shipit:

PR #477, opened 2016-07-01
Unexpected multiple_matching_tokens_detected

kpanwar:
Unexpected multiple_matching_tokens_detected even when User has been
provided. The problem occurs because AcquireTokenHandlerBase initializes
CacheQueryData in its constructor and it uses fields such as
this.UniqueId and this.DisplayableId that have not been initialized yet.
They get initialized by subclasses such as AcquireTokenSilentHandler or
AcquireTokenInteractiveHandler after the base class constructor exits.


RPangrle:
:shipit:

Issue #482, opened 2016-07-07
UserPasswordCredential doesn't support .NET Core


justinyoo:
Hi, Team.

When I import this ADAL into my .NET Core library, it complains at the `UserPasswordCredential` class like:

![image](https://cloud.githubusercontent.com/assets/1538528/16645429/37ce670e-4467-11e6-80a1-ca966a85ac76.png)

It doesn't seem that I can create a custom class inheriting the `UserCredential` class for workaround. Is there any suggestion that I can use the `UserPasswordCredential` in .NET Core?

Cheers,

kpanwar:
This is not supported by design.

justinyoo:
@kpanwar Thanks for getting back to me.

Is there any timeline that we can use `UserPasswordCredential` in .NET Core? I don't understand "not supported by design". Does that mean `UserPasswordCredential` won't support .NET Core? If not, could you provide a workaround?

kpanwar:
Yes, we will not add support for UserPasswordCredential in .NET core. There are other ways to authenticate as well. See https://github.com/Azure-Samples/active-directory-dotnet-deviceprofile

novogeek:
We have a daemon service that was authenticating to an API **as a user** and we were using UserPasswordCredential method to fetch access token (by passing username and password). But when we upgraded the solution to .Net Core (and the latest ADAL package), this has broken. If UserPasswordCredential is no more supported, what is the alternate approach to authenticate non-interactively for the aforementioned scenario?

Squiggle:
Ditto. I'm trying to authenticate against a 3rd party API in an Azure WebJob, which was previously working fine using UserPasswordCredential.

What is now the recommended approach here?

talynone:
So how would one write a headless client for lets say CRM dynamics which requires an Oath 2.0 authentication to Active Directory to access the web api in .NET core?

psignoret:
The idea here is that for apps to collect and store a user's username and password is not the correct approach for non-interactive authentication. Here are some details on why you shouldn't do this:

 - http://stackoverflow.com/a/29110987
 - http://stackoverflow.com/a/28597758
 - http://stackoverflow.com/a/26795582
 - http://stackoverflow.com/a/39250380


[...]

PR #485, opened 2016-07-12
Change scope of UserAuthType from internal to public

justinyoo:
Hi, Team.

Regarding to the issue #482 , I changed the scope of the enum `UserAuthType` from internal to public. With this change, the `UserCredential` class can be inherited for .NET Core application, like `UserPasswordCredential`, which only supports .NET full framework.

It would be great if you review this change and accept it.

Cheers,


kpanwar:
This scenario is not supported by design. We want developers to move away from resource owner grant flows.

Issue #490, opened 2016-07-15
AuthenticationAgentContinuationHelper throws null pointer exception


kpanwar:
AuthenticationAgentContinuationHelper throws null pointer exception when there is no ADAL call pending. ADAL should check if there is a call pending, then only save the response from activity result.

kpanwar:
fixed in 3.13.1

PR #491, opened 2016-07-15
OnResume throws nullpointerexception in client app

kpanwar:
AuthenticationAgentContinuationHelper throws null pointer exception when
there is no ADAL call pending. ADAL should check if there is a call
pending, then only save the response from activity result. #490


weijjia:
:sheep: :shipit: :rocket:

Issue #487, opened 2016-07-14
AuthenticationContext.AcquireTokenAsync() occasionally returns null access token from cache


rcarduner:
Since upgrading to ADAL library version 3.10.305231913 (from 2.4.x), we have been seeing occasional cases where the access token returned from the cache is null. Specifically, we’re seeing a call to AcquireTokenAsync returns a successful AuthenticationResult, but the AccessToken member is null. There is no indication of any failure.
 
We are using application identities in a multi-tenant application.  We see this issue especially when running tests under high-volume, multi-threaded conditions across multiple tenants.  In addition, we’ve seen many extra calls out to the Azure Active Directory Identity Provider that we did not expect, after the first token is successfully acquired.  Based on looking at and debugging the ADAL library, it appears that the multi-tenant scenario with application identities may not be handled correctly and that legitimate items are being altered and removed from the cache when they shouldn’t be.

The area in question is in TokenCache.cs, with code added during commit 74eb116886104222ccf6b2eb48f8e0f29f8d1d67, 28 Sep 2015.  The following was added to **LoadSingleItemFromCache()**.  We believe this case should only apply when a user identity is involved, since an application identity should not have a refresh token.  Focusing in on the highlighted lines below:

```C#
// check for tokens issued to same client_id/user_id combination, but any tenant.
if (returnValue == null)
{
    List<KeyValuePair<TokenCacheKey, AuthenticationResultEx>> itemsForAllTenants = this.QueryCache(null, cacheQueryData.ClientId, cacheQueryData.SubjectType, cacheQueryData.UniqueId, cacheQueryData.DisplayableId);
    if (itemsForAllTenants.Count != 0)
    {
        returnValue = itemsForAllTenants.First();
    }

    // check if the token was issued by AAD
//>>>>>
    if (returnValue != null &&
        Authenticator.DetectAuthorityType(returnValue.Value.Key.Authority) != AuthorityType.ADFS)
//<<<<<
    {
        //remove access token to redeem refresh token against a different tenant.
        returnValue.Value.Value.Result.AccessToken = null;
    }
    else
    {
        returnValue = null;
    }
}
```

The basic scenario is as follows:

1. Create an AuthenticationContext with a **tenant ID 1** as part of the authority URI and call AcquireTokenAsync().  Token is acquired successfully from the Azure Active Directory Identity Provider and placed in the cache.
[...]

PR #492, opened 2016-07-16
AcquireTokenAsync returns null access token

kpanwar:
AuthenticationContext.AcquireTokenAsync() occasionally returns null
access token from cache per #487
This PR also includes fix for https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/493


weijjia:
:shipit:

Issue #470, opened 2016-06-28
Why was the SecureString parameter removed from ClientCredential?


acgritt:
We are currently using ADAL 2.26.305102204 within our Service Fabric application for accessing the Azure Key Vault.  Due to how we are securing the Client Secret in the settings.xml, when we retrieve it, it returns as a SecureString.  In the version of ADAL we are using we are able to pass this into the ClientCredential.  In the most current version of ADAL this support has been removed and we are curious to know the reason for this removal and if it my be coming back.

As there isn't a safe/clean way to convert a SecureString back into a regular string this is a piece of functionality that we would prefer to have remain.

kpanwar:
Portable libraries don't have support for SecureString, but we will make API update to include the support in a different way.

kpanwar:
Fixed in 3.13.0

Issue #458, opened 2016-06-17
SecurePassword.Clear is throwing "Instance is read-only"


normanhh3:
Hi Folks,

I'm trying to use the System.Security.SecureString to forward on user credentials from my apps user to authenticate with.

When I do that using a SecureString the ApplyTo method of [this class](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/c6b050a160b79f88e5efc9e3014a0ad836b98d67/src/ADAL.PCL.Desktop/UserPasswordCredential.cs) is doing a Clear which is resulting in an InvalidOperationException with the message "Instance is read-only."

My usual searching has not turned up sufficient advise to help me through this issue.

1. Am I creating the SecureString incorrectly?  I am using the AppendChar method.
1. Should there be a check on the Clear method that uses the IsReadOnly property of SecureString to determine if it can be done safely?

Thanks for the feedback!

kpanwar:
Fixed in 3.13.0

PR #498, opened 2016-07-25
Support SecureString for client secrets

kpanwar:
Added support for SecureString as client secret per #470.
Fixed SecureString.Clear issue where ADAL was not checking if string was
readonly per #458.
Client Resiliency did no retry on connection timeouts.

### DIFF
--- a/src/ADAL.PCL.Desktop/SecureClientSecret.cs
+++ b/src/ADAL.PCL.Desktop/SecureClientSecret.cs
@@ -1,0 +1,76 @@
+﻿//----------------------------------------------------------------------
+//
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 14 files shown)


src/ADAL.PCL.Desktop/SecureClientSecret.cs
```diff
@@ -0,0 +1,76 @@
...
+            this.secureString = secret;
+        }
+
+        public void ApplyTo(IDictionary<string, string> parameters)
+        {
+            var output = new char[secureString.Length];
+            IntPtr secureStringPtr = Marshal.SecureStringToCoTaskMemUnicode(secureString);
+            for (int i = 0; i < secureString.Length; i++)
+            {
+                output[i] = (char) Marshal.ReadInt16(secureStringPtr, i*2);
```
weijjia: nit: i * 2

src/ADAL.PCL/ISecureClientSecret.cs
```diff
@@ -0,0 +1,43 @@
...
+//------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory
+{
+    /// <summary>
+    /// Interface to allow for client secret to be passed in as a SecureString
+    /// </summary>
+    public interface ISecureClientSecret
```
weijjia: Do you really need the interface?

kpanwar: Yes, this interface is what allows for dependency injection at runtime. It is an interface because there is no default implementation.

weijjia: I'm not seeing any place you're trying to inject dependency.

kpanwar: Developer passes in the SecureSecret instance when creating client credential. We use the same approach for IClientAssertionCertificate.

weijjia: k, fair.

weijjia:
:shipit:

Issue #499, opened 2016-07-26
[2.28] AuthenticationContext.AcquireToken(string, string, UserCredential) does not use UserId when retreiving tokens from TokenCache


markcowl:
When using a call pattern like this:

UserCredential credential = new UserCredential(userId, password);
result = context.AcquireToken(config.ResourceClientUri, config.ClientId, credential);

If the token cache contains a token with the same Authority, Resource, and ClientId, but a different UserId, that token is incorrectly returned.

kpanwar:
This issue is pending release. Here is the exact fix https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/498/commits/398c31be97888032760f698966839d72612fe878#diff-c16713ffecc5e8ae253814844a26e365L113

kpanwar:
Fixed in 2.28.1

PR #501, opened 2016-07-26
UserCredential.Username is not used for cache lookups.

kpanwar:
Per issue #499.
When using a call pattern like this:

UserCredential credential = new UserCredential(userId, password);
result = context.AcquireToken(config.ResourceClientUri, config.ClientId, credential);

If the token cache contains a token with the same Authority, Resource, and ClientId, but a different UserId, that token is incorrectly returned.


weijjia:
:shipit:

Issue #500, opened 2016-07-26
Null reference exception in Device authentication flow in 3.11, 3.12 for .netstandard1.6


markcowl:
The Device authentication flow will always throw a null reference exception when attempting to acquire a token by device code.

This is because the constructor for ```AcquireTokenHandlerBase``` uses [```requestData.ClientKey```](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/master/src/ADAL.PCL/AcquireTokenHandlerBase.cs#L53) and [```requestData.Resource```](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/master/src/ADAL.PCL/AcquireTokenHandlerBase.cs#L59), and they are not initialized when it is called. 

[The constructor for AcquireTokenByDeviceCodeHandler] (https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/master/src/ADAL.PCL/AcquireTokenByDeviceCodeHandler.cs#L37)  sets ```requestData.ClientKey``` [after its base constructor is called](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/master/src/ADAL.PCL/AcquireTokenByDeviceCodeHandler.cs#L40) and ```requestData.Resource``` is [never  initialized](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/master/src/ADAL.PCL/AuthenticationContext.cs#L188) 

Also, the null check on deviceCodeResult [here](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/master/src/ADAL.PCL/AcquireTokenByDeviceCodeHandler.cs#L41) occurs after deviceCodeResult is dereferenced on the previous line.

This is a blocking issue for Device Flow using both 3.11 and 3.12 versions of the library, meaning that you cannot acquire a token using the device flow if you are using the RTM CoreCLR library

kpanwar:
Thanks for filing the issue. I am in the process of making fixes and adding coverage. I will update the bug with pull request reference once done.

markcowl:
@kpanwar Thanks, I have a local fix that I'm using for now, I can prepare a pull request if you like.

kpanwar:
Fixed in 3.13.0

PR #503, opened 2016-07-26
Null reference exception in Device authentication flow.

kpanwar:
per #500, The Device authentication flow will always throw a null reference exception when attempting to acquire a token by device code. This is because the constructor for AcquireTokenHandlerBase uses requestData.ClientKey and requestData.Resource, and they are not initialized when it is called.


weijjia:
:shipit:

Issue #504, opened 2016-07-27
Add ConfigureAwait(false) to all the async calls


kpanwar:
Add ConfigureAwait(false) to all the async calls so that the developer can call the API in a synchronous manner, if needed.

PR #505, opened 2016-07-27
Add ConfigureAwait(false) to all the async calls.

kpanwar:
per #504, Add ConfigureAwait(false) to all the async calls so that the developer can call the API in a synchronous manner, if needed.


weijjia:
:shipit: :sheep: :rocket:

PR #515, opened 2016-08-16
Returning entire HttpErrorResponse as InnerException to HttpRequestException

abhishek58g:
Returning entire HttpErrorResponse as InnerException to
HttpRequestException

### DIFF
--- a/src/ADAL.PCL/EncodingHelper.cs
+++ b/src/ADAL.PCL/EncodingHelper.cs
@@ -201,7 +201,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
             return stream;
         }
 
-        private static T DeserializeResponse<T>(string response)
+        public static T DeserializeResponse<T>(string response)
         {
             DataContractJsonSerializer serializer = new DataContractJsonSerializer(typeof(T));
 
--- a/src/ADAL.PCL/EncodingHelper.cs
+++ b/src/ADAL.PCL/EncodingHelper.cs
@@ -28,6 +28,8 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.IO;
+using System.Runtime.Serialization.Json;
 using System.Text;
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory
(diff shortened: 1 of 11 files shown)


kpanwar:
There is no test case to test this change. Can you add one using the mocks where the response contains additional json keypair?

abhishek58g:
Added the test case

src/ADAL.PCL/EncodingHelper.cs
```diff
@@ -189,6 +191,31 @@ public static void SecureClear(this char[] chars)
...
+        {
+            MemoryStream stream = new MemoryStream();
+            StreamWriter writer = new StreamWriter(stream);
+            writer.Write(s);
+            writer.Flush();
+            stream.Position = 0;
+            return stream;
+        }
+
+        private static T DeserializeResponse<T>(string response)
```
kpanwar: this code is duplicate from AdalHttpClient. Please consolidate to 1 function.

src/ADAL.PCL/HttpClientWrapper.cs
```diff
@@ -129,11 +130,15 @@ public int TimeoutInMilliSeconds
                 {
                     try
                     {
-                        throw new HttpRequestException(string.Format(CultureInfo.CurrentCulture, " Response status code does not indicate success: {0} ({1}).", (int)webResponse.StatusCode, webResponse.StatusCode));
+                        StringBuilder httpErrorResponse = new StringBuilder();
```
kpanwar: if webResponse contains string then why is another stream being created just to create another string? This could be simplified to "new exception(webResponse.ResponseString)"

src/ADAL.PCL/HttpClientWrapper.cs
```diff
@@ -158,7 +158,7 @@ public int TimeoutInMilliSeconds
                 }
             }
 
-            return new HttpWebResponseWrapper(await response.Content.ReadAsStreamAsync().ConfigureAwait(false), headers, response.StatusCode);
+            return new HttpWebResponseWrapper(await response.Content.ReadAsStringAsync(), headers, response.StatusCode);
```
kpanwar: missing ConfigureAwait(false). We must be careful about not removing configureawait from the code.

kpanwar:
:shipit: :ship:

PR #518, opened 2016-08-17
Update UserInfo class for CoreCLR 1.0.4 issues

markcowl:
In CoreCLR 1.0.4 Json serialization, it is required for constructors of serialized types to be visible to the serializer.  This causes exceptions when running in CoreCLR 1.0.4


kpanwar:
This is changing the public API surface. There are other workarounds like referencing a specific serialization package version.

kpanwar:
Internal constructors exist in authentication result class as well. Wouldn't that result in error as well?

markcowl:
@kpanwar it doesn't.  This issue is the only oen that shows up in our usage.  I would guess though that any class using the DataContractSerializer for Json deserialization would have the same issue.

Issue #520, opened 2016-08-23
ComputeHash method in iOS throws null exception


kpanwar:
ComputeHash method in iOS throws null exception

PR #521, opened 2016-08-23
Fix nullpointer exception and set userbroker to false by default. #520

kpanwar:
Fix nullpointer exception and set userbroker to false by default. #520


weijjia:
:shipit:

Issue #522, opened 2016-08-23
AcquireTokenAsync fails with ArgumentException "Illegal characters in path."


nathandavidson:
AcquireTokenAsync fails with ArgumentException "Illegal characters in path."

Using ADAL 3.13.3 targeting .NET Framework 4.6.1 in a console app project, the following (anonymized) code results in an unexpected exception:

`var context = new AuthenticationContext("https://login.windows.net/myexampletenant.onmicrosoft.com");`
`var token = context.AcquireTokenAsync("https://management.azure.com/", "43214321-4321-4321-4321-432143214321", new UserCredential("user@example.com")).GetAwaiter().GetResult();`

A simple console app with only that code in the Main method crashes with the following stack trace:

> Unhandled Exception: System.ArgumentException: Illegal characters in path.
   at System.IO.Path.LegacyNormalizePath(String path, Boolean fullCheck, Int32 maxPathLength, Boolean expandShortPaths)
   at System.IO.Path.NormalizePath(String path, Boolean fullCheck, Int32 maxPathLength, Boolean expandShortPaths)
   at System.IO.Path.GetFullPathInternal(String path)
   at System.Xml.XmlResolver.ResolveUri(Uri baseUri, String relativeUri)
   at System.Xml.XmlUrlResolver.ResolveUri(Uri baseUri, String relativeUri)
   at System.Xml.XmlTextReaderImpl..ctor(String uriStr, XmlReaderSettings settings, XmlParserContext context, XmlResolver uriResolver)
   at System.Xml.XmlReaderSettings.CreateReader(String inputUri, XmlParserContext inputContext)
   at System.Xml.XmlReader.Create(String inputUri, XmlReaderSettings settings, XmlParserContext inputContext)
   at System.Xml.Linq.XDocument.Load(String uri, LoadOptions options)
   at Microsoft.IdentityModel.Clients.ActiveDirectory.MexParser.<FetchMexAsync>d__2.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.IdentityModel.Clients.ActiveDirectory.MexParser.<FetchWsTrustAddressFromMexAsync>d__1.MoveNext()
[...]

PR #524, opened 2016-08-24
update code to create stream from the response instead of passing the…

kpanwar:
… response string to XmlDocument. #522


weijjia:
:shipit:

PR #530, opened 2016-09-01
add docfx.json to generate yaml files

superyyrrzz:
@chenkennt @vwxyzh

Run DocFX using this docfx.json, you can get yaml files from **_merged** folder, which should be checked in to ADALRef-pr.

### DIFF
--- a/docfx.json
+++ b/docfx.json
@@ -90,7 +90,7 @@
     ],
     "fileMetadata": {
       "platform": {
-        "_yml/ADAL.PCL/*.yml": "Library",
+        "_yml/ADAL.PCL/*.yml": ["Android", "CoreCLR", "Desktop", "iOS", "WinRT"],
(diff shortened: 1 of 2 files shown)


docfx.json
```diff
@@ -0,0 +1,112 @@
...
+      },
+      {
+        "files": [ "*.yml" ],
+        "src": "_yml/ADAL.PCL.WinRT",
+        "dest": "_merged"
+      }
+    ],
+    "fileMetadata": {
+      "platform": {
+        "_yml/ADAL.PCL/*.yml": "Library",
```
superyyrrzz: I'm not sure whether these platform names are accurate.

kpanwar:
We only support one way of generating docs and that is by using summary tags.

vwxyzh:
:+1:

PR #535, opened 2016-09-08
remove extra space

SomkaPe:


SomkaPe:
@kpanwar  please provide release branches where we should also push "adding Versions section to readme"  change

kpanwar:
Master and servicing

PR #537, opened 2016-09-13
OBO test coverage and user assertion logic update

kpanwar:
The PR adds extensive OBO test coverage. It also adds the check to always match on user assertion when provided instead of relying on cache entry. this is because the developer may pass an updated assertion with new claims like MFA etc.


addev-ashish:
:shipit:

Issue #539, opened 2016-09-16
UWP fails when handling WebAuthenticationStatus.ErrorHttp


kpanwar:
The problem originates in WebUI.ProcessAuthorizationResult (from src\ADAL.PCL.WinRT\WebUI.cs). In the event that the system returns WebAuthenticationStatus.ErrorHttp, it creates an AuthorizationResult with the returned HTTP status code in the returnedUriInput field.

    case WebAuthenticationStatus.ErrorHttp:
        result = new AuthorizationResult(AuthorizationStatus.ErrorHttp, webAuthenticationResult.ResponseErrorDetail.ToString(CultureInfo.CurrentCulture));

However, the AuthorizationResult c’tor (in src\ADAL.PCL\AuthorizationResult.cs) only has special handling for the UserCancel and UnknownError result statuses. So, ErrorHttp results are passed off to the ParseAuthorizeResponse method which tries parsing the returnedUriInput as a Uri and throws. This exception isn’t being wrapped up in an AdalException, so it’s not processed by our auth code and eventually crashes the app.

abhishek58g:
@kpanwar Can we close this issue ?

kpanwar:
merged. Pending release.

kpanwar:
fixed in https://www.nuget.org/packages/Microsoft.IdentityModel.Clients.ActiveDirectory/3.13.7

PR #540, opened 2016-09-16
UWP fails when handling WebAuthenticationStatus.ErrorHttp 

kpanwar:
The problem originates in WebUI.ProcessAuthorizationResult (from src\ADAL.PCL.WinRT\WebUI.cs). In the event that the system returns WebAuthenticationStatus.ErrorHttp, it creates an AuthorizationResult with the returned HTTP status code in the returnedUriInput field.

case WebAuthenticationStatus.ErrorHttp:
    result = new AuthorizationResult(AuthorizationStatus.ErrorHttp, webAuthenticationResult.ResponseErrorDetail.ToString(CultureInfo.CurrentCulture));
However, the AuthorizationResult c’tor (in src\ADAL.PCL\AuthorizationResult.cs) only has special handling for the UserCancel and UnknownError result statuses. So, ErrorHttp results are passed off to the ParseAuthorizeResponse method which tries parsing the returnedUriInput as a Uri and throws. This exception isn’t being wrapped up in an AdalException, so it’s not processed by our auth code and eventually crashes the app. #539

### DIFF
--- a/src/ADAL.PCL.WinRT/WebUI.cs
+++ b/src/ADAL.PCL.WinRT/WebUI.cs
@@ -116,7 +116,12 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
                     result = new AuthorizationResult(AuthorizationStatus.Success, webAuthenticationResult.ResponseData);
                     break;
                 case WebAuthenticationStatus.ErrorHttp:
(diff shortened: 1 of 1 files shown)


weijjia:
:shipit: 🚀 🐑

PR #546, opened 2016-10-04
Fix for :- Exception in ADAL webview on EvoSTS consent screen

abhishek58g:
The javascript code is getting accounted as non https url in ADAL.


kpanwar:
:shipit:

Issue #548, opened 2016-10-07
Renewing Cert for WinRT App v2


abhishek58g:
The present cert has expired.

ncarandini:
Looking at commit 7c6898d this should be closed, doesn't it?

PR #550, opened 2016-10-07
Renewing the WinRT certificate #548

abhishek58g:
Adding the renewed certificate.


kpanwar:
pull request goes to dev. then we merge to master.

PR #556, opened 2016-10-11
Update readme.md to point to samples and new/v3 docs.

BryanLa:


abhishek58g:
:shipit:

BryanLa:
Thanks @abhishek58g. Is there someone I should also contact to have this merged, or does that happen as part of another review process?

PR #557, opened 2016-10-19
Adding an optional readonly address bar in Login Dialog Box.

Medii:
Added a new optional DiplayAddressBar property which is 'False' by default in PlatformParameter and updated all dependent classes. 
Wrapped CustomWebBrowser in UserControlWebBrowser along with a readonly TextBox that displays Adfs Address Url on top of the WebBrowser control.
It can assure the user that the login screen belongs to the ADFS.

### DIFF
--- a/src/ADAL.PCL.Desktop/ADAL.PCL.Desktop.csproj
+++ b/src/ADAL.PCL.Desktop/ADAL.PCL.Desktop.csproj
@@ -111,6 +111,10 @@
     </Compile>
     <Compile Include="StaTaskScheduler.cs" />
     <Compile Include="TokenCachePlugin.cs" />
+    <Compile Include="UserControlWebBrowser.cs">
+      <SubType>UserControl</SubType>
+    </Compile>
+    <Compile Include="UserControlWebBrowser.Designer.cs" />
     <Compile Include="UserPasswordCredential.cs" />
     <Compile Include="WebBrowserInterfaces.cs" />
     <Compile Include="WebBrowserNavigateErrorEventArgs.cs" />
@@ -134,6 +138,9 @@
     <None Include="..\..\build\35MSSharedLib1024.snk">
       <Link>35MSSharedLib1024.snk</Link>
     </None>
+  </ItemGroup>
+  <ItemGroup>
+    <EmbeddedResource Include="UserControlWebBrowser.resx" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
--- a/src/ADAL.PCL.Desktop/InteractiveWebUI.cs
+++ b/src/ADAL.PCL.Desktop/InteractiveWebUI.cs
@@ -30,12 +30,18 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal
     internal class InteractiveWebUI : WebUI
     {
         private WindowsFormsWebAuthenticationDialog dialog;
+        private readonly bool displayAddressBar;
+
+        public InteractiveWebUI(bool displayAddressBar = false)
+        {
+            this.displayAddressBar = displayAddressBar;
+        }
 
         protected override AuthorizationResult OnAuthenticate()
         {
             AuthorizationResult result;
 
-            using (this.dialog = new WindowsFormsWebAuthenticationDialog(this.OwnerWindow))
+            using (this.dialog = new WindowsFormsWebAuthenticationDialog(this.OwnerWindow, this.displayAddressBar))
             {
                 result = this.dialog.AuthenticateAAD(this.RequestUri, this.CallbackUri);
             }
--- a/src/ADAL.PCL.Desktop/PlatformParameters.cs
+++ b/src/ADAL.PCL.Desktop/PlatformParameters.cs
@@ -32,11 +32,9 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     /// </summary>
     public class PlatformParameters : IPlatformParameters
     {
-        /// <summary>
-        /// 
-        /// </summary>
-        /// <param name="promptBehavior"></param>
-        public PlatformParameters(PromptBehavior promptBehavior):this(promptBehavior, null)
+        private bool displayAdfsUrlAddressBar;
+
+        public PlatformParameters(PromptBehavior promptBehavior) : this(promptBehavior, null)
         {
         }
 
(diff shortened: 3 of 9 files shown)


kpanwar:
I am going to close this PR. Let me know if you have any other questions or details around the assurance provided.

AdamNachman:
Medii has provided feedback, and this is not an uncommon requirement. You've encouraged no debate or discussion, but simply closed this unilaterally.

Would you like to engage with the community on this before closing?

kpanwar:
I did not hear back for 6 days. We are open to discussion ofcourse

kpanwar:
After careful evaluation, we are not going to accept this PR. Please see the rationale below.

Given that application can reproduce pixel by pixel what you are trying to show here with the disabled textbox, the user has no reason to trust the ADAL dialog any more than he/she does today.

A full blown browser not same as a webview control and this generally applies across multiple platforms. A desktop application experience is not directly comparable to a web site consumed by a system browser. You can compare the webview under discussion with WAB (Windows Authentication Broker), WAM (Web Account Manager), iOS in-app webview, Android in-app webview and you will see the common theme that URLs are just not shown there.
For the future, we are considering using an actual system browser to handle the authentication and not something that looks like a browser.

Medii:
Thanks for your careful evaluation!
I'm afraid I didn't understand the problem of displaying an address bar to the user.
Technically you can't interact with an actual system browser in order to acquire a token for a desktop application.

kpanwar:
It is technically possible to use system browser to authenticate in a desktop application. Here is how google does it - https://github.com/googlesamples/oauth-apps-for-windows

Medii:
I think you haven't seen their code.  The provided sample is completely unusable in production environments.
Just have a look at the source code, a HttpListener with a random port is initiated! and then the URL of that on the fly in memory created web server is passed as redirectURI to the OAuth Idp!!
Do you think that it's a workaround?! Do you think than any organization let you to open an inbound port for their client for such a simple requirement?

kpanwar:
It is not a simple requirement, per say. Protected user credentials from the application is a big challenge that is being solved using/ by implementing specs and documents linked in the Advanced Reading section of the link I shared with you. There is an industry wide push to move away from in-app webviews.

Please refer to https://developers.googleblog.com/2016/08/modernizing-oauth-interactions-in-native-apps.html and https://www.gluu.org/blog/webviews-are-bad-use-appauth/ for understanding the industry-wide push.

The inbound port is not meant for connections from the outside as it uses loopback.

Medii:
Thanks for your comment and sorry for misunderstanding. I didn't notice the address is loopback and  I wasn't aware of https://tools.ietf.org/html/draft-ietf-oauth-native-apps-05.
You are right, but by simple requirement, I meant it in the context of your helper library and our requirement by considering current industry constraints.
Also using HttpListener in Windows Desktop applications has some difficulties like windows permissions for executing HttpListener non-admins.

kpanwar:
The sample code I shared used TcpListener and not httplistener. TCP works at a lower layer and processes that are non admin can also spin up a tcp port socket.

Medii:
Unfortunately, your sample code uses TcpListener just to find an available random port and for receiving the token it starts HttpListener.
As you mentioned it's a low level endpoint in which you can't directly communicate with another HTTP endpoint unless you implement your required http verbs!
By the way, that sample code throws permission exception when the owner user is a Non Admin user.
And the workaround for fixing that permission makes deployment difficult.
I think it's just a good example and the idea doesn't work in production.

kpanwar:
Thanks for running through the sample. The standard itself is a work in progress. Assuming the app is deployed by an IT admin, it is possible to roll out right permission set for http listener. I don't see it as a technical blocker.

Medii:
I executed the code under a normal windows account and it threw permission exception. 

-------- Original message --------From: Kanishk Panwar <notifications@github.com> Date: 25/10/2016  5:12 PM  (GMT+10:00) To: AzureAD/azure-activedirectory-library-for-dotnet <azure-activedirectory-library-for-dotnet@noreply.github.com> Cc: Mehdi Khezrian <mkhezrian@yahoo.com>, Mention <mention@noreply.github.com> Subject: Re: [AzureAD/azure-activedirectory-library-for-dotnet] Adding an optional readonly address bar in Login Dialog Box. (#557) 
The sample code I shared used TcpListener and not httplistener. TCP works at a lower layer and processes that are non admin can also spin up a tcp port socket.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/AzureAD/azure-activedirectory-library-for-dotnet","title":"AzureAD/azure-activedirectory-library-for-dotnet","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/AzureAD/azure-activedirectory-library-for-dotnet"}},"updates":{"snippets":[{"icon":"PERSON","message":"@kpanwar in #557: The sample code I shared used TcpListener and not httplistener. TCP works at a lower layer and processes that are non admin can also spin up a tcp port socket."}],"action":{"name":"View Pull Request","url":"https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/557#issuecomment-255945716"}}}

kpanwar:
If by normal account you mean standard user (non-admin), then I did not see any permission exception on my side and my point about deployment of app and permission by IT admin still stands. I think this discussion is orthogonal to the PR itself, but I would encourage you to follow up on google's github for the issues you are seeing.

AdamNachman:
> Given that application can reproduce pixel by pixel what you are trying to show here with the disabled textbox, the user has no reason to trust the ADAL dialog any more than he/she does today

The issue is that the user cannot see the url to which he is connecting, and therefore has no information as to whether he is pointing to the right location. 

The fact that another malicious application may or may not be able to replicate those UI elements pixel by pixel is largely irrelevant. That's equivalent to saying "we won't build something in case someone makes a copy and tricks the user", and I don't believe has any merit in this case.

The move away from the webview is a separate discussion entirely, and should be part of the github conversation.

vibronet:
Adam, we can agree to disagree here. ADAL is not an application, it is a library used by multiple applications, and what we do in it will have far reaching ripples. Providing a visual hint that cannot be trusted, given that there's no way for the user to ascertain whether it is faithfully reflecting what's going on, can induce a false sense of security - and that's more than enough merit for us for not doing it. 
Earlier Kanishk provided a list of many components, from multiple software vendors and platforms, which are specifically designed to drive workloads related to web authentication for native apps - and all of them have in common the absence of any affordance displaying the URL. I cannot say for certain whether all those components made that design choice following the same rationale articulated through this thread, but it is certainly telling that it is such a common choice.

Issue #565, opened 2016-11-20
Support for changing the animation style on iOS


bjartebore:
It would be nice to be able to set the animation style of the ADAL ViewController

I believe this would easily be accomplished with a ModalTransitionStyle property (and a TransitioningDelegate property) in the PlatformParameter class.

PR #566, opened 2016-11-20
ios: added posibility to add/set tranisition style for the login form

bjartebore:
Added support for setting transition style on ios #565


src/ADAL.PCL.iOS/PlatformParameters.cs
```diff
@@ -78,5 +78,24 @@ public PlatformParameters(UIViewController callerViewController, bool useBroker)
         /// </summary>
         /// <value>The preferred status bar style.</value>
         public UIStatusBarStyle PreferredStatusBarStyle { get; set; }
+
+        /// <summary>
+        /// Set ths transition style used when the login form view is presented
```
kpanwar: the*

bjartebore:
I have updated the typos you pointed out

PR #570, opened 2016-12-02
prepare for 3.13.8

kpanwar:
Version 3.13.8
=========
This release inclues
- update to nuget package to fix casing of NETStandard.Library nuget dependency
- Add serialization packages to netstandard1.4 target as they are not part of netstandard.


RPangrle:
:shipit:

PR #594, opened 2017-02-01
Expanded resiliency error codes to all 500 status codes. Reflected ch…

jennyf19:
…ange in MockHelper for testing Issue #593


kpanwar:
Looks good. :shipit:

Issue #591, opened 2017-01-18
WS-Trust does not work against sovereign clouds - Adal v3


kpanwar:
WS-Trust does not work against sovereign clouds e.g. - BlackForest because appliesTo address urn is hardcoded to PROD. It should be consumed as a value of cloud_audience_urn returned via user realm response.

jmprieur:
Fixed in Dev

PR #596, opened 2017-02-08
Jennyf/ws trust sovereign clouds

jennyf19:
#591 When logging in as a federated user, the default audience is urn:federation:MicrosoftOnline. However, this does not work for a sovereign cloud service. Code has been changed to use the JSON property cloud_audience_urn from the HTTPS GET response as the audience value. If no value is present, a default value will be used (urn:federation:MicrosoftOnline).


src/ADAL.PCL/AcquireTokenNonInteractiveHandler.cs
```diff
@@ -111,11 +113,15 @@ public AcquireTokenNonInteractiveHandler(RequestData requestData, UserAssertion
                     {
                         throw new AdalException(AdalError.MissingFederationMetadataUrl);
                     }
+                    if (string.IsNullOrWhiteSpace(userRealmResponse.CloudAudienceUrn))
```
kpanwar: this if check and defaultAppliesTo variable should be inside the WsTrustRequest class only. Let the request class make the decision on how the value is used.

tests/Test.ADAL.NET.Unit/NonInteractiveTests.cs
```diff
@@ -127,10 +127,60 @@ public void Initialize()
         }
 
         [TestMethod]
+        [Description("Cloud Audience Urn Test")]
+        public async Task CloudAudienceUrnTest()
```
kpanwar: this should be 2 separate tests.

src/ADAL.PCL/AcquireTokenNonInteractiveHandler.cs
jennyf19: Done

tests/Test.ADAL.NET.Unit/NonInteractiveTests.cs
jennyf19: Changes: "Cloud Audience Urn Test" to test the wsTrustRequest passing it a generic cloud audience urn (ex. "urn:federation:Blackforest"). "Cloud Audience Urn Null Test" to test wsTrustRequest passing it a null value for the cloud audience. 
Added a "Ws-Trust Request Test" to test WsTrustRequest, passing it a test constant value ("urn:federation:MicrosoftOnline"). Also, a "Ws-Trust Request Generic Cloud Urn Test, to test WsTrustRequest, passing it a test constant value ("urn:federation:Blackforest").

kpanwar:
:shipit:

PR #599, opened 2017-02-10
Jennyf/ws trust sovereign clouds v2

jennyf19:
#590 When logging in as a federated user, the default audience is urn:federation:MicrosoftOnline. However, this does not work for a sovereign cloud service. Code has been changed to use the JSON property cloud_audience_urn from the HTTPS GET response as the audience value. If no value is present, a default value will be used (urn:federation:MicrosoftOnline).


kpanwar:
:shipit:

PR #600, opened 2017-02-11
limit network response size to 1MB in adal v2

kpanwar:
limit network response size to 1MB in adal v2


weijjia:
🚢 it 🐑 🚀 
:shipit:

PR #602, opened 2017-02-12
Integrated auth support for client credentials

cedricboon:
ADFS supports authentication of a client by means of an AD account
instead of a client secret. This commit adds support for integrated
authentication to client credentials.

The changes consist of sending an additional oauth parameter
(use_windows_client_authentication=true) and using default credentials
when requesting an oauth token with client credentials without a
client secret.

### DIFF
--- a/src/ADAL.PCL/AcquireTokenHandlerBase.cs
+++ b/src/ADAL.PCL/AcquireTokenHandlerBase.cs
@@ -312,8 +312,15 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 
         private async Task<AuthenticationResultEx> SendHttpMessageAsync(IRequestParameters requestParameters)
         {
+            var isIntegratedAuth = this.ClientKey?.Credential != null && this.ClientKey.Credential.IsIntegratedAuth;
+
+            if (isIntegratedAuth && this.Authenticator.AuthorityType != AuthorityType.ADFS)
+            {
+                throw new NotSupportedException(AdalErrorMessage.UnsupportedIntegratedAuthentication);
+            }
+
             client = new AdalHttpClient(this.Authenticator.TokenUri, this.CallState)
-            { Client = { BodyParameters = requestParameters } };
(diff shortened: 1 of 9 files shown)


jmprieur:
@cedricboon : I'm confused. Is this PR proposing a different problem than using the overload of AcquireToken using a UserCredential?
If you use the following in the case the machine is domain joint or AAD joint, you will get an integrated authentication on Windows (leveraging Kerberos)
``
var result = await authenticationContext.AcquireTokenAsync(resource, clientId, new UserCredential());
``

cedricboon:
Hi @jmprieur,

Next to a shared secret and certificates, ADFS also supports the authenication of a confidential application by means of Windows Integrated Authentication (see attached screenshot
ad_fs_confidential_4)
. This PR adds support for the scenario where a confidential application is running under a domain service account (e.g. a gmsa account) and is trying to access a resource without an end user, so the Client Credentials flow is the correct flow to use. 

The use of the "use_windows_client_authentication" is detailed [here](https://msdn.microsoft.com/en-us/library/mt223850.aspx):
use_windows_client_authentication

OPTIONAL. An OAuth 2.0 confidential client includes this parameter in the POST body of a request to indicate that it will use Windows client authentication and authenticate via the mechanism described in [RFC4559].

The AD FS server ignores this parameter unless its ad_fs_behavior_level is AD_FS_BEHAVIOR_LEVEL_2 or higher.

roszell:
Any chance these changes will get merged anytime soon? We need this functionality... we are wanting to get access tokens from ADFS using Windows auth instead of a client secret.

bgavrilMS:
@jmprieur , @henrik-me - I propose we close this PR because it is outdated and prioritize the feature. We can still use the approach detailed in the PR, but not the code per se.

henrik-me:
@bgavrilMS : agree, it's worth keeping in mind though that the feature has not been planned and we have a number of things ahead of it.

bgavrilMS:
The source code has change substantially. Closing this.

PR #615, opened 2017-02-28
Fix build breaks in samples and release builds

duncanpMS:
This commit fixes some but not all of the build issues in the _dev_ branch.

@kpanwar FYI the _automation\AutomationApp_ and _automation\AutomationApp.UWP_ projects don't currently build on a clean machine.

It doesn't look there is currently a CI build running against the _dev_ branch, only against _master_, which is why these breaks weren't caught sooner. I have a VSTS build definition targeting the _dev_ branch that I'll set up as a CI build once we've sorted out the permissions to set up the GitHub hooks.


kpanwar:
:shipit:

PR #620, opened 2017-03-02
"Automation changes for desktop"

TanujaKulal:


automation/WinFormsAutomationApp/AuthenticationHelper.cs
```diff
@@ -230,5 +246,61 @@ private static IPlatformParameters GetPlatformParametersInstance(string promptBe
 #endif
             return platformParameters;
         }
+
+        public static string FromDictionaryToJson(this Dictionary<string, string> dictionary)
```
bgavrilMS: Should be moved to a separate class

bgavrilMS: In general it's better to use interfaces when dealing with collections.

automation/WinFormsAutomationApp/AuthenticationHelper.cs
```diff
@@ -230,5 +246,61 @@ private static IPlatformParameters GetPlatformParametersInstance(string promptBe
...
+        {
+            var kvs = dictionary.Select(kvp => string.Format("\"{0}\":\"{1}\"", kvp.Key, string.Concat(kvp.Value)));
```
bgavrilMS: Why do you need string.Concat? kvp.Value is a string

bgavrilMS: Use string interpolation to make this string easier to read: 

var kvs = dictionary.Select(kvp => $"\"{kvp.Key}\":\"{kvp.Value}\"");

automation/WinFormsAutomationApp/AuthenticationHelper.cs
```diff
@@ -230,5 +246,61 @@ private static IPlatformParameters GetPlatformParametersInstance(string promptBe
...
+                Resource = resource,
+                ClientId = clientid,
+                UniqueId = uniqueid,
+                DisplayableId = displayableid
+            });
+        }
+
+        private static void UpdateCache(KeyValuePair<TokenCacheKey, AuthenticationResultEx> item, KeyValuePair<TokenCacheKey, AuthenticationResultEx> updated)
+        {
+            bool notifiedBeforeAccessCache = false;
```
bgavrilMS: This flag does not seem to be needed. It's simply set to true and the if branch will always execute.

PR #622, opened 2017-03-05
Added support for using the default IWebProxy inside ADAL.

Strifex:
This is a useful change for enterprises who want to connect from their app to Azure AD from an internal network via proxy.  This is especially useful for Xamarin.Android applications since proxy settings on android usually only apply to the web browser/web-views but not HTTP calls from within 3rd party applications.  For some developers, specifying a proxy programmatically is the only answer for many scenarios and with this change, the ADAL http requests will respect the DefaultWebProxy that is set for that app.


Strifex:
I've updated this PR and the description to reflect a "less-invasive" change to the codebase.  This scenario also accomplishes what was requested, albeit in a different manner.

PR #640, opened 2017-03-24
AppVeyor configuration for PR-based continuous integration.

HSAR:


appveyor.yml
```diff
@@ -0,0 +1,21 @@
+branches:
+  only:
+    - dev
+    - sschen/ci
```
bgavrilMS: Is this required?

HSAR: I've now turned on the setting for "build only if appveyor.yml is in VCS", so no, this can be removed.

appveyor.yml
```diff
@@ -1,8 +1,3 @@
-branches:
-  only:
-    - dev
```
bgavrilMS: I'd keep the dev branch there

HSAR: Commits won't be built unless they have an appveyor.yml file checked in, which for the moment is only on dev. We can re-add this if it becomes a problem for any reason, I think.

HSAR:
Added AppVeyor badges to the README.

PR #656, opened 2017-04-26
Added run settings file to exclude test projects from code coverage

duncanpMS:
Added a test runsettings file that excludes test assemblies and test apps from the code coverage results.

By default, code coverage is calculated on all the projects in the solution. This gives a misleading coverage number: including the sample apps pushes the coverage number down since none of the code is executed, and including the test projects tends to push the coverage number up (tests don't tend to include much branching code so most of the code is executed most of the time).

### DIFF
--- a/build/ADAL.CodeCoverage.runsettings
+++ b/build/ADAL.CodeCoverage.runsettings
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>  
+
+<!-- Customised run settings file to exclude test assemblies from coverage.
+    See https://msdn.microsoft.com/en-us/library/jj159530.aspx for more info. -->
+
+<!-- File name extension must be .runsettings -->  
+<RunSettings>  
+  <DataCollectionRunSettings>  
+    <DataCollectors>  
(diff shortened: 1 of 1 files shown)


kpanwar:
Is there a setting to ignore files as well?

duncanpMS:
Yep. See https://msdn.microsoft.com/en-us/library/jj159530.aspx
Why, which files should to be excluded?

kpanwar:
just wanted to learn. for example, web dialog cannot be tested as we mock web ui options.

duncanpMS:
It isn't normal to exclude specific files - code coverage isn't ever expected to be 100% so it's not usually necessary. One occasion you might exclude files is if they contain generated code, but such code would normally be excluded using custom attributes anyway.

If there is a big chunk of untested code then that's normally a warning flag. It's rare there to have code that couldn't be unit tested at least partially if it were factored differently. The best approach is usually to factor out the hard-to-test bits and test everything else. The percentage of hard-to-test bits should be small enough that it doesn't have much impact on the overall coverage figure.

A quick look at the web dialog code suggests it falls into this category: some of the native calls would be very hard to test, but the rest could be tested with some refactoring.

kpanwar:
The meat of the work is in validating the URLs and handling errors during various events fired. they are specific to the browser control it self.

duncanpMS:
If there is custom code that validates URLs then it could easily be factored out into a helper class and tested directly. Simulating events firing would require more work but is still do-able.

PR #666, opened 2017-05-17
Migrate core library and CoreCLR projects to target .NET Standard 1.1

mjcheetham:
As part of the ongoing work to improve ADAL.NET support running on the CoreCLR.

These changes migrate the _ADAL.PCL_ and _ADAL.PCL.CoreCLR_ projects to the new MSBuild 15* format and now target .NET Standard 1.1 (instead of being Portable Class Libraries).
The existing projects (_Desktop_, _Android_, _iOS_, _WinRT_) remain in the MSBuild 14 format.

Due to an issue in Mono (affecting ADAL on Android and iOS) a workaround was previously added to set the `HttpClientHandler.Proxy` property to the default `IWebProxy` (`WebRequest.DefaultWebProxy`).
Because `WebRequest` type does not exist in .NET Standard (until version 2.0) a new platform plugin (`IWebProxyProvider`) was introduced to allow this workaround to continue to work.

\***NB:** MSBuild 15 / VS2017 will be required to build/open the solution.


src/ADAL.PCL.iOS/AdalInitializer.cs
```diff
@@ -25,24 +25,23 @@
...
-    /// 
-    /// </summary>
     internal static class AdalInitializer
     {
         public static void Initialize()
         {
             PlatformPluginSwitch.DynamicallyLinkAssembly = false;
 
-            PlatformPlugin.InjectDependecies(new WebUIFactory(), new TokenCachePlugin(), new Logger(), new PlatformInformation(), new CryptographyHelper(), new DeviceAuthHelper(), new BrokerHelper());
+            PlatformPlugin.InjectDependecies(
```
mjcheetham: iOS does not allow dynamic loading of executables. Updated this dependency loading mechanism to pass an instance of the new `IWebProxyProvider` plugin.

src/ADAL.PCL/Http/HttpMessageHandlerFactory.cs
```diff
@@ -42,10 +39,10 @@ internal static HttpMessageHandler GetMessageHandler(bool useDefaultCredentials)
                 return MockHandlerQueue.Dequeue();
             }
 
-            return new HttpClientHandler { UseDefaultCredentials = useDefaultCredentials, Proxy = System.Net.WebRequest.DefaultWebProxy };
+            return new HttpClientHandler { UseDefaultCredentials = useDefaultCredentials, Proxy = PlatformPlugin.WebProxyProvider.GetDefaultWebProxy() };
         }
 
-        private readonly static Queue<HttpMessageHandler> MockHandlerQueue = new Queue<HttpMessageHandler>();
+        private static readonly Queue<HttpMessageHandler> MockHandlerQueue = new Queue<HttpMessageHandler>();
```
mjcheetham: Minor reordering of modifers for consistency :)

HSAR:
Seems AppVeyor doesn't agree with your work.

mjcheetham:
@HSAR, yes.. it won't do since it now requires MSBuild 15 / VS2017 to build.

Issue #598, opened 2017-02-10
ClientAssertionCertificate not available in NETStandard 1.6.1


dimkanewtown:
Is there a timeframe when ClientAssertionCertificate will be available on NETStandard?
I'm currently on
        "Microsoft.IdentityModel.Clients.ActiveDirectory": "3.13.8",
        "NETStandard.Library": "1.6.1",

Thanks.

kpanwar:
There is no known timeframe yet.

StummeJ:
I've got a solution for this.  I'll work on this and make a PR :)

kpanwar:
Can you share your solution?

StummeJ:
By using the `Microsoft.IdentityModel.Tokens` it has all of the classes and crypto needed to implement a core version of the `ClientAssertionCertificate` class.  Then that can be used in the `Microsoft.Extensions.Configuration.AzureKeyVault` library to allow access to AzureKeyVaults in dotnet core.  This is our use case, but I cannot supply the code directly from that project.  I want to implement it here and then once it's released update the `Microsoft.Extensions.Configuration.AzureKeyVault` library to support cert based auth in core.

StummeJ:
Is there a branch for .NETStandard code?

StummeJ:
https://github.com/aspnet/Configuration/pull/606 contains the `ClientAssertionCertificateCore` class that is needed to get it to work on dotnet core.

kpanwar:
Your solution is not really going to work if you are planning to add it in ADAL itself. The PR that you shared only works when ICllientCertificateAssertion is passed from the outside. The "Netstandard" binary that you see in the nuget is actually just a PCL. There was no tooling available to have a netstandard library reference a PCL before. This has changed. The correct change IMO, would require to update the platform specific project (ADAL.PCL.CoreCLR) to netstandard, and add the code as needed. This would also require a massive change to our internal build system.

StummeJ:
Is it possible to start using an open build server like AppVeyor or TravisCI?  I'd be happy to start porting this over in my free time.

kpanwar:
It won't be possible and TravisCI does not support xamarin builds. It still would require a massive change to build process. We are putting a plan in place to get this supported in Adal and cannot commit to a date publicly, unfortunately.

StummeJ:
Ahh, AppVeyor would still work for that as they have xamarin enabled build nodes, and @aspnet is using it too.

[...]

Issue #643, opened 2017-03-31
Add System.Runtime.Serialization.Primitives as NuGet dependency


tomkerkhove:
When pulling in the `Microsoft.IdentityModel.Clients.ActiveDirectory` NuGet package it doesn't have `System.Runtime.Serialization.Primitives` as a dependency.

Unfortunately when you run it it starts complaining about the lack of that dependency:
```'Microsoft.IdentityModel.Clients.ActiveDirectory.Authenticator' threw an exception.
  ----> System.IO.FileNotFoundException : Could not load file or assembly 'System.Runtime.Serialization.Primitives, Version=4.1.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.
  ----> System.IO.FileNotFoundException : Could not load file or assembly 'System.Runtime.Serialization.Primitives, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.
   at Microsoft.IdentityModel.Clients.ActiveDirectory.Authenticator.<UpdateFromTemplateAsync>d__44.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.IdentityModel.Clients.ActiveDirectory.AcquireTokenHandlerBase.<PreRunAsync>d__60.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.IdentityModel.Clients.ActiveDirectory.AcquireTokenHandlerBase.<RunAsync>d__55.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #670, opened 2017-05-19
Add X509-based client assertion support to CoreCLR project

mjcheetham:
- Implement methods on `CryptographyHelper` and add a copy of the `ClientAssertionCertificate` type from the _Desktop_ project in the _CoreCLR_ project.
- Add a sample CoreCLR console app which exercises the auth flow using an X509 certificate.
- Update `ClientAssertionWithX509Test` to perform validation of the JWT sent with the request.

**NB:** CoreCLR project now targets .NET Standard 1.3 (previously 1.1) as the various cryptographic types are not available before that verison. This PR is for issue #598


src/ADAL.PCL.CoreCLR/ClientAssertionCertificate.cs
```diff
@@ -0,0 +1,107 @@
...
+using System;
+using System.Globalization;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory
+{
+    /// <summary>
+    /// Containing certificate used to create client assertion.
+    /// </summary>
+    public sealed class ClientAssertionCertificate : IClientAssertionCertificate
```
mjcheetham: We need must use a copy of this file because the way to access private/public keys from X509 certificates is different between NetFx and CoreFx.
Furthermore we cannot extract these differences into a platform interface (say `ICryptographHelper`) because that would require using types (`AsymmetricSecurityKey`, `X509Certificate2`) on an interface in the _ADAL.PCL_ library. These types are _not_ available in .NET Standard 1.1. Chaning the platform-agnostic assembly to target .NET Standard 1.3 would mean removing the ability to run against .NET Framework 4.5.

src/ADAL.PCL.CoreCLR/CryptographyHelper.cs
```diff
@@ -25,14 +25,46 @@
...
+
+            if (certificate == null)
+            {
+                throw new ArgumentNullException(nameof(certificate));
+            }
+
+            var privateKey = certificate.GetRSAPrivateKey();
+            var messageBytes = Encoding.UTF8.GetBytes(message);
+
+            return privateKey.SignData(messageBytes, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
```
mjcheetham: The reason behind my choice of using the PKCS#1 padding scheme here is to be consistent with [MSAL](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/7fe9410913ff277fe1045e65db8066158720a208/src/Microsoft.Identity.Client/Platforms/netstandard1.3/CryptographyHelper.cs#L68).

HSAR: This might be slightly better served as a code comment, for longveity.

mjcheetham: Added comment.

tests/Test.ADAL.NET.Unit/Test.ADAL.NET.Unit.csproj
```diff
@@ -59,6 +59,10 @@
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.Data" />
+    <Reference Include="System.IdentityModel" />
+    <Reference Include="System.IdentityModel.Tokens.Jwt, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
```
kpanwar: why are we referencing JwtTokenHandler for the test project? More specifically, what is being tested?

kpanwar:
@mjcheetham - check out https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/643

tests/Test.ADAL.NET.Unit/Test.ADAL.NET.Unit.csproj
mjcheetham: I realised that the test `ClientAssertionWithX509Test` was not actually validating that the request crafted by ADAL contained a valid JWT (i.e., what if the certificate was used incorrectly?).
I added a new property to the `MockHttpMessageHandler` to allow tests to get access to the raw request and perform any additional validation.
Specifically for `ClientAssertionWithX509Test` I'm inspecting the request body to extract the "client_assertion" and then am using `JwtTokenHandler.CanReadToken` and `JwtSecurityToken` (see file _TokenHandler.cs_) to validate relevant parts of the header and payload.
I _could_ parse the JWT (unbase64, split '.' etc...) but I thought it easier to use the `System.IdentityModel.Tokens.Jwt` package (and since this is only the test project it doesn't impact the actual ADAL library's dependencies).

I also realise that in ADAL we also have JWTHeader and JWTPayload types but didn't want to use these to validate in the test since they're what's being used to craft the JWT in the first place 😃.

src/ADAL.PCL.CoreCLR/CryptographyHelper.cs
```diff
@@ -25,14 +25,48 @@
...
+            {
+                throw new ArgumentNullException(nameof(certificate));
+            }
+
+            var privateKey = certificate.GetRSAPrivateKey();
+            var messageBytes = Encoding.UTF8.GetBytes(message);
+
+            // Using PKCS#1 padding scheme – same as in MSAL:
+            // https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/7fe94109/src/Microsoft.Identity.Client/Platforms/netstandard1.3/CryptographyHelper.cs#L68
+            return privateKey.SignData(messageBytes, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
```
duncanpMS: This version doesn't dispose of the key, whereas the original version does.

As a rule of thumb, if you're copying existing code then it's generally a good idea to make as few changes as possible, particularly if they are stylistic or layout changes. It makes maintenance simpler in the future (easier when diff-ing the files to see what the material changes are).

mjcheetham: Good spot. Changed to be same as MSAL.

PR #678, opened 2017-06-06
Removed references to internal code analysis ruleset

duncanpMS:
The SDL code analysis checks are now being carried out in the VSTS ADAL.Net CI build using the internal SDL task.


duncanpMS:
@mjcheetham re trigger on PRs: there is one for ADAL but it only works for branches in the same repo, not forks (which is why I created the other PR from a branch rather than a fork). I'll set it up for MSAL.Net at some point.

PR #679, opened 2017-06-06
Running WebApp auth on desktop.

RamjotSingh:
Allows applications declared in Web Application registrations to Acquire Token on desktop using ADAL. This allows same app to be used in Web as well as Desktop.

### DIFF
--- a/src/ADAL.PCL/AuthenticationContext.cs
+++ b/src/ADAL.PCL/AuthenticationContext.cs
@@ -479,6 +479,19 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
         /// Acquires security token from the authority.
         /// </summary>
         /// <param name="resource">Identifier of the target resource that is the recipient of the requested token.</param>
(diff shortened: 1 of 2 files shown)


vibronet:
Hi Ramjot,
thanks for the PR.
This scenario is not in scope for ADAL, and presents significant security challenges. The confidential client flow is reserved for pure browser based scenarios, where the code acquiring the tokens (and associated credentials required to perform the acquisition) lives on a server. Making client credentials available on a user-accessible machine would present challenges for key distribution and protection. Furthermore, there are a number of scenarios that apply to native clients only (like CA/device auth) and making it possible to call this method for a confidential client would create confusion and likely hard to troubleshoot issues.
In short, web applications should only be consumed via a browser, and native clients (no matter what technology they are implemented in, including Cordova/React native/Electron or any other stack using HTML/JS) should always be represented in Azure AD as public clients.
Per the above, we can't get this PR in. Thanks again!

RamjotSingh:
Thanks for the clarification :)

PR #682, opened 2017-06-08
Remove redundant file

duncanpMS:
Removed redundant file that was causing a PoliCheck false positive


duncanpMS:
Abandoning - file is required.

PR #693, opened 2017-06-20
Added missing attribute to fix VSTS bug #199584

duncanpMS:
Miscellaneous: improved error messages returned the test app

### DIFF
--- a/samples/TestApp.PCL/TokenBroker.cs
+++ b/samples/TestApp.PCL/TokenBroker.cs
@@ -27,9 +27,7 @@
 
 using Microsoft.IdentityModel.Clients.ActiveDirectory;
 using System;
-using System.Collections.Generic;
 using System.Globalization;
-using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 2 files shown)


samples/TestApp.PCL/TokenBroker.cs
```diff
@@ -75,10 +73,14 @@ public MobileAppSts()
 
     public class TokenBroker
     {
-        private AuthenticationContext context;
 
-        public Sts Sts = new AadSts();
+        private Sts sts = new AadSts();
```
duncanpMS: Minor changes in this file to fix some code analysis issues (all low risk, all in test code)

samples/TestApp.PCL/TokenBroker.cs
```diff
@@ -93,67 +95,84 @@ UserIdentifier GetUserIdentifier()
...
+                return GetErrorMessage(ex);
             }
         }
 
         public void ClearTokenCache()
         {
             TokenCache.DefaultShared.Clear();
         }
+
+        private static string GetErrorMessage(Exception ex)
```
duncanpMS: Diagnosing the test app failure was made harder than it needed to be because the useful error information nested several levels down. This routine creates are more useful error message that appears in the UI.

PR #694, opened 2017-06-21
Added missing Android Preserve attribute to fix #199584

duncanpMS:
Miscellaneous: fixed some simple code analysis issues and improved the error message shown in the test app.

### DIFF
--- a/samples/TestApp.PCL/TokenBroker.cs
+++ b/samples/TestApp.PCL/TokenBroker.cs
@@ -27,9 +27,7 @@
 
 using Microsoft.IdentityModel.Clients.ActiveDirectory;
 using System;
-using System.Collections.Generic;
 using System.Globalization;
-using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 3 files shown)


samples/TestApp.PCL/TokenBroker.cs
```diff
@@ -93,67 +95,84 @@ UserIdentifier GetUserIdentifier()
...
+                return GetErrorMessage(ex);
             }
         }
 
         public void ClearTokenCache()
         {
             TokenCache.DefaultShared.Clear();
         }
+
+        private static string GetErrorMessage(Exception ex)
```
duncanpMS: Improves the error message returned to simplify debugging. The error message now includes all of the messages from any internal exceptions.

src/ADAL.PCL.Android/ADAL.PCL.Android.csproj
```diff
@@ -52,9 +52,7 @@
     <Compile Include="..\ADAL.Common\CommonAssemblyInfo.cs">
       <Link>CommonAssemblyInfo.cs</Link>
     </Compile>
-    <Compile Include="..\ADAL.Common\WebProxyProvider.cs">
-      <Link>WebProxyProvider.cs</Link>
```
duncanpMS: Previously the Android assembly was using a shared copy of the class. However, it needs to have an Android-specific attribute applied so I've created a copy in the Android platform project.

PR #716, opened 2017-07-14
Fix #611: implement stricter auth header parsing

### LINKED ISSUE #611: Exceptions are not being thrown for some invalid authentication headers
_AuthenticationParameters.CreateFromResponseAuthenticateHeader_  does a partial validation of the supplied _authenticateHeader_ string and throws for input input. However, it does not throw for two of the cases in _UnitTests::AuthenticationParametersTest_:

```csharp
RunAuthenticationParametersNegative("Bearer authorization_uri=\"abc\"=\"de\"");
RunAuthenticationParametersNegative("Bearer authorization_uri=abc=de");
```
(the test isn't failing, but that's due to a separate bug in the test for which there is a fix available).

mjcheetham:
Throw exception when the authorisation header contains a malformed key-value pair string which isn't in the format: "X=Y" or "X".

I.e., "X=Y=Z" is now invalid.

### DIFF
--- a/src/ADAL.PCL/Utilities/EncodingHelper.cs
+++ b/src/ADAL.PCL/Utilities/EncodingHelper.cs
@@ -39,7 +39,12 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     /// </summary>
     internal static class EncodingHelper
     {
-
+        /// <summary>
+        /// URL encode the given string.
+        /// </summary>
+        /// <param name="message">String to URL encode</param>
+        /// <returns>URL encoded string</returns>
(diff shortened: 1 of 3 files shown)


src/ADAL.PCL/Utilities/EncodingHelper.cs
```diff
@@ -39,7 +39,12 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     /// </summary>
     internal static class EncodingHelper
     {
-
+        /// <summary>
```
mjcheetham: I've also added some comments to some of the other methods in `EncodingHelper`.

duncanpMS:
LGTM

PR #718, opened 2017-07-17
Fix #717: Make TokenCacheKey.GetHashCode not culture dependent

### LINKED ISSUE #717: TokenCache deserialization fails with Duplicate Key issues with Turkish culture
When we use the serialization and deserialization mechanisms ADAL provides for the `TokenCache`, we hit `System.ArgumentException: 'The key already existed in the dictionary.'` when deserializing the cache when in a Turkish culture.

Minimal repro example:
```csharp
// Repro instructions
// 1) (Not sure if this step is necessary, but I did it) Switch browser's language to Turkish
// 2) Create new user in the AAD portal that contains a capital I in the username/email field
//    (eg, testI@mytenant.onmicrosoft.com)
// 3) Fill the first code line here with the tenant's UPN, and fill out relevant client details
var userIdentifier = new UserIdentifier("testI@TENANTHERE.onmicrosoft.com", UserIdentifierType.RequiredDisplayableId);

var clientId = "CLIENT_ID_HERE"; 
var resource = "RESOURCE_HERE"; 
var redirectUri = new Uri("urn:ietf:wg:oauth:2.0:oob"); // Out-Of-Bound redirect uri
var authority = "https://login.windows.net/common/";

// Set the Turkish culture
Thread.CurrentThread.CurrentCulture = new CultureInfo("tr-TR");

var platformParams = new PlatformParameters(PromptBehavior.Always);
var originalCache = new TokenCache();
var authCtx = new AuthenticationContext(authority, originalCache);

// Grab a token with a test AAD user
var firstAuthResult = authCtx.AcquireTokenAsync(resource, clientId, redirectUri, platformParams).Result;

// At this point, the internal dictionary contains a single item.  It is succesfully found with the below
var secondAuthResult = authCtx.AcquireTokenSilentAsync(resource, clientId, userIdentifier).Result;

// Now, serialize and deserialize the token cache, and use a fresh AuthenticationContext
byte[] serializedCache = originalCache.Serialize();
TokenCache deserializedCache = new TokenCache(serializedCache);
var authCtxFromDeserializedCache = new AuthenticationContext(authority, deserializedCache);

// There is only one item in the cache *before* the call below, but afterwards, we will have two entries which 
// *appear* to be identical, but have hashed into different buckets in the internal dictionary
var thridAuthResult = authCtxFromDeserializedCache.AcquireTokenSilentAsync(resource, clientId, userIdentifier).Result;

// When the cache is serialized and deserialized again, it fails in the deserialization step, 
// throwing a System.ArgumentException: 'The key already existed in the dictionary.'
byte[] secondSerializedCache = deserializedCache.Serialize();
TokenCache secondDeserializedCache = new TokenCache(secondSerializedCache);
```

mjcheetham:
Change `TokenCacheKey.GetHashCode` to ensure that:

1. the result is *not* dependent on the current thread's culture,
1. the result is *not* dependent on mutable properties.

### DIFF
--- a/src/ADAL.PCL/Cache/TokenCacheKey.cs
+++ b/src/ADAL.PCL/Cache/TokenCacheKey.cs
@@ -68,17 +68,17 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
             this.DisplayableId = displayableId;
         }
 
-        public string Authority { get; private set; }
+        public string Authority { get; }
 
-        public string Resource { get; internal set; }
+        public string Resource { get; }
 
-        public string ClientId { get; private set; }
+        public string ClientId { get; }
 
-        public string UniqueId { get; private set; }
+        public string UniqueId { get; }
 
-        public string DisplayableId { get; private set; }
+        public string DisplayableId { get; }
 
-        public TokenSubjectType TokenSubjectType { get; private set; }
+        public TokenSubjectType TokenSubjectType { get; }
 
         /// <summary>
         /// Determines whether the specified object is equal to the current object.
@@ -102,14 +102,18 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
         /// <param name="other">The TokenCacheKey to compare with the current object. </param><filterpriority>2</filterpriority>
         public bool Equals(TokenCacheKey other)
         {
-            return ReferenceEquals(this, other) ||
-               (other != null
-               && (other.Authority == this.Authority)
-               && this.ResourceEquals(other.Resource)
-               && this.ClientIdEquals(other.ClientId)
-               && (other.UniqueId == this.UniqueId)
-               && this.DisplayableIdEquals(other.DisplayableId)
-               && (other.TokenSubjectType == this.TokenSubjectType));
+            if (ReferenceEquals(this, other))
+            {
+                return true;
+            }
+
+            return other != null
+                && other.Authority == this.Authority
+                && this.ResourceEquals(other.Resource)
+                && this.ClientIdEquals(other.ClientId)
+                && other.UniqueId == this.UniqueId
+                && this.DisplayableIdEquals(other.DisplayableId)
+                && other.TokenSubjectType == this.TokenSubjectType;
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/ADAL.PCL/Cache/TokenCacheKey.cs
```diff
@@ -68,17 +68,17 @@ internal TokenCacheKey(string authority, string resource, string clientId, Token
             this.DisplayableId = displayableId;
         }
 
-        public string Authority { get; private set; }
```
mjcheetham: I've made these properties read-only since they're used in the `GetHashCode` method.
Hash codes should _not_ be based on mutable fields.

src/ADAL.PCL/Cache/TokenCacheKey.cs
```diff
@@ -120,13 +124,14 @@ public bool Equals(TokenCacheKey other)
         /// </returns>
         public override int GetHashCode()
         {
-            const string Delimiter = ":::";
-            return (this.Authority + Delimiter 
-                + this.Resource.ToLower() + Delimiter
-                + this.ClientId.ToLower() + Delimiter
-                + this.UniqueId + Delimiter
-                + ((this.DisplayableId != null) ? this.DisplayableId.ToLower() : null) + Delimiter
```
mjcheetham: This is the culprit; `DisplayableId.ToLower()` rather than `DisplayableId.ToLowerInvariant()`.

src/ADAL.PCL/TokenCache.cs
```diff
@@ -163,7 +163,7 @@ public byte[] Serialize()
                     writer.Write(this.tokenCacheDictionary.Count);
                     foreach (KeyValuePair<TokenCacheKey, AuthenticationResultEx> kvp in this.tokenCacheDictionary)
                     {
-                        writer.Write(string.Format(CultureInfo.CurrentCulture, "{1}{0}{2}{0}{3}{0}{4}", Delimiter,
```
mjcheetham: Although not *technically* an issue (since we're only using this as a simple string-concat and not invoking any formating options), changing this to use the correct culture (the invariant one).
What we lay down on disk should not depend on the current culture as this might change between serialisation and deserialisation.

tests/Test.ADAL.NET.Unit/TokenCacheKeyTests.cs
```diff
@@ -0,0 +1,195 @@
...
+        private const TokenSubjectType SubjectType = TokenSubjectType.User;
+        private const string UniqueId = "27f324f9-4164-4a1e-b778-e13e06606127";
+        private const string DisplayableId = "A displayable ID";
+
+        #region Tests
+
+        [TestMethod]
+        [Description("Test TokenCacheKey.Equals with different instances but same internal values")]
+        [TestCategory("AdalDotNetUnit")]
+        public void TokenCacheKey_Equals_SameValuesDifferenceInstancesAreEqual()
```
mjcheetham: Added general coverage of both `Equals` and `GetHashCode` as this was missing.

tests/Test.ADAL.NET.Unit/TokenCacheKeyTests.cs
```diff
@@ -0,0 +1,195 @@
...
+                // Verify
+                Assert.AreNotEqual(baseCode, other, "Expected hash codes NOT to be equal.");
+            }
+        }
+
+
+        [TestMethod]
+        [Description("Test TokenCacheKey.GetHashCode is not culture dependent")]
+        [TestCategory("AdalDotNetUnit")]
+        public void TokenCacheKey_GetHashCode_IsNotCurrentCultureDependent()
```
mjcheetham: Added a test specifically to check culture-dependency of `GetHashCode`.

mjcheetham:
Closed and reopened to trigger another AppVeyor build.

Issue #710, opened 2017-07-10
System.ArgumentOutOfRange exception inside ADAL library


savy-g:
Write Ideas ios app app uses ADAL for authentication. On June 3rd, we observed some crash incidents of the app.
The attached image contains the stack trace. It suggests that it has something to do with the logging logic of ADAL when trying to convert b/w date formats.

ADAL library version we're using is 3.13.8.

![stack trace](https://user-images.githubusercontent.com/24301049/28033677-df04ce3c-6563-11e7-8585-0e0fe6a72245.png)

This issue has blocked the release of our Ios app "Microsoft Write Ideas" on the apple store.

savy-g:
Needs to be escalated because this has blovked a new release of our app on apple store.

savy-g:
![image](https://user-images.githubusercontent.com/24301049/28332717-9fb03eea-6baa-11e7-83f9-3736b3798b00.png)

mjcheetham:
Hi @savy-g,

Do you know which particular culture was the current culture during this exception?

Also, when you're building the app are you including the extra internationalisation settings?
https://developer.xamarin.com/guides/ios/advanced_topics/localization_and_internationalization/encodings/
![image](https://user-images.githubusercontent.com/5658207/28368107-27693c16-6c8b-11e7-818f-66a28bc9b963.png)

savy-g:
@mjcheetham ,
 I just checked the build settings of our app.
We are not using the extra internationalization settings.
Will using them solve the problem?

mjcheetham:
@savy-g, without being able to fully reproduce the problem I can't say this will fix it for sure.

The discussion on the Xamarin forums (linked above), and a [bug](https://bugzilla.xamarin.com/show_bug.cgi?id=30897) about the same exception being thrown when using the Thai locale, leads me to believe these settings _might_ cause this issue.

On the ADAL side, we're calling:

```cs
string.Format(CultureInfo.CurrentCulture, "{0} ...removed for brevity..", DateTime.UtcNow);
```

..which is resulting in the exception being thrown from within Mono types when it's getting the culture-specific string representation of the `DateTime`.

mjcheetham:
@savy-g, we've updated ADAL so that it now always uses both the Invariant Culture and the ISO format for date-times when creating log messages. I also checked that we don't do culture-dependent date-time formatting anywhere else in ADAL.

[...]

PR #719, opened 2017-07-18
Correct usage of string functions; case and culture

mjcheetham:
Correct usage of various string functions such as `Format`, `Equals`, `Compare`, and `StartsWith` to ensure the correct case and culture considerations are taken in to account.

All URL or request/response handling should be invariant/ordinal, error messages should be culture-specific.

Edit:
I've only changed string handling for "internal" operations, i.e., ones that are acting on URLs or tokens. I've reverted any changes made to string handling on error/exception messages as discussed offline.

### DIFF
--- a/src/ADAL.PCL.Android/BrokerProxy.cs
+++ b/src/ADAL.PCL.Android/BrokerProxy.cs
@@ -107,7 +107,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
                 Application.Context.PackageManager.CheckPermission(permission, Application.Context.PackageName))
             {
                 PlatformPlugin.Logger.Information(null,
-                    string.Format(CultureInfo.CurrentCulture, AdalErrorMessageAndroidEx.MissingPackagePermissionTemplate, permission));
+                    string.Format(AdalErrorMessageAndroidEx.MissingPackagePermissionTemplate, permission));
                 return false;
             }
 
--- a/src/ADAL.PCL.Android/BrokerProxy.cs
+++ b/src/ADAL.PCL.Android/BrokerProxy.cs
@@ -107,7 +107,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
                 Application.Context.PackageManager.CheckPermission(permission, Application.Context.PackageName))
             {
                 PlatformPlugin.Logger.Information(null,
-                    string.Format(AdalErrorMessageAndroidEx.MissingPackagePermissionTemplate, permission));
+                    string.Format(CultureInfo.CurrentCulture, AdalErrorMessageAndroidEx.MissingPackagePermissionTemplate, permission));
                 return false;
             }
 
@@ -371,7 +371,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
             string signatureDigest = this.GetCurrentSignatureForPackage(packageName);
             if (!string.IsNullOrEmpty(signatureDigest))
             {
-                return string.Format(CultureInfo.CurrentCulture, "{0}://{1}/{2}", RedirectUriScheme,
+                return string.Format(CultureInfo.InvariantCulture, "{0}://{1}/{2}", RedirectUriScheme,
                     packageName.ToLower(), signatureDigest);
             }
 
@@ -602,7 +602,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
                             if (chainStatus.Status != X509ChainStatusFlags.UntrustedRoot)
                             {
                                 throw new AdalException(AdalErrorAndroidEx.SignatureVerificationFailed,
-                                    string.Format(CultureInfo.InvariantCulture,
+                                    string.Format(CultureInfo.CurrentCulture,
                                         "app certificate validation failed with {0}", chainStatus.Status));
                             }
                         }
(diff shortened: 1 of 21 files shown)


src/ADAL.PCL.Android/BrokerProxy.cs
```diff
@@ -371,7 +371,7 @@ private string GetRedirectUriForBroker()
             string signatureDigest = this.GetCurrentSignatureForPackage(packageName);
             if (!string.IsNullOrEmpty(signatureDigest))
             {
-                return string.Format(CultureInfo.CurrentCulture, "{0}://{1}/{2}", RedirectUriScheme,
+                return string.Format(CultureInfo.InvariantCulture, "{0}://{1}/{2}", RedirectUriScheme,
                     packageName.ToLower(), signatureDigest);
```
mjcheetham: Not sure what the rules about case-sensitivity are for Android package names. Leaving this.

src/ADAL.PCL.Desktop/WindowsFormsWebAuthenticationDialogBase.cs
```diff
@@ -152,9 +156,16 @@ protected virtual void WebBrowserNavigatingHandler(object sender, WebBrowserNavi
 
             // check if the url scheme is of type browser://
             // this means we need to launch external browser
-            if (!e.Cancel && e.Url.Scheme.Equals("browser", StringComparison.CurrentCultureIgnoreCase))
+            if (!e.Cancel && e.Url.Scheme.Equals(BrowserScheme, StringComparison.OrdinalIgnoreCase))
             {
-                Process.Start(e.Url.AbsoluteUri.Replace("browser://", "https://"));
+                // Build the HTTPS URL for launching with an external browser
+                var httpsUrlBuilder = new UriBuilder(e.Url)
```
mjcheetham: Using `System.UriBuilder` to change the scheme from `browser://` to `https://` instead of string-replacement (which was previously not doing so ignoring case).

src/ADAL.PCL.Android/BrokerProxy.cs
HSAR: Assuming Android kept true to its Java origins, it should be case-sensentive.

duncanpMS:
Closed and re-opened pull request to re-trigger an AppVeyor build. The first AppVeyor build failed due to a timeout fetching the required packages.

mjcheetham:
Build is failing due to being unable to restore NuGet packages. Not sure what's up.

src/ADAL.PCL.Android/BrokerProxy.cs
mjcheetham: Yeah. This seems suspect to me, but must be working? I'll leave it.

mjcheetham:
There might be some of these changes which are more impactful that we might want to revert for now. Call any out in comments if you feel so.

src/ADAL.PCL.Desktop/WindowsFormsWebAuthenticationDialogBase.cs
```diff
@@ -217,19 +228,19 @@ private bool CheckForClosingUrl(Uri url)
...
             // if redirect_uri is not hit and scheme of the url is not whitelisted and url is not about:blank
             // and url scheme is not https then fail to load.
-            if (!canClose && !whiteListedSchemes.Contains(url.Scheme.ToLower(CultureInfo.CurrentCulture)) &&
-                !url.AbsoluteUri.Equals("about:blank", StringComparison.CurrentCultureIgnoreCase)
-                && !url.Scheme.Equals("https", StringComparison.CurrentCultureIgnoreCase) 
-                && !url.Scheme.Equals("javascript", StringComparison.CurrentCultureIgnoreCase))
+            if (!canClose && !WhiteListedSchemes.Contains(url.Scheme)
+                && !url.AbsoluteUri.Equals(AboutBlankUri, StringComparison.OrdinalIgnoreCase)
+                && !url.Scheme.Equals(Uri.UriSchemeHttps, StringComparison.OrdinalIgnoreCase)
+                && !url.Scheme.Equals(JavaScriptScheme, StringComparison.OrdinalIgnoreCase))
```
duncanpMS: Not something you've changed, but it seems odd that there are explicit checks for the _https_ and _javascript_ schemes rather than just whitelisting them.

mjcheetham: @duncanpMS true. Do you think I should change this?

duncanpMS: No, it's not urgent and we don't have any test coverage.

mjcheetham:
After discussing offline, I've reverted any changes made to culture string-formatting when used for error/exception messages. Even though these are using the wrong culture value, consumers of ADAL may have taken a dependency on the exact wording of the error messages for handling specific error cases (since we usually return only `AdalException` or `AdalServiceException` types for all errors that users might care about).
Also we're lacking confidence in test coverage.

Keeping the changes to things such as: token or URL string concatenation and request messages, as these really should not change depending on the user's culture.

mjcheetham:
@duncanpMS please could you double-check the remaining changes after I've reverted some. Thanks.

kpanwar:
@mjcheetham @duncanpMS - Please do change the messages. We provide error codes for them to take the dependency. Error messages are purely informational and can be changed any time. Just like server changes error_description as requirements change.

mjcheetham:
Closing and reopening to force a new AppVeyor build.
This keeps happening (the package restore keeps timing out). Not sure why…

duncanpMS:
LGTM

mjcheetham:
@kpanwar, good to know that in theory we should be able to change the error messages 😃.
However, I shall complete this PR without the changes to the exception string formats for now. There's an issue (#710) that's arisen to do with some issues with formatting strings based on the current culture (on Xamarin platforms). Probably requires some more thought.

PR #722, opened 2017-07-20
Use ISO 8601 format for logger timestamps

mjcheetham:
Explicitly use the ISO 8601 format for date/times when generating log messages. Also ensure we're using the Invariant Culture when constructing the final string (we're only using string.Format here to do concatenation).

This should hopefully also fix issue #710.


mjcheetham:
I also checked for any other instances where we were doing a `string.Format` using a `DateTime` where we were not specifying an explicit format and using the Invariant Culture.

PR #582, opened 2016-12-28
Implement PromptBehavior to iOS/Android

yamamoWorks:
Implement PromptBehavior to iOS/Android.
Add PromptBehavior Property to PlatformParameters.
Implement GetCacheLoadPolicy() and
AddPromptBehaviorQueryParamete() in PlatformInformation.
Move PromptBehavior Enum to PCL.

PR #723, opened 2017-07-20
Implement PromptBehavior to iOS/Android

mjcheetham:
Based on #582 with the following changes:

- Merged in the latest changes from the _dev_ branch
- Remove ability to specify `PromptBehavior.Never` on iOS and Android; we do not support this
- Minor changes to tidy up some code in the various `PlatformInformation` types


src/ADAL.Common/PromptBehavior.Xamarin.cs
```diff
@@ -0,0 +1,57 @@
...
+        Always,
+
+        /// <summary>
+        /// Re-authorizes (through displaying webview) the resource usage, making sure that the resulting access
+        /// token contains updated claims. If user logon cookies are available, the user will not be asked for 
+        /// credentials again and the logon dialog will dismiss automatically.
+        /// </summary>
+        RefreshSession
+
+        // NB: We do not support the "Never" option on Xamarin iOS or Android.
```
mjcheetham: This version of the `PromptBehavior` enumeration is for the Xamarin platforms only. It does not contain a "Never" member as this is not supported on those platforms.

PR #725, opened 2017-07-21
Assert inconclusive if culture not found in tests

mjcheetham:
The Jenkins build agent (and other possible machines) does not have the required set-up to switch to the Turkish culture. Assert inconclusive in this case.


duncanpMS:
LGTM

PR #729, opened 2017-07-26
Simplified NuGet packaging code

duncanpMS:
The new VSTS signing tasks update the binaries in place, so there is no need to copy the files to a location for signing, then copy them to the NuGet layout location. Instead, the files can just be copied directly from the build location to the NuGet layout folder.

This commit contains an MSBuild project that does the copying, rather than using a script. There are a couple of advantages:
* the project is more declarative and easier to read
* the project is self-contained; it doesn't depend on other scripts having been run first e.g. to initialize variables
* it can be run locally on a dev machine. This simplifies debugging and testing the project, but also makes it much simpler to create a NuGet package locally.
* the project will fail if required files cannot be found. The scripted approach copies files using wildcards so it won't fail if the files are not present, but you end up with an improperly constructed package.

The commit also contains a second reusable targets files that checks that assemblies are strong-named and Authenticode signed. These checks can be run optionally as part of a build.

Relates to VSTS work item 194064: NuGet package is created as part of the build


.gitignore
```diff
@@ -13,3 +13,4 @@ src/ADAL.PCL.Android/ADAL.PCL.Android.csproj.bak
 AppPackages
 Microsoft.IdentityModel.Clients.ActiveDirectory.XML
 *.htm
+.nugetPackageRoot/
```
duncanpMS: By default, the targets will create the NuGet folder layout under this folder (although you can specify an alternative location by setting the appropriate build property).

build/Microsoft.IdentityModel.Clients.ActiveDirectory.nuspec
```diff
@@ -18,8 +18,8 @@
     <dependencies>
 		<group targetFramework="netstandard1.4">
 			<dependency id="NETStandard.Library" version="1.6" />
-			<dependency id="System.Runtime.Serialization.Json" version="4.0.2" />
-			<dependency id="System.Runtime.Serialization.Primitives" version="4.1.1" />
+			<dependency id="System.Runtime.Serialization.Json" version="4.3.0" />
```
duncanpMS: Miscellaneous: the hard-coded nuspec file was out of date: the ADAL.Net project actually references these version.

build/Microsoft.IdentityModel.Clients.ActiveDirectory.nuspec
```diff
@@ -3,7 +3,7 @@
     <metadata minClientVersion="2.8.6">
         <title>Active Directory Authentication Library</title>
         <id>Microsoft.IdentityModel.Clients.ActiveDirectory</id>
-        <version>REPLACE_DURING_BUILD</version>
+        <version>0.0.1-LOCALBUILD</version>
```
duncanpMS: Changed the _version_ to a valid value so the project can be built locally without setting the version explicitly.

build/prepare-versioning.ps1
```diff
@@ -35,9 +35,9 @@ $nugetVersion = "{0}.{1}.{2}" -f ($versionTokens[0], $versionTokens[1], $version
 
 Write-Host "=========================="
 Write-Host "Versioning .nuspec file..."
-$filename = $env:TO_PACK_TARGET + "\" + $env:LIBRARY_NAME + ".nuspec"
+$filename = "build\Microsoft.IdentityModel.Clients.ActiveDirectory.nuspec"
```
duncanpMS: Update the source _nuspec_ file in place (the script updates the common properties file in-place the same way).
The reason for this change (apart from making the script internally consistent) is that it makes it self-contained - it no longer depends on having external variables set.

build/prepare-versioning.ps1
```diff
@@ -35,9 +35,9 @@ $nugetVersion = "{0}.{1}.{2}" -f ($versionTokens[0], $versionTokens[1], $version
...
 $content = Get-Content $filename
-$newContent = $content -replace "REPLACE_DURING_BUILD", $nugetVersion
+$newContent = $content -replace "<version>(.*)</version>", "<version>$nugetVersion</version>"
```
duncanpMS: The _version_ in the hard-coded _nuspec_ file was modified about. This change is required to the versioning script so it correctly updates the file.

build/LayoutNugetPackage.csproj
```diff
@@ -0,0 +1,134 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Layout">
```
duncanpMS: The NuGet layout can be produced locally just by building this project i.e. _msbuild build\LayoutNuGetPackage.csproj_
I'll change the project file extension to _.proj_ as it isn't a C# project.

HSAR:
We are still calling "nuget pack" seperately, right? This only lays out the files in the right location?

duncanpMS:
Yes.

PR #734, opened 2017-08-01
Address some globalisation issues in WinRT, iOS and Android platform assemblies

mjcheetham:
There are some more FxCop violations regarding globalisation rules (use of string methods without explicit regard for culture) present in the WinRT, iOS, and Android platform assemblies.

Correcting the issues in this PR.


mjcheetham:
I'm using the invariant culture in both error/exception messages as well as string-concat type usages.
Should be using the current culture for error/exception messages, but we avoided doing this previously in the core and desktop assemblies due to some other issues we encountered, and the possibility of other such issues cropping up later.

PR #741, opened 2017-08-03
Add SonarLint analysers to the solution, bound to internal SonarQube server

mjcheetham:
Adding the SonarLint Roslyn analysers to the projects. This means you should see the same* issues as are reported on the VSTS builds within the IDE itself.

I've fixed a few of the warnings (such as making fields `readonly`), and suppressed a few false-positives.
I've not made any changes which are more than "low hanging fruit" fixes.

*some of the parameterised rules (such as maximum number of parameters in a method allowed) can only be set with their default values, not the one customised on the server.


duncanpMS:
LGTM.
Perhaps delay committing this if there is going to be a new imminent release to fix the API issue.

mjcheetham:
@duncanpMS agreed. I'll leave this PR here until things are settled in the _dev_ branch.

PR #742, opened 2017-08-03
Fixed WinRT failure when logging an empty UserInfo object.

### LINKED ISSUE #727: UWP App AccquireTokenAsync(resource, clientCredential) throws exception on second run
This is easy to reproduce:

Visual Studio 2017 => New UWP App
References:
Microsoft.IdentityModel.Clients.ActiveDirectory (3.14.2)
Microsoft.NETCore.UniversalWindowsPlatform (5.3.3)

In MainPage() I put standard code to get token via client secret:

```
var authority = "https://login.microsoftonline.com/myTenant";
var authContext = new AuthenticationContext(authority);
ClientCredential credential = new ClientCredential(myClientId, myClientSecret);
var result = authContext.AcquireTokenAsync(myResource, credential).GetAwaiter().GetResult();
```

When you run the app for the first time this works fine, you get a result and it has the token as expected. Running the app a second time results in this exception:

```
System.ArgumentNullException occurred
  HResult=0x80004003
  Message=Value cannot be null.
  Source=<Cannot evaluate the exception source>
  StackTrace:
   at System.StubHelpers.HStringMarshaler.ConvertToNativeReference(String managed, HSTRING_HEADER* hstringHeader)
   at Windows.Security.Cryptography.CryptographicBuffer.ConvertStringToBinary(String value, BinaryStringEncoding encoding)
   at Microsoft.IdentityModel.Clients.ActiveDirectory.CryptographyHelper.CreateSha256Hash(String input)
   at Microsoft.IdentityModel.Clients.ActiveDirectory.AcquireTokenHandlerBase.LogReturnedToken(AuthenticationResult result)
   at Microsoft.IdentityModel.Clients.ActiveDirectory.AcquireTokenHandlerBase.PostRunAsync(AuthenticationResult result)
   at Microsoft.IdentityModel.Clients.ActiveDirectory.AcquireTokenHandlerBase.<RunAsync>d__55.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationContext.<AcquireTokenForClientCommonAsync>d__49.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationContext.<AcquireTokenAsync>d__26.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at AdalTester.MainPage..ctor() in c:\blah\AdalTester\AdalTester\MainPage.xaml.cs:line 32
   at AdalTester.AdalTester_XamlTypeInfo.XamlTypeInfoProvider.Activate_0_MainPage() in c:\blah\AdalTester\AdalTester\obj\x86\Debug\XamlTypeInfo.g.cs:line 178
   at AdalTester.AdalTester_XamlTypeInfo.XamlUserType.ActivateInstance() in c:\blah\AdalTester\AdalTester\obj\x86\Debug\XamlTypeInfo.g.cs:line 334
```

How can I have the client secret flow work in a UWP app. The actual app I need this to work in is a Kiosk type scenario so a user cannot log in, it needs to be able to do this silently.

HSAR:
In `AcquireTokenHandlerBase.cs` line 355, the following line appears:

`result.UserInfo != null
                            ? PlatformPlugin.CryptographyHelper.CreateSha256Hash(result.UserInfo.UniqueId)
                            : "null"`

When the token is first returned from the server, the UserInfo object is null and `CreateSha256Hash()` is not called. However, on subsequent returns the UserInfo object is no longer null, but it is empty (`result.UserInfo.UniqueId == null`). Thus `CreateSha256Hash()` is called with a null argument whenever a token is returned from cache.

`CreateSha256Hash()` is a platform-specific method which is implemented in `CryptographyHelper.cs`. The Desktop, CoreCLR and iOS platform libraries have defensive coding that checks for null strings, but the WinRT and Android libraries do not. Most of the other methods in the WinRT implementation of CryptographyHelper have null checks, but this one does not, and causes an exception to be thrown.

Side note: The above method appears to be not implemented on Android (!!!)

To fix the issue, this PR brings the WinRT CryptographyHelper in line with its other methods and the other implementations by adding the same `string.IsNullOrWhiteSpace(...)` check. In the interest of standardisation, I changed the iOS check to this statement, and added it to the Android implementation too.

This should fix #727.

### DIFF
--- a/src/ADAL.PCL.Android/CryptographyHelper.cs
+++ b/src/ADAL.PCL.Android/CryptographyHelper.cs
@@ -34,6 +34,11 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     {
         public string CreateSha256Hash(string input)
         {
+            if (string.IsNullOrWhiteSpace(input))
+            {
+                return null;
+            }
+
             return string.Empty;    // TODO: Update
         }
         
--- a/src/ADAL.PCL.WinRT/CryptographyHelper.cs
+++ b/src/ADAL.PCL.WinRT/CryptographyHelper.cs
@@ -43,6 +43,11 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 
         public string CreateSha256Hash(string input)
         {
+            if (string.IsNullOrWhiteSpace(input))
+            {
+                return null;
+            }
+            
             IBuffer inputBuffer = CryptographicBuffer.ConvertStringToBinary(input, BinaryStringEncoding.Utf8);
 
             var hasher = HashAlgorithmProvider.OpenAlgorithm("SHA256");
--- a/src/ADAL.PCL.iOS/CryptographyHelper.cs
+++ b/src/ADAL.PCL.iOS/CryptographyHelper.cs
@@ -35,16 +35,16 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     {
         public string CreateSha256Hash(string input)
         {
-            if (input != null)
+            if (string.IsNullOrWhiteSpace(input))
             {
-                using (SHA256Managed sha = new SHA256Managed())
-                {
-                    UTF8Encoding encoding = new UTF8Encoding();
-                    return Convert.ToBase64String(sha.ComputeHash(encoding.GetBytes(input)));
-                }
+                return null;
             }
 
-            return null;
+            using (SHA256Managed sha = new SHA256Managed())
+            {
+                UTF8Encoding encoding = new UTF8Encoding();
+                return Convert.ToBase64String(sha.ComputeHash(encoding.GetBytes(input)));
+            }
         }
     }
 }


src/ADAL.PCL.Android/CryptographyHelper.cs
```diff
@@ -34,6 +34,11 @@ internal class CryptographyHelper : ICryptographyHelper
     {
         public string CreateSha256Hash(string input)
         {
+            if (string.IsNullOrWhiteSpace(input))
+            {
+                return null;
+            }
+
             return string.Empty;    // TODO: Update
```
mjcheetham: As discussed offline, maybe also ask @kpanwar as to why the Android `CryptographyHelper` does not have an implementation (whereas iOS does – i.e., the functionality is supported in Mono-based runtimes).
Probably should be a work item to cover this `//TODO`

HSAR: Done and done. Thanks.

kpanwar: This check exists the way it is because, we do not expect developers to use confidential client on native applications for obvious security reasons. this exception (not thrown by intent) at least prevented the developer to do wrong things :)

PR #750, opened 2017-08-09
Minor tweak to ruleset

duncanpMS:
Disable two rules


jmprieur:
LGTM
Given that this is a legacy library, using interop, [S101: Types should be named in camel case](http://www.sonarlint.org/visualstudio/rules/index.html#sonarLintVersion=3.4.0&ruleId=S101&language=C#) does not make sense, and [S107: Methods should not have too many parameters](http://www.sonarlint.org/visualstudio/rules/index.html#sonarLintVersion=3.4.0&ruleId=S107&language=C#) either as the library has been like this for years now.

PR #767, opened 2017-08-22
Fix NuGet package autolayout for new multi-target build system.

HSAR:


HSAR:
Re-pushing to trigger a new build.

PR #769, opened 2017-08-24
Fixed automation app not parsing snake_cased user_identifier_type.

HSAR:
Example implementation on Objective C, which matches the snake_case system seen in all other values:
https://github.com/AzureAD/azure-activedirectory-library-for-objc/blob/61e6a751fe390f2cf4cab64e17c84a6139378042/ADAL/ADALAutomation/ADAutoMainViewController.m#L158

As `UserIdentifierType.TryParse(...)` was expecting `"OptionalDisplayableId"`, `"optional_displayable"` wasn't being recognised. Because C#'s enum parsing is pretty bad, I'm just using some if/then statements with `string.Equals(...)` calls. We could also have tried platform-specific strings, but this way seems neater.

### DIFF
--- a/automation/WinFormsAutomationApp/AuthenticationHelper.cs
+++ b/automation/WinFormsAutomationApp/AuthenticationHelper.cs
@@ -33,8 +33,23 @@ namespace WinFormsAutomationApp
                 }
                 else if (input.ContainsKey("user_identifier") && input.ContainsKey("user_identifier_type"))
                 {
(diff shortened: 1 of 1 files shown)


HSAR:
Oh - this is a fix for the test `AcquireTokenPromptAlwaysWithHint_Managed_Graph_NotFederated`, which wasn't showing a hint because the user ID type wasn't being passed correctly.

Issue #721, opened 2017-07-18
Obtaining the Retry-After header from ADAL on 429 Too Many Requests errors


tehcrashxor:
In our discussions with AAD STS and Graph teams following last week's throttling of the VSTS and the entire MS tenant, we've been asked to use the value of the `Retry-After` headers they send on 429: Too Many Request errors to know how long to we should wait before a retry. However, ADAL is currently not providing that (nor any other header value) onto the `AdalServiceException`s.

I don't see one, but is there any way we can get those values today?  If not, any thoughts on adding this functionality?

Additionally, we found that we received three different error traces.  Examples:

1)

    Microsoft.IdentityModel.Clients.ActiveDirectory.AdalServiceException: Too Many Requests ---> System.Net.Http.HttpRequestException:  Response status code does not indicate success: 429 (429). ---> System.Exception: Too Many Requests
       --- End of inner exception stack trace ---
       at Microsoft.IdentityModel.Clients.ActiveDirectory.HttpClientWrapper.<GetResponseAsync>d__29.MoveNext()
       --- End of inner exception stack trace ---
       at Microsoft.IdentityModel.Clients.ActiveDirectory.AdalHttpClient.<GetResponseAsync>d__21`1.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
       at Microsoft.IdentityModel.Clients.ActiveDirectory.AdalHttpClient.<GetResponseAsync>d__20`1.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---

2)

    Microsoft.IdentityModel.Clients.ActiveDirectory.AdalServiceException: AADSTS90055: The server has terminated the request due to excessive request rate. Please wait a few seconds and try again.
    Trace ID: b0f35b59-7342-406f-9379-00a2e3de2400
    Correlation ID: 7c8e8729-ab77-4638-afa9-394d550e774c
[...]

PR #771, opened 2017-08-30
Trwalke/retry after header #721

trwalke:
Added HttpResponseHeaders to the AdalServiceException that contains all headers returned in the web response. #721

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AdalServiceException.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AdalServiceException.cs
@@ -29,6 +29,7 @@ using System.Globalization;
 using System;
 using System.Net;
 using System.Threading.Tasks;
+using System.Collections.Generic;
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 {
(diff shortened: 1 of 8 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/AdalServiceException.cs
```diff
@@ -116,6 +119,11 @@ internal AdalServiceException(string errorCode, string message, string[] service
         public string[] ServiceErrorCodes { get; set; }
 
         /// <summary>
+        /// Contains headers from the response that indicated an error
+        /// </summary>
+        public HttpResponseHeaders Headers { get; set; }
```
kpanwar: there should be no public setters. make set internal.

tests/Test.ADAL.NET.Unit/Mocks/MockHelpers.cs
```diff
@@ -92,6 +92,18 @@ public static HttpResponseMessage CreateFailureResponseMessage(string message)
             return responseMessage;
         }
 
+        public static HttpResponseMessage CreateCustomHeaderFailureResponseMessage(IEnumerable<KeyValuePair<string, string>> headers)
+        {
+            HttpResponseMessage responseMessage = CreateHttpErrorResponse();
+
+            foreach (KeyValuePair<string, string> header in headers)
+                responseMessage.Headers.Add(header.Key, header.Value);
```
kpanwar: put curly braces around foreach loop code

Issue #534, opened 2016-09-08
iOS: ViewController passed with PlatformParameters is leaked


MKuckert:
The ViewController passed into `AcquireTokenAsync` is leaked by the library and never collected by the garbage collector.
The `AcquireTokenInteractiveHandler` class assigns the given parameters instance to `PlatformPlugin.BrokerHelper.PlatformParameters`, which in turn is never null-ed. This leads to a hard reference to the given ViewController.

kpanwar:
I will make the view controller assignment a weak reference

kpanwar:
@MKuckert - I have been super busy. Feel free to send me a PR and I will review, merge and release it.

kpanwar:
We need to update processBrokerResponse() to set platformParameters = null

PR #775, opened 2017-08-31
dereference UIViewController for iOS

kpanwar:
dereference UIViewController instances in Xamarin.iOS upon completion of the request. not doing so causes the OS to retain the object and causes memory leak, per #534


SomkaPe:
I think there will be no leak anyway - AcquireTokenInteractiveHandler and BrokerHelper instances are local  to AcquireTokenAsync  method, so after exiting from AcquireTokenAsync  both instances  will be collected by GC and no references to PlatformParameters will be available

SomkaPe:
:shipit:

Issue #777, opened 2017-09-01
Add exception to AdalClaimChallengeException


jennyf19:
Pass HttpRequestWrapperException innerException to AdalClaimChallengeException to allow use of the inner exception to get the missing claims.

jennyf19:
https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/778

PR #778, opened 2017-09-01
Jennyf/claims inner exception 

jennyf19:
#777

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AdalClaimChallengeException.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AdalClaimChallengeException.cs
@@ -35,12 +35,12 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     public class AdalClaimChallengeException : AdalServiceException
     {
         /// <summary>
-        /// Claims returned from eSTS during token acquisition.
+        /// Represents claims returned from AdalClaimChallengeException
         /// </summary>
         public string Claims { get; internal set; }
 
-        /// <summary
-        ///  Initializes a new instance of the exception class for handling claims.
+        /// <summary>
+        /// Initializes a new instance of the exception class for handling claims.
         /// </summary>
         public AdalClaimChallengeException(string errorCode, string message, Exception innerException, string claims)
             : base(errorCode, GetErrorMessage(errorCode), null, innerException)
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Http/AdalHttpClient.cs
```diff
@@ -128,8 +128,7 @@ public AdalHttpClient(string uri, CallState callState)
                     if (ex.WebResponse.StatusCode == HttpStatusCode.BadRequest &&
                         tokenResponse.Error == AdalError.InteractionRequired)
                     {
-                        throw new AdalClaimChallengeException(tokenResponse.Error, tokenResponse.ErrorDescription,
-                            tokenResponse.Claims);
+                        throw new AdalClaimChallengeException(tokenResponse.Error, tokenResponse.ErrorDescription, ex.InnerException, tokenResponse.Claims);
```
kpanwar: passed in exception seems to be incorrect. check what exception object is passed in when it is not claims challenge exception

kpanwar:
we should update the unit test for this

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Http/AdalHttpClient.cs
jennyf19: Done.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AdalClaimChallengeException.cs
```diff
@@ -33,17 +35,17 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     public class AdalClaimChallengeException : AdalServiceException
     {
         /// <summary>
-        ///  Initializes a new instance of the exception class.
+        /// Claims returned from eSTS during token acquisition.
```
kpanwar: why has the documentation changed?

jennyf19: fixed.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AdalClaimChallengeException.cs
```diff
@@ -33,17 +35,17 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     public class AdalClaimChallengeException : AdalServiceException
     {
         /// <summary>
-        ///  Initializes a new instance of the exception class.
+        /// Represents claims returned from AdalClaimChallengeException
```
kpanwar: this is still different

jennyf19: done.

kpanwar:
:shipit:

Issue #752, opened 2017-08-14
Using a certificate with non-exportable private key with AcquireTokenAsync(string resource, IClientAssertionCertificate clientCertificate) on dotNet 4.7


EranBE:
Related to this issue - [https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/708](url)

If you target .Net 4.7 and try to use a certificate client assertion with a certificate that has a non-exportable private key, a `CryptographicException` will be thrown since the current code calls `certificate.PrivateKey.ToXmlString(true)` which cannot export the private key.
Prior to .Net 4.7 the same certificate worked since it returned `RSACryptoServiceProvider` as its private key instance.

dmcweeney:
Hi,

We are having same problem here. Upgraded projects to target .Net Framework 47. Also did a pile of package updates - mainly upgrading from asp.net core 1.2 to 2.0.

On my machine code runs fine but on other dev machines we are getting the same error:

System.Security.Cryptography.CryptographicException: Key not valid for use in specified state.
at System.Security.Cryptography.CryptographicException.ThrowCryptographicException(Int32 hr)
at System.Security.Cryptography.Utils._ExportKey(SafeKeyHandle hKey, Int32 blobType, Object cspObject)
at System.Security.Cryptography.RSACryptoServiceProvider.ExportParameters(Boolean includePrivateParameters)
at System.Security.Cryptography.RSA.ToXmlString(Boolean includePrivateParameters)
at Microsoft.IdentityModel.Clients.ActiveDirectory.CryptographyHelper.GetCryptoProviderForSha256(X509Certificate2 certificate)
at Microsoft.IdentityModel.Clients.ActiveDirectory.CryptographyHelper.SignWithCertificate(String message, X509Certificate2 certificate)
at Microsoft.IdentityModel.Clients.ActiveDirectory.ClientAssertionCertificate.Sign(String message)
at Microsoft.IdentityModel.Clients.ActiveDirectory.JsonWebToken.Sign(IClientAssertionCertificate credential)
at Microsoft.IdentityModel.Clients.ActiveDirectory.ClientKey.AddToParameters(IDictionary`2 parameters)
at Microsoft.IdentityModel.Clients.ActiveDirectory.AcquireTokenHandlerBase.<SendTokenRequestAsync>d__64.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #779, opened 2017-09-02
Fix to use RSACng when RSA CSP is not resolved

kpanwar:
This PR includes fix to use RSACng instead of RSACSP in case of  .net 4.7. Existing code was attempting to export private key and create RSA CSP from it. This would fail if the key is marked as non-exportable, per #752

The CI builds will fail as they expect multi-target project configuration. Internal build system will still be able to create a successful build

### DIFF
--- a/src/ADAL.PCL.Desktop/CryptographyHelper.cs
+++ b/src/ADAL.PCL.Desktop/CryptographyHelper.cs
@@ -31,6 +31,13 @@ using System.IdentityModel.Tokens;
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
 using System.Text;
+using Microsoft.Win32.SafeHandles;
+using System.Runtime.ConstrainedExecution;
+using System.Runtime.InteropServices;
+using System.Security;
+using System.Security.Permissions;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Native;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 {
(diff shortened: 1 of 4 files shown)


src/ADAL.PCL.Desktop/CryptographyHelper.cs
```diff
@@ -72,20 +79,20 @@ public byte[] SignWithCertificate(string message, X509Certificate2 certificate)
...
-                    // Instead we try and create an RSACryptoServiceProvider based on the private key from the
-                    // certificate.
-                    newRsa = GetCryptoProviderForSha256(certificate);
-                }
-
-                using (var sha = new SHA256Cng())
-                {
-                    return newRsa.SignData(messageBytes, sha);
+                    CngKey key = GetCngPrivateKey(certificate);
+                    using (RSACng rsaCng = new RSACng(key))
```
jmprieur: The doc https://msdn.microsoft.com/en-us/library/dn957360(v=vs.110).aspx says that if the key is null, the constructor of RSACng will throw an exception. I'm not clear what happens when the user is not allowed to access the certificate, @kpanwar  (the full trust assertion). Will this provoke an exception which will be bubbled-up from this method?

src/ADAL.PCL.Desktop/CryptographyHelper.cs
```diff
@@ -145,5 +139,90 @@ private static RSACryptoServiceProvider GetCryptoProviderForSha256(RSACryptoServ
 
             return rsaProvider;
         }
+
+
+        /// <summary>
+        ///     <para>
```
jmprieur: I don't think that the first 2 para apply to this method? Is this the standard XML comment syntax?

Issue #654, opened 2017-04-25
Invalid methods documentation 


anaili:
**This issue concerns only versions 3+.**
As I understand, in versions 3+ the AquireToken methods doesn't return a refresh token anymore but after a successful login it will be managed internally in the TokenCache instance and the user can re-implement some methods to persist it or serialize and store the cache state. The issue is the documentation is still saying (which is very confusing) :

> Returns System.Threading.Tasks.Task\<AuthenticationResult\> :
>    It contains Access Token, Refresh Token and the Access Token's expiration time.

Issue #624, opened 2017-03-07
UserPasswordCredential triple slash comment is incorrect


skwan:
This is for username and password not client ID and secret (ROPG not client creds).  I bet there are other instances of this error for UserCredential.

https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/src/ADAL.PCL.Desktop/UserPasswordCredential.cs#L50

Issue #527, opened 2016-08-27
ADALTrace removed; diagnostics instructions need updating?


ScottArbeit:
The [README.md](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/README.md) file refers to the `ADALTrace` static class as a way to configure diagnostic logging for ADAL.NET. That class seems to have been removed now.

Is there a new method for enabling diagnostic logging that we can update the documentation for?

Thanks!

JosephEarl:
We had the same problem and found you can use `LoggerCallbackHandler.Callback` and set it to a custom implementation of `IAdalLogCallback` - https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/master/src/ADAL.PCL/IAdalLogCallback.cs.

ssaroiu:
JosephEarl's solution worked for me for ADAL version 3.13.5. Thanks!

hamish-tushar:
Can you tell me how to enable logging and diagnostics for ADAL?

pschaeflein:
Sample implementation at https://www.schaeflein.net/adal-v3-diagnostic-logging/

hamish-tushar:
Thank you so much

PR #788, opened 2017-09-07
doc fixes

SomkaPe:
UserPasswordCredential triple slash comment is incorrect  https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/624

ADALTrace removed; diagnostics instructions need updating https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/527

Invalid methods documentation https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/654

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ This is obviously the first diagnostic.  We try to provide helpful error message
 
 ### Logs
 
-In order to configure logging, static property Callback of the LoggerCallbackHandler class should be set to the instance of class implementing IAdalLogCallback interface
+In order to configure logging, static property Callback of the LoggerCallbackHandler class should be set to the instance of a class implementing IAdalLogCallback interface
 
 ### Brokered Authentication for iOS
 
--- a/README.md
+++ b/README.md
@@ -66,7 +66,22 @@ This is obviously the first diagnostic.  We try to provide helpful error message
 
 ### Logs
 
-In order to configure logging, static property Callback of the LoggerCallbackHandler class should be set to the instance of a class implementing IAdalLogCallback interface
+In order to configure logging, implementation of IAdalLogCallback interface should be provided
+
+```C#
+class LoggerCallbackImpl : IAdalLogCallback
+{
+    public void Log(LogLevel level, string message)
+    {
+        // process log message, for example write it to your favorite logging framework
+    }
+}
+```
+static property Callback of the LoggerCallbackHandler class should be set to the instance of a class implementing IAdalLogCallback interface
+
+```C#
+LoggerCallbackHandler.Callback = new LoggerCallbackImpl();
+```
 
 ### Brokered Authentication for iOS
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Active Directory Authentication Library (ADAL) for .NET, Windows Store, Xamarin iOS and Xamarin Android. 
+# Active Directory Authentication Library (ADAL) for .NET, Windows Store, Xamarin iOS and Xamarin Android.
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -66,101 +66,7 @@ This is obviously the first diagnostic.  We try to provide helpful error message
...
-}
-
-```
-
-There is also a default event listener which writes logs to a local file named **"AdalTraces.log"**. You can control the level of tracing to that event listener using the property ```AdalTrace.Level```. By default, trace level for this event listener is set to "None" and to enable tracing to this particular listener, you need to set the above property. This is an example:
-
-```
-AdalTrace.Level = AdalTraceLevel.Informational;
-```
+In order to configure logging, static property Callback of the LoggerCallbackHandler class should be set to the instance of class implementing IAdalLogCallback interface
```
jmprieur: the instance of **a** class?
Also do we have such a class in our devApp? if yes, can we point to the line were LoggerCallbackHandler is set?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -145,7 +145,7 @@ public Guid CorrelationId
         /// </summary>
         /// <param name="resource">Identifier of the target resource that is the recipient of the requested token.</param>
         /// <param name="clientId">Identifier of the client requesting the token.</param>
-        /// <returns>It contains Access Token, Refresh Token and the Access Token's expiration time.</returns>
+        /// <returns>It contains Access Token, its expiration time, user information.</returns>
```
jmprieur: That's a device code? doesn't it contain info about the device?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/UserAssertion.cs
```diff
@@ -44,7 +44,7 @@ public UserAssertion(string assertion) : this(assertion, OAuthGrantType.JwtBeare
         }
 
         /// <summary>
-        /// Constructor to create credential with client id, assertion and assertionType
+        /// Constructor to create credential with assertion and assertionType
```
jmprieur: We probably want to link to a doc where this flow is explained? As a user I've always found these overload confusing. We need to help the developers who are not Identity geeks, there :)

README.md
SomkaPe: yes we have it, but since it is straightforward how to set it and dev applications might be changed/ refactored I would not use line references.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/UserAssertion.cs
SomkaPe: we do not have flows explanations in source code, so we might add link to https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-protocols-oauth-on-behalf-of , but since it is web resource It might become unavailable later

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
SomkaPe: good catch, will fix it

src/Microsoft.IdentityModel.Clients.ActiveDirectory/UserAssertion.cs
SomkaPe: provided sample

README.md
SomkaPe: provided sample

jmprieur: Great, thanks!

Issue #794, opened 2017-09-09
Update WinFormsAutomationApp for claims


jennyf19:
For TestAutomation, update WinFormsAutomationApp to handle claims

PR #795, opened 2017-09-09
Update to WinFormsAutomationApp to handle claims

jennyf19:
#794

### DIFF
--- a/automation/WinFormsAutomationApp/AuthenticationHelper.cs
+++ b/automation/WinFormsAutomationApp/AuthenticationHelper.cs
@@ -40,8 +40,8 @@ namespace WinFormsAutomationApp
                     {
                         userIdentifierType = UserIdentifierType.UniqueId;
                     }
-                     else if (string.Equals(input["user_identifier_type"], "optional_displayable",
-                        StringComparison.InvariantCultureIgnoreCase))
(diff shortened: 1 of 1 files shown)


automation/WinFormsAutomationApp/AuthenticationHelper.cs
```diff
@@ -50,17 +50,30 @@ internal static class AuthenticationHelper
                     {
                         userIdentifierType = UserIdentifierType.RequiredDisplayableId;
                     }
+
                     string prompt = input.ContainsKey("prompt_behavior") ? input["prompt_behavior"] : null;
-                    result = await ctx.AcquireTokenAsync(input["resource"], input["client_id"], new Uri(input["redirect_uri"]),
-                        GetPlatformParametersInstance(prompt), 
+
+                    // extraQueryParameters can be null, but key must be present to use correct Acquire Token call for claims overload
+                    if (input.ContainsKey("claims") && (input.ContainsKey("extra_query_parameters")))
```
kpanwar: you can call claims API without extra QP as well. so, there should not be a check to ensure extra_query_parameters are present for calling claims overload

jennyf19: Done.

Issue #804, opened 2017-09-14
Ensure device profile flow integration tests exist in ADAL source code


trwalke:
Ensure device profile flow integration tests exist in ADAL source code.
- Test case file DeviceProfileFlowTest.cs moved to integration test project.
- One positive test validating the full flow (acquireDeviceCode followed by acquireTokenUsingDeviceCode).
- One negative test for acquiring device code exists.
- one negative test for acquiring token using device code where the polling times out exists.
VS Online Story 284945

PR #805, opened 2017-09-14
Test case file DeviceProfileFlowTest.cs moved to integration test pro…

trwalke:
Test case file DeviceProfileFlowTest.cs moved to integration test project. #804


kpanwar:
:shipit:

PR #816, opened 2017-09-19
Authority Migration: TokenCache read/write changes

rayluo:
* Use the preferred_cache, given host, and aliases from the Authority validation cache to look up tokens in the token cache
* Use the preferred_cache when writing tokens into cache

~PS: This PR is now ready for review, but please do NOT merge this PR yet. Because it is based on a [previous PR](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/796/files) which has not yet been merged as of today. I'll need to change the base branch later.~

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
@@ -136,15 +136,13 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
                     this.NotifyBeforeAccessCache();
                     notifiedBeforeAccessCache = true;
 
-                    var host = GetHostCaseSensitive(this.Authenticator.Authority);
-                    var metadata = await InstanceDiscovery.GetMetadataEntry(host, this.Authenticator.ValidateAuthority, this.CallState).ConfigureAwait(false);
-                    var aliasedAuthorities = new List<string>(new string[] {metadata.PreferredCache, host});
-                    aliasedAuthorities.AddRange(metadata.Aliases ?? Enumerable.Empty<string>());
+                    var aliasedAuthorities = await GetOrderedAliases(
+                        GetHostCaseSensitive(this.Authenticator.Authority), this.Authenticator.ValidateAuthority, this.CallState).ConfigureAwait(false);
                     foreach (var aliasedAuthority in aliasedAuthorities)
                     {
                         CacheQueryData.Authority = ReplaceHost(this.Authenticator.Authority, aliasedAuthority);
                         ResultEx = this.tokenCache.LoadFromCache(CacheQueryData, this.CallState);
-                        if (ResultEx?.Result != null) break;
+                        if (ResultEx?.Result != null) {break;} //NOSONAR
                     }
 
                     extendedLifetimeResultEx = ResultEx;
@@ -206,6 +204,14 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
                     this.NotifyAfterAccessCache();
                 }
             }
+        }
+
+        public static async Task<List<string>> GetOrderedAliases(string host, bool validateAuthority, CallState callState)
+        {
+            var metadata = await InstanceDiscovery.GetMetadataEntry(host, validateAuthority, callState).ConfigureAwait(false);
+            var aliasedAuthorities = new List<string>(new string[] { metadata.PreferredCache, host });
+            aliasedAuthorities.AddRange(metadata.Aliases ?? Enumerable.Empty<string>());
+            return aliasedAuthorities;
         }
 
         private async Task<bool> StoreResultExToCache(bool notifiedBeforeAccessCache)
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
@@ -28,6 +28,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Threading.Tasks;
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory
@@ -126,7 +127,6 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
                 if (this.LoadFromCache)
                 {
                     CallState.Logger.Verbose(CallState, "Loading from cache.");
-                    CacheQueryData.Authority = Authenticator.Authority;
                     CacheQueryData.Resource = this.Resource;
                     CacheQueryData.ClientId = this.ClientKey.ClientId;
                     CacheQueryData.SubjectType = this.TokenSubjectType;
@@ -135,7 +135,18 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 
                     this.NotifyBeforeAccessCache();
                     notifiedBeforeAccessCache = true;
-                    ResultEx = this.tokenCache.LoadFromCache(CacheQueryData, this.CallState);
+
+                    var host = GetHostCaseSensitive(this.Authenticator.Authority);
+                    var metadata = await InstanceDiscovery.GetMetadataEntry(host, this.Authenticator.ValidateAuthority, this.CallState).ConfigureAwait(false);
+                    var aliasedAuthorities = new List<string>(new string[] {metadata.PreferredCache, host});
+                    aliasedAuthorities.AddRange(metadata.Aliases ?? Enumerable.Empty<string>());
+                    foreach (var aliasedAuthority in aliasedAuthorities)
+                    {
+                        CacheQueryData.Authority = ReplaceHost(this.Authenticator.Authority, aliasedAuthority);
+                        ResultEx = this.tokenCache.LoadFromCache(CacheQueryData, this.CallState);
+                        if (ResultEx?.Result != null) break;
+                    }
+
                     extendedLifetimeResultEx = ResultEx;
 
                     if (ResultEx?.Result != null &&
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -205,10 +223,25 @@ private void StoreResultExToCache(ref bool notifiedBeforeAccessCache)
...
+                    ResultEx, ReplaceHost(this.Authenticator.Authority, metadata.PreferredCache), this.Resource,
                     this.ClientKey.ClientId, this.TokenSubjectType, this.CallState);
             }
+            return notifiedBeforeAccessCache;
+        }
+
+        private string GetHost(string uri)
+        {
+            // Note: host is supposed to be case insensitive, but we would like to preserve its case to match a previously cached token
+            return uri.Split('/')[2];
```
partnerinflight: hm that seems a bit dangerous... don't we want to do some validation that this uri is actually a uri, and is valid?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -205,10 +223,25 @@ private void StoreResultExToCache(ref bool notifiedBeforeAccessCache)
...
+            // return new Uri(uri).Host; // This canonical implementation normalizes the host into lower case
+        }
+
+        private string ReplaceHost(string oldUri, string newHost)
+        {
+            return $"https://{newHost}{new Uri(oldUri).AbsolutePath}";
```
partnerinflight: again, not a whole lot of error checking here.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
rayluo: Solution 1: Well, if you prefer, we can add a `new Uri(uri)` before this line, so that all the invalid input that fails [the default checks in Uri() constructor](https://msdn.microsoft.com/en-us/library/z6c2z492(v=vs.110).aspx#Exceptions) would end up with an (different) exception.

Solution 2: the reason that I did not go with the elagent `new Uri(uri).Host` implementation is that, this [pre-existing line in ANOTHER test case](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/Release-3.17.0/tests/Test.ADAL.Common/TokenCacheTests.cs#L177) somehow uses mixed case as input. So after the implicit normalization of `new Uri(uri).Host`, it gives us a lower case host name which won't match the previous record in the cache. Note that, in practice, host would and probably should always be in lower case, so this problem should not exist. If you don't mind, I can go ahead and adjust that aforementioned test case, and then we can use `new Uri(uri).Host` here.

What do you think?

kpanwar:
Why aren't your PRs going to dev? We would need to do multiple code reviews of the same thing when this is sent to dev branch.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -135,7 +135,16 @@ protected AcquireTokenHandlerBase(RequestData requestData)
...
                     notifiedBeforeAccessCache = true;
-                    ResultEx = this.tokenCache.LoadFromCache(CacheQueryData, this.CallState);
+
+                    var aliasedAuthorities = await GetOrderedAliases(
+                        GetHost(this.Authenticator.Authority), this.Authenticator.ValidateAuthority, this.CallState).ConfigureAwait(false);
+                    foreach (var aliasedAuthority in aliasedAuthorities)
+                    {
+                        CacheQueryData.Authority = ReplaceHost(this.Authenticator.Authority, aliasedAuthority);
+                        ResultEx = this.tokenCache.LoadFromCache(CacheQueryData, this.CallState);
+                        if (ResultEx?.Result != null) break;
```
aadidgit: CRITICAL Refactor this code to not nest more than 3 control flow statements. [rule](http://iddp-sonarqube:9000/coding_rules#rule_key=csharpsquid%3AS134)

aadidgit:
SonarQube analysis reported 1 issue
* ![CRITICAL][CRITICAL] 1 critical

Watch the comments in this conversation to review them.

[CRITICAL]: https://sonarsource.github.io/sonar-github/severity-critical.png 'Severity: CRITICAL'

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -205,10 +223,28 @@ private void StoreResultExToCache(ref bool notifiedBeforeAccessCache)
...
+                this.tokenCache.StoreToCache(
+                    ResultEx, ReplaceHost(this.Authenticator.Authority, metadata.PreferredCache), this.Resource,
                     this.ClientKey.ClientId, this.TokenSubjectType, this.CallState);
             }
+            return notifiedBeforeAccessCache;
+        }
+
+        private string GetHost(string uri)
+        {
+            new Uri(uri); // This line serves as a validation for uri. Relevant exceptions will be throwed.
```
aadidgit: MAJOR Either remove this useless object instantiation of class 'Uri' or use it. [rule](http://iddp-sonarqube:9000/coding_rules#rule_key=csharpsquid%3AS1848)

aadidgit:
SonarQube analysis reported 1 issue
* ![MAJOR][MAJOR] 1 major

Watch the comments in this conversation to review them.

[MAJOR]: https://sonarsource.github.io/sonar-github/severity-major.png 'Severity: MAJOR'

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -135,7 +135,16 @@ protected AcquireTokenHandlerBase(RequestData requestData)
...
                     notifiedBeforeAccessCache = true;
-                    ResultEx = this.tokenCache.LoadFromCache(CacheQueryData, this.CallState);
+
+                    var aliasedAuthorities = await GetOrderedAliases(
+                        GetHost(this.Authenticator.Authority), this.Authenticator.ValidateAuthority, this.CallState).ConfigureAwait(false);
+                    foreach (var aliasedAuthority in aliasedAuthorities)
+                    {
+                        CacheQueryData.Authority = ReplaceHost(this.Authenticator.Authority, aliasedAuthority);
+                        ResultEx = this.tokenCache.LoadFromCache(CacheQueryData, this.CallState);
+                        if (ResultEx?.Result != null) {break;} //NOSONAR
```
HSAR: Stylistic comment: expand one-line if-statements for readability.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -205,10 +223,29 @@ private void StoreResultExToCache(ref bool notifiedBeforeAccessCache)
...
+            new Uri(uri); //NOSONAR
+
+            // Note: host is supposed to be case insensitive, and would be normalized to lowercase by: new Uri(uri).Host
+            // but we would like to preserve its case to match a previously cached token
+            return uri.Split('/')[2];
+        }
+
+        private string ReplaceHost(string oldUri, string newHost)
+        {
+            if (string.IsNullOrEmpty(oldUri) || string.IsNullOrEmpty(newHost)) {throw new ArgumentNullException();}
```
HSAR: Stylistic comment: expand one-line if-statements for readability.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -196,7 +209,15 @@ protected AcquireTokenHandlerBase(RequestData requestData)
...
-        private void StoreResultExToCache(ref bool notifiedBeforeAccessCache)
+        public static async Task<List<string>> GetOrderedAliases(string host, bool validateAuthority, CallState callState)
+        {
+            var metadata = await InstanceDiscovery.GetMetadataEntry(host, validateAuthority, callState).ConfigureAwait(false);
+            var aliasedAuthorities = new List<string>(new string[] { metadata.PreferredCache, host });
+            aliasedAuthorities.AddRange(metadata.Aliases ?? Enumerable.Empty<string>());
+            return aliasedAuthorities;
+        }
+
+        private async Task<bool> StoreResultExToCache(bool notifiedBeforeAccessCache)
```
kpanwar: why do we need to return bool here? we can simply update the global variable instead

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -135,7 +135,19 @@ protected AcquireTokenHandlerBase(RequestData requestData)
 
                     this.NotifyBeforeAccessCache();
                     notifiedBeforeAccessCache = true;
-                    ResultEx = this.tokenCache.LoadFromCache(CacheQueryData, this.CallState);
+
+                    var aliasedAuthorities = await GetOrderedAliases(
+                        GetHost(this.Authenticator.Authority), this.Authenticator.ValidateAuthority, this.CallState).ConfigureAwait(false);
+                    foreach (var aliasedAuthority in aliasedAuthorities)
```
kpanwar: this could be moved into TokenCache class.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
rayluo: Because:

* That `notifiedBeforeAccessCache` thing is not global. It is a local variable in the calling function
* This method becomes an async method now, and an async method can not use the `ref` way to update the input variable
* Last but not the least, when a function is going to provide something back to the caller, it seems using return value would be the most intuitive way.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
rayluo: It could. But then it would require the TokenCache to be aware of not just an authority string, but also the entire InstanceCache thing in order to swap different authoiry aliases. While such conceptual difference is not necessarily bad, the substantial change is that `LoadFromCache(...)` would then have to become an async method, and potentially wait arbitrarily long for slow network I/O. It does not sound right for a cache class which is supposed to be fast, non-block, and decisive.

rayluo: As an alternative, in the latest commit I try a different approach. I group this inline loop into a standalone helper `LoadResultExFromCache()` which still sits inside this `AcquireTokenHandlerBase` class, to team up with its pre-existing cousin `StoreResultExToCache()`.

kpanwar: this is not an alternative. We clearly discussed the separation of concerns and various aspects. I though we agreed to move this in token cache. Is there a disconnect?

rayluo:
Closing this one, because we switched to a different approach in PR #851

PR #818, opened 2017-09-19
Ensure prompt behavior for each value of prompt behavior

jennyf19:
#817 
- test calling promptBehavior.Auto with AT in cache
- Create class for PromptBehavior tests and move existing prompt tests from AdalDotNetTests


tests/Test.ADAL.NET.Unit/PromptBehaviorTests.cs
```diff
@@ -0,0 +1,320 @@
+﻿//----------------------------------------------------------------------
```
kpanwar: these should move to integration tests project

tests/Test.ADAL.NET.Unit/PromptBehaviorTests.cs
```diff
@@ -0,0 +1,320 @@
...
+    public class PromptBehaviorTests
+    {
+        [TestInitialize]
+        public void Initialize()
+        {
+            HttpMessageHandlerFactory.ClearMockHandlers();
+        }
+        
+        [TestMethod]
+        [Description("Test for Force Prompt with PromptBehavior.Auto")]
```
kpanwar: description is incorrect. Prompt.Auto means that you are not forcing prompt

tests/Test.ADAL.NET.Unit/PromptBehaviorTests.cs
```diff
@@ -0,0 +1,320 @@
...
+            context.TokenCache.Clear();
+
+            HttpMessageHandlerFactory.AddMockHandler(new MockHttpMessageHandler()
+            {
+                Method = HttpMethod.Post,
+                ResponseMessage = new HttpResponseMessage(HttpStatusCode.Unauthorized)
+                {
+                    Content = new StringContent("{\"token_type\":\"Bearer\",\"expires_in\":\"3599\",\"access_token\":\"some-access-token\"}")
+                },
+                PostData = new Dictionary<string, string>()
```
kpanwar: client secret should not be sent for public clients. this is not a client_credentials grant. Code should execute refresh_token grant

tests/Test.ADAL.NET.Unit/PromptBehaviorTests.cs
jennyf19: This is the description from the pre-existing test case. I will change it.

tests/Test.ADAL.NET.Unit/PromptBehaviorTests.cs
jennyf19: Done.

tests/Test.ADAL.NET.Unit/PromptBehaviorTests.cs
jennyf19: Done.

tests/Test.ADAL.NET.Unit/PromptBehaviorTests.cs
kpanwar: I still see this in tests/Test.ADAL.NET.Unit/PromptBehaviorTests.cs

PR #819, opened 2017-09-20
Created a negative test for acquiring device code. #804

trwalke:


tests/Test.ADAL.NET.Integration/DeviceProfileTests.cs
```diff
@@ -143,5 +143,30 @@ public void Initialize()
...
+            try
+            {
+                DeviceCodeResult dcr = await ctx.AcquireDeviceCodeAsync("some-resource", "some-client");
+            }
+            catch (Exception ex)
+            {
+                Assert.AreEqual(ex.Message, "AADSTS90014: The request body must contain the following parameter: 'client_id'.\r\nTrace ID: 290d2ab9-40f2-4716-92e2-4a72fc480000\r\nCorrelation ID: 2eee49ee-620e-42c2-9a3c-dcf81955b20f\r\nTimestamp: 2017-09-20 23:05:56Z");
+                return;
+            }
+            Assert.Fail();
```
jmprieur: @trwalke : I'm not sure I understand what is expected here? the test method will fail?
Did you intend to use the [ExpectedExceptionAttribute](https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.testtools.unittesting.expectedexceptionattribute.aspx)

jmprieur: @trwalke : I'm not sure I understand what is expected from Assert.Fail() here? the test method will fail?
Did you intend to use the [ExpectedExceptionAttribute](https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.testtools.unittesting.expectedexceptionattribute.aspx)

tests/Test.ADAL.NET.Unit/Mocks/MockHelpers.cs
```diff
@@ -113,6 +113,12 @@ public static HttpResponseMessage CreateHttpErrorResponse()
                     "{\"ErrorSubCode\":\"70323\",\"error\":\"invalid_grant\",\"error_description\":\"AADSTS70002: Error validating credentials.AADSTS70008: The provided access grant is expired or revoked.Trace ID: f7ec686c-9196-4220-a754-cd9197de44e9Correlation ID: 04bb0cae-580b-49ac-9a10-b6c3316b1eaaTimestamp: 2015-09-16 07:24:55Z\",\"error_codes\":[70002,70008],\"timestamp\":\"2015-09-16 07:24:55Z\",\"trace_id\":\"f7ec686c-9196-4220-a754-cd9197de44e9\",\"correlation_id\":\"04bb0cae-580b-49ac-9a10-b6c3316b1eaa\"}");
```
kpanwar: We should not create error messages that have messages specific to certain failures. Consider creating a generic error response. there should be helper functions to do this already

tests/Test.ADAL.NET.Integration/DeviceProfileTests.cs
trwalke: Yes, this was just a way to verify that an exception was thrown because this was supposed to be a test that fails. I was not aware of the ExpectedExceptionAttribute and I will incorporate that into my test. thanks

PR #821, opened 2017-09-21
fixes for Android and IOS platforms

SomkaPe:
- adding scrollable log section to XFormsApp
- adding padding to IOS XFormsApp to keep all ui elements visible
- fixing CallerViewController Null pointer exception for IOS platform,
It is related to recent fix for "memory leaks on IOS"
Since PlatformParameter instance might an probably should be used across few AcquireToken calls, and because AcquireToken methods do not  own it (PlatformParameter created and owned by outer to AcquireToken method code )  It should not be changed inside  AcquireToken calls 
- fix for null pointer exception on Android while calling CanSwitchToBroker
- fix for applying SupportToAddUserThroughBroker logic to Company Portal broker

### DIFF
--- a/devApps/XFormsApp/SecondPage.cs
+++ b/devApps/XFormsApp/SecondPage.cs
@@ -29,6 +29,7 @@ using Microsoft.IdentityModel.Clients.ActiveDirectory;
 using System;
 using System.Text;
 using Xamarin.Forms;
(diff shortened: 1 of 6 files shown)


kpanwar:
@SomkaPe - don't forget to send PR into DEV from the same branch

PR #824, opened 2017-09-21
Authority Migration: Use preferred network to talk to endpoints (by changing endpoints early)

rayluo:
~CREATED at Sep 21: This is not yet a formal PR. It is for discussion purpose.~

~UPDATE at Sep 27: There is an alternative implementation in another PR #832. The current PR will probably be closed soon.~

UPDATE at Oct 9: This PR receives new updates. We are now reconsidering it as a qualified candidate.

### DIFF
--- a/tests/Test.ADAL.NET.Unit/InstanceDiscoveryTests.cs
+++ b/tests/Test.ADAL.NET.Unit/InstanceDiscoveryTests.cs
@@ -35,6 +35,9 @@ using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal;
 using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Http;
 using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Instance;
 using Test.ADAL.NET.Common.Mocks;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Flows;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.ClientCreds;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Cache;
 
 namespace Test.ADAL.NET.Unit
 {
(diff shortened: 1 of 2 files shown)


kpanwar:
@rayluo - this is still a relevant PR? if not then can you please close this?

rayluo:
@kpanwar Yes we will eventually close this PR, but I assume it still has some reference value until its successor #832 officially get merged. Specifically, we can see how those design/implementation decisions are not an issue in this PR's approach.

tests/Test.ADAL.NET.Unit/InstanceDiscoveryTests.cs
```diff
@@ -179,12 +182,130 @@ public void TestInstanceDiscovery_WhenMultipleSimultaneousCallsWithTheSameAuthor
...
+        {
+            HttpMessageHandlerFactory.AddMockHandler(new MockHttpMessageHandler
+            {
+                Method = HttpMethod.Get,
+                Url = $"https://{host}/common/discovery/instance",
+                ResponseMessage = new HttpResponseMessage(HttpStatusCode.OK)
+                {
+                    Content = new StringContent(
+                        @"{
+                        ""tenant_discovery_endpoint"" : ""https://login.microsoftonline.com/v1/.well-known/openid-configuration"",
```
kpanwar: it should be {host} instead of login.microsoftonline.com

rayluo: Thanks for the hint that triggers an investigation. As it turns out that, the server response tends to present the querying `host` as-is into the `tenant_discovery_endpoint`, such as [this one](https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https://login.windows.net/adfs/oauth2/authorize) and [that one](https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https://login.microsoftonline.us/adfs/oauth2/authorize), but [not always so](https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https://sts.windows.net/adfs/oauth2/authorize). So, using `{host}` instead of `login.microsoftonline.com` is not necessarily always a pricise choice.

But I guess none of above really matters. From the client-side perspective, ADAL does not have an opinion what endpoint URL the server returns. Client would just follow it. And in this particular case, ADAL does not actually send out a tenant discovery request at all. So, it doesn't matter. We could go with a `login.placeholder.com` and our test cases would still pass.

Lastly, this test value is currently a verbatim string containing a json blob. It would be very difficult to get string interpolation right.

All in all, I will probably change the current hardcoded `login.microsoftonline.com` to `placeholder` and that's it.

kpanwar: nah, keep it the same way then.

PR #825, opened 2017-09-22
Created one negative test for acquiring token using device code where…

trwalke:
… the polling times out.

Resolved issue where ADAL did not return the STS response after a device code has expired.

#804


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/DeviceCode/AcquireTokenByDeviceCodeHandler.cs
```diff
@@ -47,7 +47,7 @@ public AcquireTokenByDeviceCodeHandler(RequestData requestData, DeviceCodeResult
         {
             TimeSpan timeRemaining = deviceCodeResult.ExpiresOn - DateTimeOffset.UtcNow;
             AuthenticationResultEx resultEx = null;
-            while (timeRemaining.TotalSeconds > 0)
+            while (timeRemaining.TotalSeconds + deviceCodeResult.Interval > 0)
```
trwalke: The loop ended before the final request was sent resulting in a null resultEx. Now one more request will be sent to get the final error message from the STS.

Issue #826, opened 2017-09-27
Ensure AcquireTokenSilent tests exist in ADAL.NET for public clients


trwalke:
Ensure AcquireTokenSilent tests exist in ADAL.NET for public clients. Also ensure that ADAL returns correct exception and inner exceptions in case of http failures
VSTS 292916

AC
Ensure that we have --
- Test case where a valid access token exists in the cache.
- Test case where a expired access token and a valid refresh token exist in cache. ADAL should acquire a new token using RT and override old cache entry.
- Test case where cache is empty. ADAL should return error.
- Test case where a expired access token and a invalid refresh token exist in cache. ADAL should fail to acquire a new token using RT and STS should return invalid_grant.

PR #827, opened 2017-09-27
Created a test case where a valid access token exists in the cache fo…

trwalke:
…r AcquireTokenAsync.

#826


tests/Test.ADAL.NET.Integration/AcquireTokenSilentTests.cs
```diff
@@ -0,0 +1,113 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Test.ADAL.Common;
+using Test.ADAL.NET.Unit.Mocks;
+
+namespace Test.ADAL.NET.Unit
```
kpanwar: Update the namespace to integration instead of unit

tests/Test.ADAL.NET.Integration/AcquireTokenSilentTests.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        }
+
+        [TestMethod]
+        [TestCategory("AcquireTokenSilentTests")]
+        //292916 Ensure AcquireTokenSilent tests exist in ADAL.NET for public clients
+        public async Task AcquireTokenSilentTestWithValidTokenInCache()
+        {
+            var context = new AuthenticationContext(TestConstants.DefaultAuthorityHomeTenant, true);
+            context.TokenCache.Clear();
```
kpanwar: You can choose to pass in a cache instance via constructor instead. That will make the test resilient to such issues in the future.

tests/Test.ADAL.NET.Integration/PromptBehaviorTests.cs
```diff
@@ -67,6 +67,7 @@ public void Initialize()
             });
 
             var context = new AuthenticationContext(TestConstants.DefaultAuthorityHomeTenant, true);
+            context.TokenCache.Clear();
```
kpanwar: Same feedback as above. This applies to all changes

PR #832, opened 2017-09-28
Authority Migration: Use preferred network to talk to endpoints

rayluo:
This PR implements this behavior:

When an authority is valid and metadata is returned, preferred_network host is used for all subsequent network requests.

UPDATE at Oct 6: We realized that this PR addresses only the token request scenario. In ADAL dotnet, there are a Device Code request class and a UserRealm discovery process that need to pick up similar change. Those are not currently implemented in this PR.

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
@@ -332,10 +332,21 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 
         private async Task<AuthenticationResultEx> SendHttpMessageAsync(IRequestParameters requestParameters)
         {
-            client = new AdalHttpClient(this.Authenticator.TokenUri, this.CallState)
+            var metadata = await InstanceDiscovery.GetMetadataEntry(
+                new Uri(Authenticator.TokenUri).Host, Authenticator.ValidateAuthority, this.CallState).ConfigureAwait(false);
+            client = new AdalHttpClient(ReplaceHost(Authenticator.TokenUri, metadata.PreferredNetwork), this.CallState)
                 {Client = {BodyParameters = requestParameters}};
             TokenResponse tokenResponse = await client.GetResponseAsync<TokenResponse>().ConfigureAwait(false);
             return tokenResponse.GetResult();
+        }
+
+        private static string ReplaceHost(string oldUri, string newHost)
+        {
+            if (string.IsNullOrEmpty(oldUri) || string.IsNullOrEmpty(newHost))
+            {
+                throw new ArgumentNullException();
+            }
+            return $"https://{newHost}{new Uri(oldUri).AbsolutePath}";
         }
 
         private void NotifyBeforeAccessCache()
--- a/tests/Test.ADAL.NET.Unit/InstanceDiscoveryTests.cs
+++ b/tests/Test.ADAL.NET.Unit/InstanceDiscoveryTests.cs
@@ -103,7 +103,7 @@ namespace Test.ADAL.NET.Unit
 
         [TestMethod]
         [TestCategory("InstanceDiscoveryTests")]
-        public async Task TestInstanceDiscovery_WhenMultipleSimultaneousCallsWithTheSameAuthority_ShouldMakeOnlyOneRequest()
+        public void TestInstanceDiscovery_WhenMultipleSimultaneousCallsWithTheSameAuthority_ShouldMakeOnlyOneRequest()
         {
             for (int i = 0; i < 2; i++) // Prepare 2 mock responses
             {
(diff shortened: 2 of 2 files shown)


tests/Test.ADAL.NET.Unit/InstanceDiscoveryTests.cs
```diff
@@ -157,5 +157,59 @@ public void TestInitialize()
...
+
+            var privateObject = new PrivateObject(new AcquireTokenForClientHandler(new RequestData
+            {
+                Authenticator = authenticator,
+                Resource = "resource1",
+                ClientKey = new ClientKey(new ClientCredential("client1", "something")),
+                SubjectType = TokenSubjectType.Client,
+                ExtendedLifeTimeEnabled = false
+            }));
+            await (Task) privateObject.Invoke("SendTokenRequestAsync");
```
kpanwar: why is this method being invoke via reflection? you could run runHandlerAsync(); instead

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -332,12 +332,23 @@ protected virtual void PostTokenRequest(AuthenticationResultEx result)
 
         private async Task<AuthenticationResultEx> SendHttpMessageAsync(IRequestParameters requestParameters)
         {
-            client = new AdalHttpClient(this.Authenticator.TokenUri, this.CallState)
+            var metadata = await InstanceDiscovery.GetMetadataEntry(
+                new Uri(Authenticator.TokenUri).Host, Authenticator.ValidateAuthority, this.CallState).ConfigureAwait(false);
+            client = new AdalHttpClient(ReplaceHost(Authenticator.TokenUri, metadata.PreferredNetwork), this.CallState)
```
kpanwar: would it not be easier to make this decision in AdalHttpClient class? That way the request classes stay agnostic to details of the metadata and the change applies to all cases?

tests/Test.ADAL.NET.Unit/InstanceDiscoveryTests.cs
rayluo: Assuming you mean the `RunAsync()` method [defined here](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/9073fe37b6cb9f3d02436d0ce8103345c9e8a29e/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs#L117), yes I actually considered it too, and that would also work.

I prefer to not use it, because `RunAsync()`, which is basically the only available public methods in the `AcquireTokenHandlerBase` family, indirectly wraps the entire 400+ lines AcquireTokenHandlerBase behaviors: PreRunAsync(), PostRunAsync(), AcquireTokenUsingBroker(), CheckAndAcquireTokenUsingBroker(), SendTokenRequestAsync(), SendHttpMessageAsync(), to name a few. That means any changes in those helpers in the future could potentially break this test case. Now, given the fact that this PR actually only implements the new behavior inside SendHttpMessageAsync(), it seems more desirable to minimize blast radius to the scope of adjustments, which in this case happens to be a private method.

Whether or how we should unit test private methods is an inspiring topic in itself. Feel free to read [this stackoverflow question, its answers AND its comments](https://stackoverflow.com/questions/9122708/unit-testing-private-methods-in-c-sharp).

kpanwar: The method is never called outside the pipeline that you described. this should be tested as a integration test from top to bottom. your test approach would be more effective if this functionality was in the AdalHttpClient as suggested below. then you could write tests hitting the adalhttpclient in various degrees

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
rayluo: That was also my first reaction when I started considering on this direction. But then something pops up:

* There are 7 files in the "src/.../Http" folder of this repo, and at least 2 files seem relevant, the [AdalHttpClient](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/rayluo/lwn-mso-migration-preferred-network-2/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Http/AdalHttpClient.cs) and its underlying helper, the [HttpClientWrapper](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/rayluo/lwn-mso-migration-preferred-network-2/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Http/HttpClientWrapper.cs). So one of them could be a candidate to receive this PR change. But I/we can not easily decide which one, if at all, is the suitable place to receive changes, without reading through the entire 33000+ line code base to know what other behaviors/components in this code base are also using those HttpClient classes.
* In fact I already know the InstanceDiscovery is using AdalHttpClient to talk to a given host. If we would change the AdalHttpClient behavior, that could mean the InstanceDiscovery host would also be alterred to the PreferredNetwork. A subtle change of behavior, and probably not what we want.

So I go with the safer way, to add the PreferredNetwork behavior into the exact place where we want it happen, rather than elsewhere.

tests/Test.ADAL.NET.Unit/InstanceDiscoveryTests.cs
rayluo: Based on the recently discovered issues caused by test cases unnecessarily coupling each other, it seems like the **UNIT** tests should better stay away from "this should be tested as a integration test from top to bottom" approach. If every unit test case would indirectly cover (and therefore be potentially affected by) more than necessary scope, it won't be scalable in the long term. (We can talk in person.)

kpanwar: Like I said before, refactoring the change into AdalHttpClient would warrant for unit tests (still mocked for network calls) as that class gets hit from multiple points in the code. Do you agree?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
kpanwar: InstanceDiscovery should use HttpClientWrapper directly and everything else will fall in place if you move this logic in AdalHttpClient

rayluo:
We are now reconsidering PR #824

PR #833, opened 2017-09-28
Created a test case where cache is empty. ADAL should return error.

trwalke:
Created a test case where cache is empty. ADAL should return error.
Created a test case where a expired access token and a invalid refresh token exist in cache.
#826

### DIFF
--- a/tests/Test.ADAL.NET.Integration/AcquireTokenSilentTests.cs
+++ b/tests/Test.ADAL.NET.Integration/AcquireTokenSilentTests.cs
@@ -28,6 +28,8 @@
 using Microsoft.IdentityModel.Clients.ActiveDirectory;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
+using System.Collections.Generic;
+using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Test.ADAL.Common;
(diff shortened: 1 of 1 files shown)


tests/Test.ADAL.NET.Integration/AcquireTokenSilentTests.cs
```diff
@@ -106,5 +108,59 @@ public void AcquireTokenSilentServiceErrorTestAsync()
...
+        [TestCategory("AcquireTokenSilentTests")]
+        //292916 Ensure AcquireTokenSilent tests exist in ADAL.NET for public clients
+        public void AcquireTokenSilentWithEmptyCacheTest()
+        {
+            var context = new AuthenticationContext(TestConstants.DefaultAuthorityHomeTenant, true, new TokenCache());
+            AuthenticationResult result;
+            AdalSilentTokenAcquisitionException ex = AssertException.TaskThrows<AdalSilentTokenAcquisitionException>(async () => 
+            result = await context.AcquireTokenSilentAsync(TestConstants.DefaultResource, TestConstants.DefaultClientId, new UserIdentifier("unique_id", UserIdentifierType.UniqueId)));
+
+            Assert.AreEqual(ex.ErrorCode, AdalError.FailedToAcquireTokenSilently);
```
kpanwar: assert that inner exception is null

tests/Test.ADAL.NET.Integration/AcquireTokenSilentTests.cs
```diff
@@ -106,5 +108,59 @@ public void AcquireTokenSilentServiceErrorTestAsync()
...
+            {
+                RefreshToken = "some-rt",
+                ResourceInResponse = TestConstants.DefaultResource,
+                Result = new AuthenticationResult("Bearer", "existing-access-token",
+                    DateTimeOffset.UtcNow)
+            };
+
+            AuthenticationResult result = null;
+
+            AdalSilentTokenAcquisitionException ex = AssertException.TaskThrows<AdalSilentTokenAcquisitionException>(async () =>
```
kpanwar: assert that inner exception is not null

tests/Test.ADAL.NET.Integration/AcquireTokenSilentTests.cs
```diff
@@ -106,5 +108,59 @@ public void AcquireTokenSilentServiceErrorTestAsync()
...
+        {
+            var context = new AuthenticationContext(TestConstants.DefaultAuthorityHomeTenant, new TokenCache());
+
+            HttpMessageHandlerFactory.AddMockHandler(new MockHttpMessageHandler()
+            {
+                Method = HttpMethod.Post,
+                ResponseMessage = MockHelpers.CreateInvalidGrantTokenResponseMessage(),
+                PostData = new Dictionary<string, string>()
+                {
+                    {"client_id", TestConstants.DefaultClientId},
```
kpanwar: add RefreshToken = "some-rt" as a part of the postdata validation

trwalke: @kpanwar is this what you mean?
                    { "client_id", TestConstants.DefaultClientId},
                    {"grant_type", "refresh_token"},
                    {"refresh_token", "some-rt" }

there is no property named RefreshToken to set

Issue #840, opened 2017-10-05
Ensure Managed User Credentials integration tests exist in ADAL.NET


trwalke:
Ensure that we have -
For MANAGED USER
==============
- Refactor existing tests to integration project
- Test case with empty cache where network call is made to user realm endpoint and then to token endpoint to get token from AAD
- Test case where valid access token already exists in cache. this should result in cache token returned back.
- Test case where valid access token for user1 already exists in cache. API is called for user2 and in this case network call is made to user realm endpoint and then to token endpoint to get token from AAD. Now the cache should have 2 entries.
- Test case with expired access token and valid refresh token in cache. This should result in refresh token being used to get new AT instead of user creds.
- Test case where user realm discovery fails  because user does not exist in AAD.
- Test case where user realm discovery cannot determine the user type
VSTS 292990

PR #849, opened 2017-10-10
Ensure Managed User Credentials integration tests exist in ADAL.NET PR

trwalke:
Test case where valid access token already exists in cache. this should result in cache token returned back.

Test case where valid access token for user1 already exists in cache. API is called for user2 and in this case network call is made to user realm endpoint and then to token endpoint to get token from AAD. Now the cache should have 2 entries.

Test case with expired access token and valid refresh token in cache. This should result in refresh token being used to get new AT instead of user creds.

Test case where user realm discovery fails because user does not exist in AAD.

Test case where user realm discovery cannot determine the user type

#840
VSTS 292990

### DIFF
--- a/tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
+++ b/tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
@@ -1,16 +1,19 @@
 ﻿using Microsoft.IdentityModel.Clients.ActiveDirectory;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
 using System.Collections.Generic;
-using System.IdentityModel.Tokens;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
-using System.Text;
 using System.Threading.Tasks;
 using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Http;
+using Test.ADAL.Common;
 using Test.ADAL.NET.Common;
 using Test.ADAL.NET.Common.Mocks;
 using AuthenticationContext = Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationContext;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Cache;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.OAuth2;
 
 namespace Test.ADAL.NET.Integration
 {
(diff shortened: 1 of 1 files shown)


tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -69,5 +72,212 @@ public void Initialize()
             Assert.AreEqual(TestConstants.DefaultDisplayableId, result.UserInfo.DisplayableId);
             Assert.AreEqual(TestConstants.DefaultUniqueId, result.UserInfo.UniqueId);
         }
+
+        [TestMethod]
+        [Description("Positive Test for AcquireToken with a token already in cache")]
+        public async Task AcquireTokenWithValidTokenInCachePositiveTestAsync()
```
kpanwar: remove positive test. add _ReturnsCachedToken

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -69,5 +72,212 @@ public void Initialize()
...
+        {
+            HttpMessageHandlerFactory.AddMockHandler(new MockHttpMessageHandler()
```
kpanwar: for valid token in cache, no network call is made to user realm. you should update all tests to ensure that all mocks were consumed.

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -69,5 +72,212 @@ public void Initialize()
...
+
+            Assert.IsNotNull(result);
+            Assert.AreEqual("existing-access-token", result.AccessToken);
+            Assert.AreEqual(TestConstants.DefaultAuthorityHomeTenant, context.Authenticator.Authority);
+            Assert.IsNotNull(result.UserInfo);
+        }
+
+        [TestMethod]
+        [Description("Positive Test for AcquireToken for a user when a valid access token already exists in cache for another user.")]
+        public async Task AcquireTokenWithValidAccessTokenInCacheForAnotherUserPositiveTestAsync2()
```
kpanwar: remove 2. rename to _GetsNewTokenFromService

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -69,5 +72,212 @@ public void Initialize()
...
+                    context.AcquireTokenAsync(TestConstants.DefaultResource, TestConstants.DefaultClientId,
+                        new UserPasswordCredential("user2@id.com", TestConstants.DefaultPassword));
+
+            Assert.IsNotNull(result);
+            Assert.AreEqual(TestConstants.DefaultAuthorityHomeTenant, context.Authenticator.Authority);
+            Assert.AreEqual(result.AccessToken, "some-access-token");
+            Assert.IsNotNull(result.UserInfo);
+            Assert.AreEqual("user2@id.com", result.UserInfo.DisplayableId);
+            Assert.AreEqual(TestConstants.DefaultUniqueId, result.UserInfo.UniqueId);
+            Assert.AreNotEqual(context.TokenCache.tokenCacheDictionary.Values.First(), context.TokenCache.tokenCacheDictionary.Values.Last());
```
kpanwar: this order is not guaranteed. so remove this assert

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -69,5 +72,212 @@ public void Initialize()
...
+                QueryParams = new Dictionary<string, string>()
+                {
+                    {"api-version", "1.0"}
+                }
+            });
+
+            HttpMessageHandlerFactory.AddMockHandler(new MockHttpMessageHandler()
+            {
+                Method = HttpMethod.Post,
+                ResponseMessage = MockHelpers.CreateSuccessTokenResponseMessage(TestConstants.DefaultUniqueId, "user2@id.com", TestConstants.DefaultResource),
```
kpanwar: uniqueid should be different. if they are not in the real world, then acquireTokenSilent will fail when you pass uniqueId as an identifier.

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -69,5 +72,212 @@ public void Initialize()
...
+
+            Assert.IsNotNull(result);
+            Assert.AreEqual(TestConstants.DefaultAuthorityHomeTenant, context.Authenticator.Authority);
+            Assert.AreEqual(result.AccessToken, "some-access-token");
+            Assert.IsNotNull(result.UserInfo);
+            Assert.AreEqual("user2@id.com", result.UserInfo.DisplayableId);
+            Assert.AreEqual(TestConstants.DefaultUniqueId, result.UserInfo.UniqueId);
+            Assert.AreNotEqual(context.TokenCache.tokenCacheDictionary.Values.First(), context.TokenCache.tokenCacheDictionary.Values.Last());
+            // There should be only two cache entrys.
+            Assert.AreEqual(2, context.TokenCache.Count);
```
kpanwar: add check to ensure 2 cache entries are for 2 different users. make sure  you check the key and the value.

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -69,5 +72,212 @@ public void Initialize()
...
+        }
+
+        [TestMethod]
+        [Description("Test case with expired access token and valid refresh token in cache. This should result in refresh token being used to get new AT instead of user creds")]
+        public async Task AcquireTokenWithExpiredAccessTokenAndValidRefreshTokenTestAsync()
```
kpanwar: _GetsATUsingRT

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -69,5 +72,212 @@ public void Initialize()
...
+            Assert.AreEqual(TestConstants.DefaultAuthorityHomeTenant, context.Authenticator.Authority);
+            Assert.IsNotNull(result.UserInfo);
+
+            // There should be only one cache entry.
+            Assert.AreEqual(1, context.TokenCache.Count);
+        }
+
+        [TestMethod]
+        [Description("Test case where user realm discovery fails.")]
+        public async Task UserRealmDiscoveryNegativeTestAsync()
```
kpanwar: FailsTestAsync

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -69,5 +72,212 @@ public void Initialize()
...
+        {
+            AuthenticationContext context = new AuthenticationContext(TestConstants.DefaultAuthorityCommonTenant);
+            await context.Authenticator.UpdateFromTemplateAsync(null);
+
+            HttpMessageHandlerFactory.AddMockHandler(new MockHttpMessageHandler()
+            {
+                Method = HttpMethod.Get,
+                ResponseMessage = new HttpResponseMessage(HttpStatusCode.BadRequest)
+                {
+                    Content = new StringContent("{ \"error\":\"user realm discovery failed\", \"ver\":\"1.0\",\"account_type\":\"Unknown\", \"cloudinstancename\":\"login.microsoftonline.com\"}")
```
kpanwar: bad requests do not return JSON typically. just use string like "Bad Request Received".

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -69,5 +72,212 @@ public void Initialize()
...
+                },
+                QueryParams = new Dictionary<string, string>()
+                {
+                    {"api-version", "1.0"}
+                }
+            });
+
+            var ex = AssertException.TaskThrows<AdalException>(() =>
+            UserRealmDiscoveryResponse.CreateByDiscoveryAsync(context.Authenticator.UserRealmUri, TestConstants.DefaultDisplayableId, CallState.Default));
```
kpanwar: these are integration tests. you should be calling acquiretoken API. we only test public API surface here

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -69,5 +72,212 @@ public void Initialize()
...
+            Assert.IsNotNull(ex.ErrorCode, AdalErrorMessage.UserRealmDiscoveryFailed);
+        }
+
+        [TestMethod]
+        [Description("Test case where user realm discovery cannot determine the user type.")]
+        public async Task UnknownUserRealmDiscoveryNegativeTestAsync()
```
kpanwar: remove negative

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -69,5 +72,212 @@ public void Initialize()
...
+
+            Assert.AreEqual("1.0", userRealmResponse.Version);
+            Assert.AreEqual(userRealmResponse.AccountType, "Unknown");
+            Assert.IsNull(userRealmResponse.FederationActiveAuthUrl);
+            Assert.IsNull(userRealmResponse.FederationMetadataUrl);
+            Assert.IsNull(userRealmResponse.FederationProtocol);
+
+
+            var ex = AssertException.TaskThrows<AdalException>(() =>
+                UserRealmDiscoveryResponse.CreateByDiscoveryAsync(context.Authenticator.UserRealmUri, null, CallState.Default));
```
kpanwar: call the acquireToken API instead

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -27,7 +57,7 @@ public void Initialize()
 
         [TestMethod]
         [Description("Positive Test for AcquireToken with an empty cache")]
-        public async Task AcquireTokenWithEmptyCachePositiveTestAsync()
+        public async Task AcquireTokenWithEmptyCache_GetsNewTokenFromService()
```
kpanwar: update all tests to ensure that mocks were consumed. there is check for that in other tests that you can reference

PR #851, opened 2017-10-10
Authority Migration: TokenCache read/write changes (implemented inside TokenCache class)

rayluo:
Functionality:

* Use the preferred_cache, given host, and aliases from the Authority validation cache to look up tokens in the token cache
* Use the preferred_cache when writing tokens into cache

Difference between | PR #816 | This PR
--- | ------- | --------
Functionality | See above | Same
Implemented in | class AcquireTokenHandlerBase | class TokenCache
Files changed | 2 | 7 (Because the `LoadFromCache()` and `StoreToCache()` become async methods now, we do need to adjust all pre-existing test cases 

Pick the one you like to review. Once merged, we will close the other PR.

### DIFF
--- a/automation/WinFormsAutomationApp/AuthenticationHelper.cs
+++ b/automation/WinFormsAutomationApp/AuthenticationHelper.cs
@@ -118,7 +118,7 @@ namespace WinFormsAutomationApp
                     {
                         var updated = item;
                         updated.Value.Result.ExpiresOn = DateTime.UtcNow;
-                        await UpdateCache(item, updated);
+                        await UpdateCache(item, updated).ConfigureAwait(false);
                     }
                 }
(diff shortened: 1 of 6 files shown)


automation/WinFormsAutomationApp/AuthenticationHelper.cs
```diff
@@ -118,7 +118,7 @@ internal static class AuthenticationHelper
                     {
                         var updated = item;
                         updated.Value.Result.ExpiresOn = DateTime.UtcNow;
-                        UpdateCache(item, updated);
+                        await UpdateCache(item, updated);
```
kpanwar: all await calls should have ConfigureAwait(false);

rayluo: Ah. My bad. New commit is on the way!

Issue #863, opened 2017-10-17
Cannot pack fork into nuget package


kevingoos:
I run the following command on the Developer Command Prompt to try to pack my fork into a nuget package.
`nuget Pack src\Microsoft.IdentityModel.Clients.ActiveDirectory\Microsoft.IdentityModel.Clients.ActiveDirectory.csproj`

**And this is the error I get:**
`Unable to find 'bin\Debug\Microsoft.IdentityModel.Clients.ActiveDirectory\bin\Debug\'. Make sure the project has been built.`

But **the project is build succesfull and contains the following folder in the debug folder:**
![folder](https://user-images.githubusercontent.com/1328773/31671100-739f42aa-b359-11e7-946d-1450855679a2.PNG)

kevingoos:
I sombody can explain this to me I will also add this info to the documentation I wrote: [Documentation](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/wiki/ADAL-.NET-Build-&-Run)

kpanwar:
pack command is not really supported because our nuget package also includes netcore45 target that is build separately and merged into a single package.

kevingoos:
@kpanwar is there then an other way to actually create a nuget package for xamarin?

PR #871, opened 2017-10-20
Create accessibility to the cookiecontainer in the httpclient

kevingoos:
Because I need to access the cookie in one of my project to get access to the internal network I added this part of code to get access to the low level cookiecontainer.


kpanwar:
We cannot take this PR because it adds a surface area that could potentially be a security issue as it would enable the ability to transplant the cookies. The library would end up establishing an implicit contract to to become cookie aware when it is not today. This could lead to a plethora a future issues which don't exist today (like channel/token binding). You could consider forking the project for your needs, but we don't support the forks and the onus would be upon you to sync back with this repo.

kevingoos:
Ok I understanc, I already made a fork, but the problem now is that I cannot get it to work to actually create the nuget package to install into my project. See [issue](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/863)

kevingoos:
@kpanwar I understand the security issue. But I want to say that this issue is not specific to my project. (I got the same request from another company) This is issue has to do with having a reverse proxy device on the network that works also with the same ADFS authentication. 

So is there maybe an other secure way implement this functionality so that the library actually can work with Reverse Proxy authentication?

I also made a fork, and I am working on testing the fork now. I don't have a problem with keeping a fork for this functionality (only a little bit annoying for future updates of course). But the problem is that I cannot get the package created, to put it on my private nuget, because I don't know what the alternative command you guys use to create the package. I already got an issue closed on this, but it doesn't give info on how to actually do it, see [issue](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/863)

Issue #855, opened 2017-10-11
Re-enable sovereign cloud support for 3.17.1


kpanwar:
Re-enable sovereign cloud support for 3.17.1

jmprieur:
blocked on the fact that we don't have a cloud environment where we can test end to end.

PR #873, opened 2017-10-24
Jennyf/reenable sovereign cloud

jennyf19:
#855


tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
```diff
@@ -44,11 +78,55 @@ public void Initialize()
 
             HttpMessageHandlerFactory.AddMockHandler(new MockHttpMessageHandler
             {
+                Method = HttpMethod.Get,
```
kpanwar: add another test case where, another instance discovery call is made because the authority was not already in the instance cache

jennyf19: I added in a test case, but not sure if the implementation is correct.

PR #876, opened 2017-10-24
Authority Migration implementation includes tenant info now

rayluo:


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
```diff
@@ -81,19 +82,19 @@ internal static class InstanceDiscovery
 
         private static SemaphoreSlim semaphore = new SemaphoreSlim(1, 1);
 
-        public static async Task<InstanceDiscoveryMetadataEntry> GetMetadataEntry(string host, bool validateAuthority,
+        public static async Task<InstanceDiscoveryMetadataEntry> GetMetadataEntry(Uri authority, bool validateAuthority,
             CallState callState)
         {
             InstanceDiscoveryMetadataEntry entry = null;
-            if (!InstanceCache.TryGetValue(host, out entry))
+            if (!InstanceCache.TryGetValue(authority.Host, out entry))
```
brentschmaltz: we need to check for null on authority.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
```diff
@@ -110,15 +111,19 @@ public static string FormatAuthorizeEndpoint(string host, string tenant)
             return string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/oauth2/authorize", host, tenant);
         }
 
+        public static string GetTenant(Uri uri)
+        {
+            return uri.AbsolutePath.Split('/')[1];  // Will generate exception when uri contains no tenant
```
brentschmaltz: we should have parameter checking here.
null check
split.length >=1

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
```diff
@@ -81,19 +82,19 @@ internal static class InstanceDiscovery
...
 
-        public static async Task<InstanceDiscoveryMetadataEntry> GetMetadataEntry(string host, bool validateAuthority,
+        public static async Task<InstanceDiscoveryMetadataEntry> GetMetadataEntry(Uri authority, bool validateAuthority,
             CallState callState)
         {
             InstanceDiscoveryMetadataEntry entry = null;
-            if (!InstanceCache.TryGetValue(host, out entry))
+            if (!InstanceCache.TryGetValue(authority.Host, out entry))
             {
                 await semaphore.WaitAsync().ConfigureAwait(false); // SemaphoreSlim.WaitAsync() will not block current thread
```
jennyf19: I'm curious as to why you are using a [concurrent dictionary](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=netframework-4.7.1), which is thread safe, _and_ using a semaphore, during which time a network call is occurring. I think you could just use the concurrent dictionary, since you are only reading and writing to the instance cache. It might be better to avoid performing time consuming operations (ex. a network call) while holding a lock.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
rayluo: I can add some `if(...) throw new SomeException(...)` at the beginning of this function, but before that I just want to make sure: Do we use such pattern simply because it is a convention?

In this particular case, the current implementation is a one-liner. So, if the input `uri` happens to be `null`, it would automatically throw a `System.NullReferenceException` and the stack trace would be at exact the same line. Also if for whatever reason (more on this later) the "split.length" is less than 2, the `...[1]` part would also automatically throw a `System.IndexOutOfRangeException` and again the stack trace would be at the same line.

brentschmaltz: Throwing a System.NullReferenceException is not good practice, in my opinion, and usually means a programming error. 
I think it is better to recognize that the uri is not well formed and pass a meaningful message back to the user.

kpanwar: This uri path validation happens way before Instance discovery is called. If we are here in the code, then the input was already deemed sane.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
rayluo: @jennyf19 Good question! I still remembered Brent and I spend quite some time discussing this tricky topic.

Long story short, the `ConcurrentDictionary` ensures its data can be safely accessed *via its methods* in multithread environment. In our case, the first `TryGetValue(...)` can safely try to get a value no matter what.

But `ConcurrentDictionary` does not and will not regulate how *other* code path prepare/consume those data. In our case, we want to make sure when the value does not exist, there will be one and only one network call occurring (more on this later). Such network call i.e. `DiscoverAsync(...)` happens outside of Dictionary methods, so we have to use a dedicate lock for it.

Also, in most other places, it is true that we won't want to hold a lock for a time consuming operations such as I/O calls, but in this case we happen to have a handy `SemaphoreSlim.WaitAsync()` which "locks" asynchronizely, so it does not actully block anything.

Lastly, ideally you would like to avoid such tricky requirement in the first place. Arguably, we could have relax the "only one network call at a time" requirement, and then such explicit lock won't be necessary. But that is a different topic anyway.

jennyf19: Thanks for the explanation, @rayluo

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -362,7 +361,7 @@ internal void OnBeforeWrite(TokenCacheNotificationArgs args)
             return resultEx;
         }
 
-        private string GetHost(string uri)
+        public static string GetHost(string uri)
```
brentschmaltz: If this is public, we need to validate parameters.

kpanwar:
@rayluo  - I see that you have addressed all of brent's comments. I will dismiss @brentschmaltz's review to unblock your commit.

PR #877, opened 2017-10-25
adding PiiLoggingEnabled flag, refactoring

SomkaPe:


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Platform/LoggerBase.cs
```diff
@@ -47,9 +60,49 @@ internal static string GetCallerFilename(string callerFilePath)
...
+        }
+
+        internal void WarningPii(CallState callState, string message)
+        {
+            Log(callState, LogLevel.Warning, message, true);
+        }
+
+        internal void Error(CallState callState, Exception ex)
+        {
+            Log(callState, LogLevel.Error, ex.ToString(), false);
```
brentschmaltz: ex.ToString may have PII, it is hard to tell. Should we use ex.GetType() when PII is off.

SomkaPe: I feel like ex.GetType()  is not quite informative, may be even almost useless for troubleshooting,
taking to account that PII logging is disabled by default It might be a problem

src/Microsoft.IdentityModel.Clients.ActiveDirectory/IAdalLogCallback.cs
```diff
@@ -66,7 +66,7 @@ public interface IAdalLogCallback
         /// </summary>
         /// <param name="level">Log level</param>
         /// <param name="message">message to be logged</param>
-        void Log(LogLevel level, string message);
```
kpanwar: this is an API breaking change. we need to come up with other ways to support this.

kpanwar:
@SomkaPe - As part of these changes, we need to make sure that Pii logs are NEVER written to default output

src/Microsoft.IdentityModel.Clients.ActiveDirectory/IAdalLogCallback.cs
```diff
@@ -106,11 +119,37 @@ public static IAdalLogCallback Callback
...
+            {
+                _localCallback?.Log(level, message);
+            }
+        }
+
+        private static volatile LogCallback _logCallback;
+
+        /// <summary>
+        ///     Callback instance that can be provided by the developer to consume and publish logs in a custom manner.
+        ///     The property can only be set once and it will throw an ArgumentException if called twice.
```
kpanwar: I don't see the code throwing on calling twice.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/IAdalLogCallback.cs
```diff
@@ -106,11 +119,37 @@ public static IAdalLogCallback Callback
...
+            lock (LockObj)
+            {
+                _localCallback?.Log(level, message);
+            }
+        }
+
+        private static volatile LogCallback _logCallback;
+
+        /// <summary>
+        ///     Callback instance that can be provided by the developer to consume and publish logs in a custom manner.
```
kpanwar: add documentation that if delegate is configured then obsolete callback will not be executed.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Platform/LoggerBase.cs
```diff
@@ -50,7 +50,15 @@ internal abstract class LoggerBase
...
 
-            LoggerCallbackHandler.ExecuteCallback(logLevel, formattedMessage, containsPii);
+            if (LoggerCallbackHandler.LogCallback != null)
+            {
+                LoggerCallbackHandler.ExecuteCallback(logLevel, formattedMessage, containsPii);
+            }
+            else if (!containsPii)
+            {
+                // execute obsolete IAdalLogCallback only if LogCallback is not set and message does not contain Pii
+                LoggerCallbackHandler.ExecuteCallback(logLevel, formattedMessage);
```
kpanwar: code is confusing to read. We should either use different name for methods of rename to old function to contain the word obsolete.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Extensions.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Globalization;
+using System.Text;
+
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory
+{
+    internal static class Extensions
+    {
+        internal static string PiiLessToString(this Exception ex)
```
kpanwar: rename to something clearer. like GetPIIScrubbedDetails?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Extensions.cs
```diff
@@ -0,0 +1,44 @@
...
+                {
+                    sb.Append(string.Format(CultureInfo.CurrentCulture, ", StatusCode: {0}", adalServiceException.StatusCode));
+                }
+
+                if (ex.InnerException != null)
+                {
+                    sb.Append(" ---> " + PiiLessToString(ex.InnerException) + Environment.NewLine +
+                              "--- End of inner exception stack trace ---");
+                }
+                if (ex.StackTrace != null)
```
kpanwar: add line break

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -61,15 +61,22 @@ protected AcquireTokenHandlerBase(RequestData requestData)
...
-            CallState.Logger.Information(this.CallState,
-                string.Format(CultureInfo.CurrentCulture,
-                    "=== Token Acquisition started:\n\tAuthority: {0}\n\tResource: {1}\n\tClientId: {2}\n\tCacheType: {3}\n\tAuthentication Target: {4}\n\t",
-                    requestData.Authenticator.Authority, requestData.Resource, requestData.ClientKey.ClientId,
-                    (tokenCache != null)
-                        ? tokenCache.GetType().FullName +
-                          string.Format(CultureInfo.CurrentCulture, " ({0} items)", tokenCache.Count)
-                        : "null",
-                    requestData.SubjectType));
+            if (LoggerCallbackHandler.PiiLoggingEnabled)
```
kpanwar: we should not need to check for every log. we should be able to just call the method and if PII Logging is enabled then it gets logged.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -61,15 +61,22 @@ protected AcquireTokenHandlerBase(RequestData requestData)
...
+            {
+                CallState.Logger.InformationPii(this.CallState,
+                    string.Format(CultureInfo.CurrentCulture,
+                        "=== Token Acquisition started: \n\tAuthority: {0}\n\tResource: {1}\n\tClientId: {2}\n\tCacheType: {3}\n\tAuthentication Target: {4}\n\t",
```
kpanwar: this log content is not PII. we can check the onenote together.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -376,14 +383,21 @@ private void LogReturnedToken(AuthenticationResult result)
...
-                        accessTokenHash,
-                        result.ExpiresOn,
-                        result.UserInfo != null
-                            ? CryptographyHelper.CreateSha256Hash(result.UserInfo.UniqueId)
-                            : "null"));
+                if (LoggerCallbackHandler.PiiLoggingEnabled)
+                {
+                    CallState.Logger.InformationPii(this.CallState,
+                        string.Format(CultureInfo.CurrentCulture,
+                            "=== Token Acquisition finished successfully. An access token was returned: Access Token Hash: {0}\n\tExpiration Time: {1}\n\tUser id: {2}\n\t",
```
kpanwar: update the log message to not log access token hash and uniqueID. these values are not very helpful anyways.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Platform/LoggerBase.cs
```diff
@@ -27,18 +27,52 @@
...
     internal abstract class LoggerBase
     {
-        internal abstract void Verbose(CallState callState, string message, [System.Runtime.CompilerServices.CallerFilePath] string callerFilePath = "");
+        internal abstract void DefaultLog(LogLevel logLevel, string message);
 
-        internal abstract void Information(CallState callState, string message, [System.Runtime.CompilerServices.CallerFilePath] string callerFilePath = "");
+        private void Log(CallState callState, LogLevel logLevel, string message, bool containsPii,
+            [CallerFilePath] string callerFilePath = "")
+        {
+            if (!LoggerCallbackHandler.PiiLoggingEnabled && containsPii)
```
kpanwar: you already have this method that does the check. you should not need if conditions above.

tests/Test.ADAL.NET.Unit/Test.ADAL.NET.Unit.csproj
```diff
@@ -45,6 +45,9 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="NSubstitute, Version=2.0.3.0, Culture=neutral, PublicKeyToken=92dd2e9066daa5ca, processorArchitecture=MSIL">
```
kpanwar: this reference should not be needed.

tests/Test.ADAL.NET.Unit/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="NSubstitute" version="2.0.3" targetFramework="net462" />
```
kpanwar: this file is not needed.

tests/Test.ADAL.NET.Unit/Test.ADAL.NET.Unit.csproj
SomkaPe: build fails If I delete this

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Platform/LoggerBase.cs
SomkaPe: changed

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/TokenCachePlugin.cs
```diff
@@ -57,14 +57,8 @@ public static void BeforeAccess(TokenCacheNotificationArgs args)
...
-                if (LoggerCallbackHandler.PiiLoggingEnabled)
-                {
-                    CallState.Default.Logger.WarningPii(null, "Failed to load cache: " + ex.ToString());
-                }
-                else
-                {
-                    CallState.Default.Logger.Warning(null, "Failed to load cache: " + ex.PiiLessToString());
-                }
+                CallState.Default.Logger.Warning(null, "Failed to load cache: ");
+                CallState.Default.Logger.ErrorPii(null, ex);
```
kpanwar: add "Failed to load cache" in the message like before

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/TokenCachePlugin.cs
```diff
@@ -91,14 +85,8 @@ public static void AfterAccess(TokenCacheNotificationArgs args)
...
-                    if (LoggerCallbackHandler.PiiLoggingEnabled)
-                    {
-                        CallState.Default.Logger.WarningPii(null, "Failed to save cache: " + ex.ToString());
-                    }
-                    else
-                    {
-                        CallState.Default.Logger.Warning(null, "Failed to save cache: " + ex.PiiLessToString());
-                    }
+                    CallState.Default.Logger.Warning(null, "Failed to save cache: ");
+                    CallState.Default.Logger.ErrorPii(null, ex);
```
kpanwar: add "Failed to save cache" in Pii

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/TokenCachePlugin.cs
```diff
@@ -71,14 +71,8 @@ public static void BeforeAccess(TokenCacheNotificationArgs args)
...
-                if (LoggerCallbackHandler.PiiLoggingEnabled)
-                {
-                    CallState.Default.Logger.WarningPii(null, "Failed to load cache: " + ex.ToString());
-                }
-                else
-                {
-                    CallState.Default.Logger.Warning(null, "Failed to load cache: " + ex.PiiLessToString());
-                }
+                CallState.Default.Logger.Warning(null, "Failed to load cache: ");
+                CallState.Default.Logger.ErrorPii(null, ex);
```
kpanwar: same as above

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/TokenCachePlugin.cs
```diff
@@ -120,14 +114,8 @@ public static void AfterAccess(TokenCacheNotificationArgs args)
...
-                    if (LoggerCallbackHandler.PiiLoggingEnabled)
-                    {
-                        CallState.Default.Logger.WarningPii(null, "Failed to save cache: " + ex.ToString());
-                    }
-                    else
-                    {
-                        CallState.Default.Logger.Warning(null, "Failed to save cache: " + ex.PiiLessToString());
-                    }
+                    CallState.Default.Logger.Warning(null, "Failed to save cache: ");
+                    CallState.Default.Logger.ErrorPii(null, ex);
```
kpanwar: same as above

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/winrt/TokenCachePlugin.cs
```diff
@@ -61,14 +61,8 @@ public static void BeforeAccess(TokenCacheNotificationArgs args)
...
-                    if (LoggerCallbackHandler.PiiLoggingEnabled)
-                    {
-                        CallState.Default.Logger.WarningPii(null, "Failed to load cache: " + ex.ToString());
-                    }
-                    else
-                    {
-                        CallState.Default.Logger.Warning(null, "Failed to load cache: " + ex.PiiLessToString());
-                    }
+                    CallState.Default.Logger.Warning(null, "Failed to load cache: ");
+                    CallState.Default.Logger.ErrorPii(null, ex);
```
kpanwar: same as above

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/winrt/TokenCachePlugin.cs
```diff
@@ -87,14 +81,8 @@ public static void AfterAccess(TokenCacheNotificationArgs args)
...
-                    if (LoggerCallbackHandler.PiiLoggingEnabled)
-                    {
-                        CallState.Default.Logger.WarningPii(null, "Failed to save cache: " + ex.ToString());
-                    }
-                    else
-                    {
-                        CallState.Default.Logger.Warning(null, "Failed to save cache: " + ex.PiiLessToString());
-                    }
+                    CallState.Default.Logger.Warning(null, "Failed to save cache: ");
+                    CallState.Default.Logger.ErrorPii(null, ex);
```
kpanwar: same as above

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/TokenCachePlugin.cs
SomkaPe: that message is logged as Warning, add to what ?

SomkaPe: pii logs includes non pii logs , so what is the point of inclusion of "Failed to load cache" message into errorPii ? also ErrorPii accept only one parameter - exception

kpanwar: this is needed because developer can choose to write PII logs in a separate location. I think MS office pushes PII logs to a separate and secure location with security parameters applied it/

SomkaPe: i see,  actually in that case they will have to duplicate all log messages - as pii and as no pii. I would say it might be better to handle this at call back level - so if app developer want to have separate pii and non pii logs, his log delegate implementation might do like: if message does contain pii write message to pii and non-pii logs, if message does contain pii - write it only to pii log

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Extensions.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
```
jennyf19: Remember to add the header info

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Extensions.cs
```diff
@@ -0,0 +1,46 @@
...
+                    var adalException = (AdalException) ex;
+                    sb.Append(string.Format(CultureInfo.CurrentCulture, ", ErrorCode: {0}", adalException.ErrorCode));
+                }
+
+                if (ex is AdalServiceException)
+                {
+                    var adalServiceException  = (AdalServiceException) ex; 
+                    sb.Append(string.Format(CultureInfo.CurrentCulture, ", StatusCode: {0}", adalServiceException.StatusCode));
+                }
+
```
jennyf19: Do you need to do anything about AdalClaimsChallengeException?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Extensions.cs
SomkaPe: good point

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Extensions.cs
SomkaPe: we do not log claims

Issue #875, opened 2017-10-24
Move extension methods to be public instance methods in the primary assembly


kpanwar:
Move extension methods to be public instance methods in the primary assembly. movement of the public instance methods in the primary assembly to be extension methods is a binary breaking change

Issue #874, opened 2017-10-24
Move select account enum to the end of the promptbehavior list


kpanwar:
Move select account enum to the end of the promptbehavior list as this breaks existing code that takes dependency on the order.

PR #878, opened 2017-10-25
Hot fix to resolve breaking changes

kpanwar:
This PR contains fix for -- 
- Move extension methods to be public instance methods in the primary assembly. movement of the public instance methods in the primary assembly to be extension methods is a binary breaking change (#875)
- Move select account enum to the end of the promptbehavior list as this breaks existing code that takes dependency on the order. (#874)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Features/WinCommon/AuthenticationContext.cs
```diff
@@ -36,314 +36,273 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
...
-        public static async Task<Uri> GetAuthorizationRequestUrlAsync(this AuthenticationContext ctx, string resource,
+        public async Task<Uri> GetAuthorizationRequestUrlAsync(string resource,
             string clientId, Uri redirectUri,
             UserIdentifier userId, string extraQueryParameters)
         {
-            return await GetAuthorizationRequestUrlAsync(ctx, resource, clientId, redirectUri, userId,
+            return await GetAuthorizationRequestUrlAsync(resource, clientId, redirectUri, userId,
                 extraQueryParameters,
-                null);
+                null).ConfigureAwait(false);
```
jmprieur: @kpanwar is the addition of .ConfigureAwait(false) directly related to the hotfix?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Features/WinCommon/AuthenticationContext.cs
```diff
@@ -36,314 +36,273 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     /// <summary>
     /// Extension class to support confidential client flows.
```
jmprieur: The comment needs updating. It's no longer an extension class (but a partial implementation) to support confidential client flows

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Features/WinCommon/AuthenticationContext.cs
kpanwar: @jmprieur - not but it should have been there. I just caught it as I was cleaning up the code.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Features/WinCommon/AuthenticationContext.cs
kpanwar: fixed

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Features/WinCommon/AuthenticationContext.cs
jmprieur: Thanks for confirming @kpanwar

AArnott: Tip: you can guard against accidental misses like this by installing the `AsyncUsageAnalyzers` and then setting the `ConfigureAwait` analyzer to Error (or warning, or whatever you want). It's a life saver for library development.

rayluo:
Ship it

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AdalError.cs
```diff
@@ -0,0 +1,280 @@
+﻿//----------------------------------------------------------------------
```
jmprieur: Is it related to the breaking change?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -461,5 +467,336 @@ internal IWebUI CreateWebAuthenticationDialog(IPlatformParameters parameters)
...
+                .ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Acquires security token from the authority.
+        /// </summary>
+        /// <param name="resource">Identifier of the target resource that is the recipient of the requested token.</param>
+        /// <param name="clientCredential">The client credential to use for token acquisition.</param>
+        /// <returns>It contains Access Token and the Access Token's expiration time. Refresh Token property will be null for this overload.</returns>
+        public async Task<AuthenticationResult> AcquireTokenAsync(string resource, ClientCredential clientCredential)
```
jmprieur: Is this a public application flow?

kpanwar: @jmprieur - this is confidential client flow. specifically, this is client credential flow where an app acquires token for itself

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AdalError.cs
kpanwar: No, but chris said that this file was being reported as internal namespace. not sure why, so I moved out the public static class to its own file for safety

PR #884, opened 2017-10-30
Jennyf/reenable sovereign cloud

jennyf19:
#855


tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
```diff
@@ -61,15 +97,103 @@ public void Initialize()
...
+                TestConstants.DefaultClientId,
+                TestConstants.DefaultRedirectUri, platformParameters, UserIdentifier.AnyUser, "instance_aware=true");
+
+            Assert.AreEqual(5, InstanceDiscovery.InstanceCache.Count());
+
+            // One mock remaining
+            Assert.AreEqual(1, HttpMessageHandlerFactory.MockHandlersCount());
+
+            // German authority not included in instance cache
+            Assert.AreEqual(false, InstanceDiscovery.InstanceCache.Keys.Contains("login.microsoftonline.de"));
```
kpanwar: why is this false? DE instance should have been added to the cache

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
```diff
@@ -61,15 +97,103 @@ public void Initialize()
...
+
+            var authenticationResult = 
+                await authenticationContext.AcquireTokenAsync(TestConstants.DefaultResource,
+                TestConstants.DefaultClientId,
+                TestConstants.DefaultRedirectUri, platformParameters, UserIdentifier.AnyUser, "instance_aware=true");
+
+            Assert.AreEqual(5, InstanceDiscovery.InstanceCache.Count());
+
+            // One mock remaining
+            Assert.AreEqual(1, HttpMessageHandlerFactory.MockHandlersCount());
```
kpanwar: no mock should remain.

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
```diff
@@ -61,15 +97,103 @@ public void Initialize()
...
                 authenticationContext.TokenCache.tokenCacheDictionary.Keys.FirstOrDefault().Authority);
-        }*/
+
+            // all mocks are consumed
+            Assert.AreEqual(0, HttpMessageHandlerFactory.MockHandlersCount());
+        }
+
+        [TestMethod]
+        [Description("Instance discovery call is made because authority was not already in the instance cache")]
+        public async Task AuthorityNotInInstanceCache_InstanceDiscoverCallMadeTestAsync()
```
kpanwar: this test case should behave just like the one above where we get the tokens as expected. Only caveat is that InstanceDiscovery makes another call because DE cloud authority is not in the cache.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationResult.cs
```diff
@@ -124,7 +124,7 @@ internal AuthenticationResult(string accessTokenType, string accessToken, DateTi
         /// <summary>
         /// Gets the authority which has issued tokens.
```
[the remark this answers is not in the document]
jennyf19: done.

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
```diff
@@ -1,34 +1,68 @@
...
 
-namespace Test.ADAL.NET.Unit
+namespace Test.ADAL.NET.Integration
 {
     [TestClass]
     public class SovereignCloudsTests
     {
         private PlatformParameters platformParameters;
 
+        const string sovereignAuthorityHost = "login.some-sovereign-cloud";
```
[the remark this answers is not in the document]
jennyf19: done.

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
jennyf19: I removed extra code that wasn't needed

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
jennyf19: this mock is for the instance discovery call, but I'll remove this check here.

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
jennyf19: it's false because the instance discover call hasn't happened yet. Just verifying that the DE instance is not in the instance cache.

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
```diff
@@ -60,16 +96,78 @@ public void Initialize()
...
+            Assert.AreEqual(1, InstanceDiscovery.InstanceCache.Count());
+            Assert.AreEqual(true, InstanceDiscovery.InstanceCache.Keys.Contains("login.microsoftonline.com"));
+            // DE cloud authority not yet included in instance cache
+            Assert.AreEqual(false, InstanceDiscovery.InstanceCache.Keys.Contains("login.microsoftonline.de"));
+
+            // One mock remaining for instance discovery call
+            Assert.AreEqual(1, HttpMessageHandlerFactory.MockHandlersCount());
+            
+            // Instance discovery makes another call because DE cloud authority is not in the cache
+            var entry = await InstanceDiscovery.GetMetadataEntry(new Uri("https://login.microsoftonline.de"), true, new CallState(Guid.NewGuid())).ConfigureAwait(false);
```
kpanwar: why this your test case making this call? It should be ADAL code calling the InstanceDiscovery not the test itself. remove this GetMetadataEntry() call.

jennyf19: I fixed this

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
```diff
@@ -60,16 +97,90 @@ public void Initialize()
...
+
+            HttpMessageHandlerFactory.AddMockHandler(new MockHttpMessageHandler
+            {
+                Method = HttpMethod.Get,
+                ResponseMessage =
+                    MockHelpers.CreateSuccessTokenResponseMessage(TestConstants.DefaultUniqueId,
+                    TestConstants.DefaultDisplayableId, TestConstants.DefaultResource)
+            });
+
+            HttpMessageHandlerFactory.AddMockHandler(new MockHttpMessageHandler
```
kpanwar: why is this mock needed?

jennyf19: I removed it...needed to move a few things around.

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
```diff
@@ -1,75 +1,162 @@
...
 
             MockHelpers.ConfigureMockWebUI(
                 new AuthorizationResult(AuthorizationStatus.Success, authReturnedUriInputMock),
                 // validate that authorizationUri passed to WebUi contains instance_aware query parameter
-                new Dictionary<string, string> {{"instance_aware", "true"}});
+                new Dictionary<string, string> { { "instance_aware", "true" } });
+
+            HttpMessageHandlerFactory.AddMockHandler(MockHelpers.CreateInstanceDiscoveryMockHandler(TestConstants.GetDiscoveryEndpoint(TestConstants.DefaultAuthorityCommonTenant)));
 
             HttpMessageHandlerFactory.AddMockHandler(new MockHttpMessageHandler
```
kpanwar: why the url not being enforced here? Travis made the change already. it should force you to pass in url

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
```diff
@@ -1,75 +1,162 @@
...
+
+            MockHelpers.ConfigureMockWebUI(
+                new AuthorizationResult(AuthorizationStatus.Success, authReturnedUriInputMock),
+                // validate that authorizationUri passed to WebUi contains instance_aware query parameter
+                new Dictionary<string, string> { { "instance_aware", "true" } });
+
+            HttpMessageHandlerFactory.AddMockHandler(MockHelpers.CreateInstanceDiscoveryMockHandler(TestConstants.GetDiscoveryEndpoint
+                (TestConstants.DefaultAuthorityCommonTenant)));
+
+            HttpMessageHandlerFactory.AddMockHandler(new MockHttpMessageHandler
```
kpanwar: URL should be getting enforced here.

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
kpanwar: this test looks exactly like the one above. what is the goal of this test then? I think we have a misunderstanding here. can you please outline what this test is about and what are the network calls that will be made and why?

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
jennyf19: updated this.

tests/Test.ADAL.NET.Integration/SovereignCloudsTests.cs
jennyf19: done.

PR #889, opened 2017-11-07
Refactor pre-existing test cases to reduce dependency

rayluo:
While this PR is triggered by the modification for the default Instance Discovery mock, the actual adjustment to those affected test cases (some of them are 1 year old) is orthogonal to the Instance Discovery feature, and rightfully so. In fact, after those test cases refactor, they would pass no matter the default Instance Discovery mock contain Authority Migration metadata or not.

### DIFF
--- a/tests/Test.ADAL.NET.Integration/FederatedUserCredentialTests.cs
+++ b/tests/Test.ADAL.NET.Integration/FederatedUserCredentialTests.cs
@@ -54,6 +54,11 @@ namespace Test.ADAL.NET.Integration
         public void Initialize()
         {
             HttpMessageHandlerFactory.ClearMockHandlers();
+            ResetInstanceDiscovery();
+        }
+
+        public void ResetInstanceDiscovery()
+        {
             InstanceDiscovery.InstanceCache.Clear();
             HttpMessageHandlerFactory.AddMockHandler(MockHelpers.CreateInstanceDiscoveryMockHandler(TestConstants.GetDiscoveryEndpoint(TestConstants.DefaultAuthorityCommonTenant)));
         }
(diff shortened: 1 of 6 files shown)


kpanwar:
@rayluo - Can you clarify what you mean by, "In fact, after those test cases refactor, they would pass no matter the default Instance Discovery mock contain Authority Migration metadata or not."? test case behavior should change if the metadata response is changed. if the tests are not dependent on the behavior then they are not actually doing the right thing.

kpanwar:
@rayluo - It seems that you are inserting data in the cache by invoking instance discovery and when the actual test runs, there is no instance discovery executed. its quite smart way of circumventing the cache management issue, but kind of defeats the purpose. you could however, load the cache, clear instance discovery cache and add the mock response back in the queue. it saves you headache of managing cache entries like the way you want and we also get coverage on instance discovery execution.

tests/Test.ADAL.NET.Integration/FederatedUserCredentialTests.cs
```diff
@@ -320,7 +333,18 @@ public void Initialize()
             Assert.AreEqual(1, context.TokenCache.Count);
 
             // Cache entry updated with new access token
-            Assert.AreEqual("some-access-token", context.TokenCache.tokenCacheDictionary[key].Result.AccessToken);
+            var entry = await context.TokenCache.LoadFromCache(new CacheQueryData
```
kpanwar: instead of asking cache to load, you should explicitly examine the dictionary. that would circumvent any false positives that LoadFromCache() may throw

tests/Test.ADAL.NET.Integration/ManagedUsercredentialTests.cs
```diff
@@ -244,7 +256,17 @@ public void Initialize()
             Assert.IsNotNull(result.UserInfo);
 
             // Cache entry updated with new access token
-            Assert.AreEqual("some-access-token", context.TokenCache.tokenCacheDictionary[key].Result.AccessToken);
+            var entry = await context.TokenCache.LoadFromCache(new CacheQueryData
```
kpanwar: same as above

Issue #895, opened 2017-11-20
ADAL.Net does no longer build with VS 2017 (Version 15.4.4)


jmprieur:
**Repro steps**
- Install Visual Studio 2017.4.4. (all options to be sure that I'm not missing any frameworks)
- clone the ADAL.Net repository
```git clone https://github.com/AzureAD/azure-activedirectory-library-for-dotnet.git```
- open the ADAL.NET.NoWinRT.sln solution in Visual Studio.

**Expected:**
I can build the solution (possibly after updating the NuGet packages)

**Actual**
- after opening the solution for the first time, I get "Error occurred while restoring NuGet packages: The operation failed as details for project Microsoft.IdentityModel.Clients.ActiveDirectory could not be loaded"
- the Microsoft.IdentityModel.clients.ActiveDirectory project's Dependencies node shows icons warnings (as the NuGet packages were not restored)
![image](https://user-images.githubusercontent.com/13203188/33007971-55949c6e-cdd1-11e7-8efe-20280c079ded.png)

- The solution does not build

kpanwar:
@jmprieur what you see is just a VS thing. there are other issues that need to be addressed for xamarin forms etc which are causing the build failure.

kpanwar:
You can now use command line to restore and then use vs to build

PR #896, opened 2017-11-21
ADAL.Net does no longer build with VS 2017 (Version 15.4.4) 

kpanwar:
ADAL.Net does no longer build with VS 2017 (Version 15.4.4) , per #895.

This PR makes multiple fixes for the issue..
- Updates the incorrect test project structure that was still using VS2015 style csproj. We should NEVER use packages.config to add dependencies.
- Update Xamarin forms version for the sample app
- deletes .nuget folder
- ensures that tests projects do not depend on each other. they should only depend on common lib.


jennyf19:
🐈

jennyf19 (approved):
:shipit: 🐱

PR #897, opened 2017-11-21
Fix ArgumentException message

CyberNyanta:
> ...should be set to

merges with `computedRedirectUri`,  and it's quite confusing in log


CyberNyanta:
Thanks

PR #911, opened 2017-12-08
changing type of UserIdentifier.AnyUser from uniqueId to OptionalDisp…

SomkaPe:
UserIdentifier.AnyUser used to be initialized with type UniqueId which was an issue for IOS broker - it does expect object id value in username parameter for username_type uniqueId


src/Microsoft.IdentityModel.Clients.ActiveDirectory/UserIdentifier.cs
```diff
@@ -61,7 +61,7 @@ public sealed class UserIdentifier
         private const string AnyUserId = "AnyUser";
 
         private static readonly UserIdentifier AnyUserSingleton =
-            new UserIdentifier(AnyUserId, UserIdentifierType.UniqueId);
+            new UserIdentifier(AnyUserId, UserIdentifierType.OptionalDisplayableId);
```
kpanwar: this looks like a breaking change in behavior. can you confirm?

SomkaPe: as I see from the code - Nothing depends on AnyUser type, we only pass it to the broker,  Android brokers does not care about it, IOS broker  expect object id for uniqueId type

kpanwar:
@SomkaPe - please add unit tests as discussed.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/UserIdentifier.cs
HenrikMetzger: it's public, thus it could be used by external parties

---
In reply to: [156450874](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/911#discussion_r156450874) [](ancestors = 156450874)

HenrikMetzger:
>         }

Consider extracting to helpers (perhaps move to EncodingHelpers)

---
Refers to: src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs:175 in 6181110. [](commit_id = 6181110750889217ffb32d3dce9fbcaa298e4e12, deletion_comment = False)

HenrikMetzger:
+1 on the utest. pls. also consider other places where canonicalization should be used. Also the URI class does canonicalization automatically, thus it should be considered using that (though not sure if that is available on .net core).

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
```diff
@@ -164,7 +164,7 @@ internal static AuthorityType DetectAuthorityType(string authority)
             return authorityType;
         }
 
-        private static string CanonicalizeUri(string uri)
+        internal static string CanonicalizeUri(string uri)
```
HenrikMetzger: >internal static string CanonicalizeUri(string uri) [](start = 8, length = 50)

The URI class does canonicalization automatically, thus it should be considered using that (though not sure if that is available on .net core).

HenrikMetzger: Or perhaps the method really should be called EnsureUrlEndsWithForwardslash ...

---
In reply to: [157048639](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/911#discussion_r157048639) [](ancestors = 157048639)

HenrikMetzger:
> using System;

Pls. add header to the file.

---
Refers to: tests/Test.ADAL.NET.Unit/BrokerParametersTests.cs:1 in 19a4638. [](commit_id = 19a46388b3ae407344d364110e0ce82ec21daa9a, deletion_comment = False)

tests/Test.ADAL.NET.Unit/BrokerParametersTests.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.Clients.ActiveDirectory;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Cache;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.ClientCreds;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Flows;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Helpers;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Instance;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Test.ADAL.NET.Common;
```
HenrikMetzger: Any reason using statements are not inside the namespace? (it's a best practice to do that)

tests/Test.ADAL.NET.Unit/BrokerParametersTests.cs
```diff
@@ -0,0 +1,91 @@
...
+
+namespace Test.ADAL.NET.Unit
+{
+    [TestClass]
+    public class BrokerParametersTests
+    {
+        private const string Authority = "https://login.microsoftonline.com/test";
+        private static readonly string CanonicalizedAuthority = Authenticator.CanonicalizeUri(Authority);
```
HenrikMetzger: pls. see code comment from iteration 2

tests/Test.ADAL.NET.Unit/BrokerParametersTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            var acquireTokenInteractiveHandler = new AcquireTokenInteractiveHandler(_requestData,
+                TestConstants.DefaultRedirectUri, null, UserIdentifier.AnyUser,
+                ExtraQueryParameters, null, Claims);
+
+            Assert.AreEqual(11, acquireTokenInteractiveHandler.brokerParameters.Count);
+
+            var brokerParams = acquireTokenInteractiveHandler.brokerParameters;
+
+            Assert.AreEqual(CanonicalizedAuthority, brokerParams["authority"]);
```
HenrikMetzger: >"authority" [](start = 65, length = 11)

was it ever considered to add these names to a static (parameter) helper class?

SomkaPe: good point,  will add

tests/Test.ADAL.NET.Unit/BrokerParametersTests.cs
SomkaPe: first reason is to be consistent with all other classes in the library, second - Visual Studio and ReSharper put it outside by default.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
SomkaPe: it is not general canonicalization of uri, It is authority specific canonicalization, So meant to be used only for Authority uri

src/Microsoft.IdentityModel.Clients.ActiveDirectory/UserIdentifier.cs
SomkaPe: yes,  but it is not supposed to be used outside of Adal Api

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
henrik-me: that's why I suggested to rename the method

tests/Test.ADAL.NET.Unit/BrokerParametersTests.cs
henrik-me: I should have asked: What is the background for this decision? 

Aside: ReSharper defaults would likely want to put these inside (depending on version maybe).

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
SomkaPe: ok, will rename

tests/Test.ADAL.NET.Unit/BrokerParametersTests.cs
SomkaPe: We use default settings of ReSharper and it put it outside. Also most of classes in .NET Foundation (for example - https://github.com/dotnet/coreclr/tree/master/src/mscorlib/src/System) use outside option

henrik-me:
>         }

Separation of concern. IMO: if using this from two different places (which is done partially by making it internal) you should move this to the helper class. IMO: the Authenticator class should not be responsible for this.

---
In reply to: [351820359](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/911#issuecomment-351820359) [](ancestors = 351820359)

---
Refers to: src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs:175 in 6181110. [](commit_id = 6181110750889217ffb32d3dce9fbcaa298e4e12, deletion_comment = False)

Issue #923, opened 2017-12-18
Move persistance logic to common core


kpanwar:
The work includes 
- Refactor TokenCachePlugin classes into common core
- Move Logging infrastructure to core
- Replace CallState with RequestContext
- Move Crypto classes to core

jmprieur:
done

PR #924, opened 2017-12-20
Move persistance logic to common core

kpanwar:
Per (#923), the work includes
- Refactor TokenCachePlugin classes into common core
- Move Logging infrastructure to core
- Replace CallState with RequestContext
- Move Crypto classes to core


src/Microsoft.Identity.Core/Platforms/android/LegacyCachePersistance.cs
```diff
@@ -28,66 +28,54 @@
...
-                    byte[] state = Convert.FromBase64String(stateString);
-                    args.TokenCache.Deserialize(state);
+                    return Convert.FromBase64String(stateString);
                 }
             }
             catch (Exception ex)
             {
-                CallState.Default.Logger.Warning(null, "Failed to load cache: ");
-                CallState.Default.Logger.ErrorPii(null, ex);
+                CoreLoggerBase.Default.Warning("Failed to load cache: " + ex.Message);
```
jennyf19: I noticed several places where we don't double up the log messages for pii and no-pii. I thought we agreed to double up on messages so something that would be logged just to no-pii would also get logged in pii logs for log purposes on our end. Otherwise, if dev has piiEnabled, we will only get a small amount of log messages. I can make the changes once you merge. Also, in this case, why is one log warning and the other errorPii?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Platform/LoggerBase.cs
```diff
@@ -29,22 +29,29 @@
 using System.Globalization;
 using System.Runtime.CompilerServices;
 using System.Text;
+using Microsoft.Identity.Core;
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
 {
-    internal abstract class LoggerBase
+    internal abstract class LoggerBase : CoreLoggerBase
```
jennyf19: I like this implementation of LoggerBase

src/Microsoft.Identity.Core/Platforms/android/LegacyCachePersistance.cs
kpanwar: I kept the message level as is. I agree some logs updates are needed but they are orthogonal to the change here. Also, I think errorPii is used because it logs the exception object, but still inconsistent.

jennyf19: Ok....that's why I approved the PR because I knew these changes weren't necessary. Since your changes impacted all the logging functionality, it was easy to see.  I will update the logs.

ADAL.NET.VS2015.sln
```diff
@@ -13,6 +13,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AdalUniversalTestApp.Window
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ADAL.WinRT.VS2015.Platform", "src\Microsoft.IdentityModel.Clients.ActiveDirectory.Platform\ADAL.WinRT.VS2015.Platform.csproj", "{62659826-9AC1-4592-A1DB-1E7EFBDBD426}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Identity.Core.VS2015", "src\Microsoft.Identity.Core\Microsoft.Identity.Core.VS2015.csproj", "{72F03823-45DC-4F77-BDCB-29E880BF1AC8}"
```
jmprieur: we need to change the GUID for the project. It's already used for ADAL.WinRT.VS2015.Platform

src/Microsoft.Identity.Core/Platforms/android/LegacyCachePersistance.cs
jmprieur: @jennyf19 let's raise a bug for the missing log messages.

src/Microsoft.Identity.Core/CoreLoggerBase.cs
```diff
@@ -31,18 +31,21 @@
...
-        {
-            // Default implementation, do nothing
-        }
+        public static ILogger Default = null;
 
-        public static void AfterAccess(TokenCacheNotificationArgs args)
-        {
-            // Default implementation, do nothing
-        }
+        public abstract void Error(string message);
```
jmprieur: We might want to have verbs ? (WriteWarning? etc ...?)

src/Microsoft.Identity.Core/Platforms/android/LegacyCachePersistance.cs
```diff
@@ -28,66 +28,54 @@
...
                     editor.Apply();
-                    args.TokenCache.HasStateChanged = false;
                 }
                 catch (Exception ex)
-                {
-                    CallState.Default.Logger.Warning(null, "Failed to save cache: ");
-                    CallState.Default.Logger.ErrorPii(null, ex);
-                }
+            {
+                CoreLoggerBase.Default.Warning("Failed to save cache: " + ex.Message);
```
henrik-me: >"Failed to save cache: " [](start = 47, length = 24)

IMO: All strings should be in resource files mostly as it makes finding strings easier (no hard coding) and thus it will be possible to find by looking for references.

src/Microsoft.Identity.Core/CoreLoggerBase.cs
```diff
@@ -31,18 +31,21 @@
...
+        public abstract void ErrorPii(string message);
```
henrik-me: >Pii [](start = 34, length = 3)

Pii : newer terminology exists and Pii is not longer fully representing gdpr. EUII, Customer Data, etc. IMO, we should consider using a different terminology. I can only think of bad names right now.

src/Microsoft.Identity.Core/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,50 @@
...
+
+[assembly: AssemblyFileVersion("1.0.0.0")]
+
+// Setting ComVisible to false makes the types in this assembly not visible 
+// to COM components.  If you need to access a type in this assembly from 
+// COM, set the ComVisible attribute to true on that type.
+[assembly: ComVisible(false)]
+[assembly: CLSCompliant(true)]
+// The following GUID is for the ID of the typelib if this project is exposed to COM
+[assembly: Guid("ff47962a-d498-4c63-b7e9-4db3653ad7db")]
```
henrik-me: many of the assembly values can be shared across projects. e.g. company and copyright.

src/Microsoft.Identity.Core/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,50 @@
...
+[assembly: AssemblyProduct("Microsoft Identity Core")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCulture("")]
+[assembly: AssemblyCompany("Microsoft Corporation")]
+[assembly: AssemblyCopyright("Copyright (c) Microsoft Corporation. All rights reserved.")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyMetadata("Serviceable", "True")]
+
+[assembly: AssemblyFileVersion("1.0.0.0")]
```
henrik-me: Version number should be the same for all the assemblies right? And it should reflect the official version no. Is this available in another file in the repo?

henrik-me:
IMO: the strategy used for loggers seems to require a lot of duplicates. It seems like it should be possible to have just one logger which ensures the pii/euii/customerdata/etc. is passed when three is some and also that this is piped to the corresponding external logger as needed

Issue #929, opened 2017-12-29
Move AuthenticationResultEx to common core


kpanwar:
Move AuthenticationResultEx to common core. This is needed for cache code to work in core. The work includes
- Moving AuthenticationResultEx with rename.
- Create equivalent classes for AuthenicationResult and UserInfo because they are public and cannot be moved without issues of confusion and conflict with MSAL class with the same name.
- Update public AuthenticationResult class to return values from new equivalent class.
- Introcude CoreCacheItem class that contains both MSAL and ADAL cache objed representation. This is just an idea, but not finalized with approach.

henrik-me:
closing

PR #930, opened 2017-12-29
changes to move AuthenticationResultEx to core

kpanwar:
per #929, Move AuthenticationResultEx to common core. This is needed for cache code to work in core. The work includes

Moving AuthenticationResultEx with rename.
Create equivalent classes for AuthenicationResult and UserInfo because they are public and cannot be moved without issues of confusion and conflict with MSAL class with the same name.
Update public AuthenticationResult class to return values from new equivalent class.
Introcude CoreCacheItem class that contains both MSAL and ADAL cache objed representation. This is just an idea, but not finalized with approach.

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCacheItem.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCacheItem.cs
@@ -36,91 +36,79 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     /// </summary>
     public sealed class TokenCacheItem
     {
+        private readonly TokenCacheKey _key;
+        private readonly AdalResult _result;
+
         /// <summary>
         /// Default constructor.
         /// </summary>
         internal TokenCacheItem(TokenCacheKey key, AdalResult result)
         {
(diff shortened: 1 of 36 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCacheItem.cs
```diff
@@ -38,7 +39,7 @@ public sealed class TokenCacheItem
         /// <summary>
         /// Default constructor.
         /// </summary>
-        internal TokenCacheItem(TokenCacheKey key, AuthenticationResult result)
+        internal TokenCacheItem(TokenCacheKey key, AdalResult result)
```
jennyf19: Since you're passing in AdalResult, you can do the same thing you did with AuthenticationResult: ```_adalResult = adalResult ?? throw new ArgumentNullException();
            UserInfo = new UserInfo(_adalResult.UserInfo);```
And then set the properties like you did in AuthenticationResult ```public string AccessTokenType => _adalResult.AccessTokenType;```

jennyf19 (approved):
:shipit:

Issue #932, opened 2018-01-03
Dual Write RTs when persisting cache


kpanwar:
Dual Write RTs when persisting cache in iOS/Android/WinRT. RTs should be written in existing format and MSAL format. This will allow apps using MSAL and same client id and same publisher to get SSO.

PR #943, opened 2018-01-12
Dual Write RTs when persisting cache

kpanwar:
Dual Write RTs when persisting cache in iOS/Android/WinRT. RTs should be written in existing format and MSAL format. This will allow apps using MSAL and same client id and same publisher to get SSO. #932

### DIFF
--- a/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/StorageDelegates.cs
+++ b/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/StorageDelegates.cs
@@ -51,20 +51,18 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal
         public static void BeforeAccess(TokenCacheNotificationArgs args)
         {
 #if ANDROID || iOS || WINDOWS_APP
-            if (args != null && args.TokenCache != null && args.TokenCache.Count > 0)
+            if (args != null && args.TokenCache != null)
             {
-                // We assume that the cache has not changed since last write
-                return;
+                args.TokenCache.Deserialize(LegacyCachePersistance.LoadCache());
             }
 
-            args.TokenCache.Deserialize(LegacyCachePersistance.LoadCache());
 #endif
         }
 
         public static void AfterAccess(TokenCacheNotificationArgs args)
         {
 #if ANDROID || iOS || WINDOWS_APP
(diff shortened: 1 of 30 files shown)


core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -0,0 +1,181 @@
...
+                //property in AdalResultWrapper and its usage. Stronger design would be for the STS to return resource
+                //for which the token was issued as well on v2 endpoint.
+                ResourceInResponse = scope
+            };
+
+#if !FACADE && !NETSTANDARD1_3
+            IDictionary<AdalTokenCacheKey, AdalResultWrapper> dictionary = AdalCacheOperations.Deserialize(LegacyCachePersistance.LoadCache());
+            dictionary[key] = wrapper;
+            AdalCacheOperations.Serialize(dictionary);
+#endif
```
kpanwar: call legacypersistence.writeCache to write it back.

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -0,0 +1,181 @@
...
+                {
+                    listToProcess = clientInfoEntries;
+                }
+            }
+
+            //if upn is provided then use it to filter
+            if (!string.IsNullOrEmpty(upn))
+            {
+                //TODO - authority check needs to be updated for alias check
+                List<KeyValuePair<AdalTokenCacheKey, AdalResultWrapper>> upnEntries = listToProcess.Where(p => upn.Equals(p.Key.DisplayableId)).ToList();
```
kpanwar: move comment to line 108

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/StorageDelegates.cs
```diff
@@ -51,20 +51,18 @@ internal static class StorageDelegates
         public static void BeforeAccess(TokenCacheNotificationArgs args)
         {
 #if ANDROID || iOS || WINDOWS_APP
-            if (args != null && args.TokenCache != null && args.TokenCache.Count > 0)
+            if (args != null && args.TokenCache != null)
```
jennyf19: Thanks for remembering to fix this!

jennyf19 (approved):
:shipit: 😸

Issue #947, opened 2018-01-19
PiiLoggingEnabledFalse_TenantIdUserIdSetToNullValueTest Failure


jennyf19:
**Repro steps**
In Test Explorer, Run All tests for Combined.NoWinRT

**Actual**
In VSTS build - The test ```PiiLoggingEnabledFalse_TenantIdUserIdSetToNullValueTest``` fails. The Error Message is ```Assert.AreEqual failed. Expected:<(null)>. Actual:<A6xnQhbz4Vx2HuGl4lXwZ5U2I8iziLRFnhP5eNfIRvQ>```
In local build - The test ```PiiLoggingEnabledFalse_TenantIdUserIdSetToNullValueTest``` fails. The Error Message is ```System.NullReferenceException: Object reference not set to an instance of an object```

**Expected**
Since this test is part of testing Telemetry with PiiLoggingEnabled set to false, the TenantId & UserId should return as null values in the telemetry requests. Previously, they were returning null values. 

**Next Steps**
Figure out why this test is now failing. Might need to go through commit history to determine the problem. At which point did this test start to fail? Why? What is the fix?

jennyf19:
Due to refactoring for Core, we needed to move the TelemetryTests.cs file into Core where the other Telemetry functionality is located. All unit tests are now passing.

PR #948, opened 2018-01-19
Add ignore attribute to PiiLoggingEnabledFalse_TenantIdUserIdSetToNul…

jennyf19:
…lValueTest until we investigate the broken test case relating to issue #947

### DIFF
--- a/msal/tests/Test.MSAL.NET.Unit/TelemetryTests.cs
+++ b/msal/tests/Test.MSAL.NET.Unit/TelemetryTests.cs
@@ -313,6 +313,7 @@ namespace Test.MSAL.NET.Unit
 
         [TestMethod]
         [TestCategory("PiiLoggingEnabled set to false, TenantId & UserId set to null values")]
+        [Ignore] // Disable until we investigate and fix this broken test
         public void PiiLoggingEnabledFalse_TenantIdUserIdSetToNullValueTest()
         {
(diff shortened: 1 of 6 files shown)


msal/tests/Test.MSAL.NET.Unit/TelemetryTests.cs
```diff
@@ -313,6 +313,7 @@ public void PiiLoggingEnabledTrue_TenantAndUserIdHashedTest()
 
         [TestMethod]
         [TestCategory("PiiLoggingEnabled set to false, TenantId & UserId set to null values")]
+        [Ignore] // Disable until we investigate and fix this broken test
```
jmprieur: Could you please raise a GitHub issue to express that the text fails, and add the GitHub issue number in the Ignore attribute. This way this is traceable in both direction. When the test or tested product code is fixed, we'll remove the IgnoreAttribute
`` 
[TestMethod]
[Ignore(GitHubIssueNumber)]
public void SomeTestWhichGotIgnoredWithTraceabilityToGitHubIssueToInvestiigateAndFix()
{
``

jennyf19: We don't need this anymore. We needed to move the TelemetryTests to core, so this GitHub issue is no longer relevant

PR #951, opened 2018-01-22
Trwalke/x5c claim hotfix proposal

trwalke:
Proposed implementation of X5C public certificate claim.
This claim is not sent by default so the user needs to opt in to send it.
Changing public API with changes to AuthenticationContext.cs Ln:763

### DIFF
--- a/devApps/AdalCoreCLRTestApp/Program.cs
+++ b/devApps/AdalCoreCLRTestApp/Program.cs
@@ -55,7 +55,7 @@ namespace AdalCoreCLRTestApp
         {
             AuthenticationContext context = new AuthenticationContext("https://login.microsoftonline.com/common", true);
             var certificate = GetCertificateByThumbprint("<CERT_THUMBPRINT>");
-            var result = await context.AcquireTokenAsync("https://graph.windows.net", new ClientAssertionCertificate("<CLIENT_ID>", certificate));
+            var result = await context.AcquireTokenAsync("https://graph.windows.net", new ClientAssertionCertificate("<CLIENT_ID>", certificate), true);
 
             string token = result.AccessToken;
             Console.WriteLine(token + "\n");
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
@@ -765,13 +765,11 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
         /// </summary>
         /// <param name="resource">Identifier of the target resource that is the recipient of the requested token.</param>
         /// <param name="clientCertificate">The client certificate to use for token acquisition.</param>
-        /// <param name="sendX5c">Sends the x509 public certificate to the STS as a x5c claim.
-        /// This parameter enables application developers to opt-in to easy certificates roll-up
-        /// in Azure AD: in the case the certificate was renewed on the machine executing ADAL.NET,
-        /// setting this parameter to true will send the certificate public key to Azure AD in
-        /// addition to the certificate thumbprint, so that Azure AD rolls-over it's internal
-        /// knowledge of the certificate. This avoid the azure AD application admin to have to
-        /// change the application manifest in the Azure portal (or to run a powershell / CLI script)</param>
+        /// <param name="sendX5c">This parameter enables application developers to achieve easy certificates roll-up
+        /// in Azure AD: setting this parameter to true will send the certificate bits to Azure AD 
+        /// along with the token request, so that Azure AD can use it to update the application’s certificate registration.
+        /// This saves the application admin from the need to explicitly manage the certificate rollover
+        /// (either via portal or powershell/CLI operation)</param>
         /// <returns>It contains Access Token and the Access Token's expiration time. Refresh Token property will be null for this overload.</returns>
 #if ANDROID || iOS || WINDOWS_APP
         [Obsolete("As a security hygiene, this confidential flow API should not be used on this platform which only supports public client applications. For details please see https://aka.ms/AdalNetConfFlows")] 
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
@@ -761,11 +761,11 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
         }
 
         /// <summary>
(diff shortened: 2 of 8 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -760,6 +760,23 @@ internal IWebUI CreateWebAuthenticationDialog(IPlatformParameters parameters)
                 .ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Acquires security token from the authority and sends the x509 public certificate
+        /// </summary>
+        /// <param name="resource">Identifier of the target resource that is the recipient of the requested token.</param>
+        /// <param name="clientCertificate">The client certificate to use for token acquisition.</param>
+        /// <param name="sendX5c">Sends the x509 public certificate as a x5c claim</param>
```
jmprieur: sendX5c.
I would use as a comment: "
Sends the x509 public certificate to the STS as a x5c claim. This parameter enables application developers to opt-in to easy certificates roll-up in Azure AD: in the case the certificate was renewed on the machine executing ADAL.NET, setting this parameter to true will send the certificate public key to Azure AD in addition to the certificate thumbprint, so that Azure AD rolls-over it's internal knowledge of the certificate. This avoid the azure AD application admin to have to change the application manifest in the Azure portal (or to run a powershell / CLI script)
"

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -760,6 +760,23 @@ internal IWebUI CreateWebAuthenticationDialog(IPlatformParameters parameters)
                 .ConfigureAwait(false);
         }
 
+        /// <summary>
```
trwalke: @vibronet @jmprieur

devApps/AdalCoreCLRTestApp/Program.cs
```diff
@@ -55,7 +55,7 @@ static void Main(string[] args)
         {
             AuthenticationContext context = new AuthenticationContext("https://login.microsoftonline.com/common", true);
             var certificate = GetCertificateByThumbprint("<CERT_THUMBPRINT>");
-            var result = await context.AcquireTokenAsync("https://graph.windows.net", new ClientAssertionCertificate("<CLIENT_ID>", certificate));
+            var result = await context.AcquireTokenAsync("https://graph.windows.net", new ClientAssertionCertificate("<CLIENT_ID>", certificate), true);
```
henrik-me: >true [](start = 146, length = 4)

Thinking the app should include a scenario where this is false (e.g. make a request afterwards where this is false to ensure the cert is not in the request)?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
```diff
@@ -86,11 +90,15 @@ public JsonWebToken(IClientAssertionCertificate certificate, string audience)
...
 
-        public ClientAssertion Sign(IClientAssertionCertificate credential)
+        public ClientAssertion Sign(IClientAssertionCertificate credential, bool sendX5c)
         {
+            _sendX5c = sendX5c;
+            string token;
             // Base64Url encoded header and claims
-            string token = this.Encode(credential);     
-
+            lock (LockObject)
```
henrik-me: >ock (LockObject) [](start = 13, length = 16)

why is this needed?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
```diff
@@ -203,9 +210,36 @@ internal class JWTPayload
...
+                if (credential is ClientAssertionCertificate cert)
+                {
+                    X509CertificatePublicCertValue = Convert.ToBase64String(cert.Certificate.GetRawCertData());
+                }
+#elif NETSTANDARD1_3
+                if (credential is ClientAssertionCertificate cert)
+                {
+                    X509CertificatePublicCertValue = Convert.ToBase64String(cert.Certificate.RawData);
+                }
+#endif
```
henrik-me: can this difference be hidden in a helper/extension for each of the platforms?

henrik-me:
pls. ensure the changes also gets to the dev branch

devApps/AdalCoreCLRTestApp/Program.cs
trwalke: good point

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
trwalke: At the moment, I'm not sure. it may require a lot of changes.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
```diff
@@ -70,6 +71,8 @@ internal class JsonWebToken
 
         private readonly JWTPayload payload;
 
+        private bool _sendX5c = false;
```
henrik-me: not consistent with the rest of the files. pls. remove _ unless we decide that is out new norm

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
henrik-me: it will require a few changes.... an extension method in each of the libraries which extends the functionality of the Certificate type or you can do that as an extension to ether the type of the Certificate property  (or partial class) to the ClientAssertionCertificate. this is not a big deal, however I suggest creating a refactoring task for this one which we can include in dev, then we can work on it together

henrik-me:
looks ok. pls. ensure this is properly tested :)

jmprieur:
LGTM. Is it something we want to add in 3.17.4 ? in that case I guess this would be 3.18.0 as the public API changes?

henrik-me:
We would like to get this into the next release possible. @jmprieur if you are good with the public api change then we should be good to go.

jmprieur:
I'm good with it. @vibronet do you want to get a chance to apply a veto?

jennyf19:
Are we adding this to 3.18 release? @jmprieur @henrik-me

PR #953, opened 2018-01-23
WebBrowserNavigateErrorEventHandler namespace change breaks public API

### LINKED ISSUE #954: proposed commit for ADAL 3.17.4 has a public API breaking change (WebBrowserNavigateErrorEventHandler in different namespace)
WebBrowserNavigateErrorEventHandler used to be in Microsoft.IdentityModel.Clients.ActiveDirectory.Internal in ADAL.Net 3.17.3, and therefore changing the namespace of all this file to Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform constituted a public API breaking change (even if the delegate was not used by customers)

We need to restore the namespace for this type only (WebBrowserNavigateErrorEventHandler)

jmprieur:
WebBrowserNavigateErrorEventHandler  was in Microsoft.IdentityModel.Clients.ActiveDirectory.Internal in ADAL.Net 3.17.3, and therefore changing the namespace of all this file to Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform constituted a public API breaking change (even if the delegate was not used by customers)

Restoring the namespace for this type only (WebBrowserNavigateErrorEventHandler) to fix #954

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/CustomWebBrowser.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/CustomWebBrowser.cs
@@ -182,8 +182,8 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
                 if (msg.message != WM_CHAR && (ModifierKeys == Keys.Shift || ModifierKeys == Keys.Alt ||
                                                ModifierKeys == Keys.Control))
                 {
-                    int num = (int) msg.wParam | (int) ModifierKeys;
-                    Shortcut s = (Shortcut) num;
+                    int num = (int)msg.wParam | (int)ModifierKeys;
+                    Shortcut s = (Shortcut)num;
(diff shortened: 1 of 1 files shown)


jennyf19:
Will this need to go into dev as well? maybe in a separate PR?

jmprieur:
Yes, you are right @jennyf19, this also needs to go to dev. Can you please cherry-pick this change, @jennyf19  or @rayluo  ? 
cc: @henrik-me

rayluo:
@jmprieur @jennyf19 Just to clarify, since this change has been rightfully merged into release branch, there is no need for cherry-pick. We just need to merge the release branch back to the dev branch, which is one step among the normal release procedure. Besides, we don't really like 🍒 anyway.

We might need to have a separated conversation about how we deal with the current 2 different "dev" branches: the `dev` and the `adalv3/dev`, though.

cc: @henrik-me

jennyf19:
@rayluo Do you mean merging the release branch ```release-3.17.4``` into the ```adalv3/dev``` branch or the ```dev``` branch?

PR #959, opened 2018-01-25
HOTFIX: Fixed unit tests for X5c public cert on the request JWT.

trwalke:
Proposed implementation of X5C public certificate claim.
This claim is not sent by default so the user needs to opt in to send it.


src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -766,6 +766,28 @@ internal IWebUI CreateWebAuthenticationDialog(IPlatformParameters parameters)
                 .ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Acquires security token from the authority.
```
jmprieur: Acquires a security token from the authority while enabling simplifying Azure AD certificate roll-over?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -766,6 +766,28 @@ internal IWebUI CreateWebAuthenticationDialog(IPlatformParameters parameters)
...
+        /// </summary>
+        /// <param name="resource">Identifier of the target resource that is the recipient of the requested token.</param>
+        /// <param name="clientCertificate">The client certificate to use for token acquisition.</param>
+        /// <param name="sendX5c">This parameter enables application developers to achieve easy certificates roll-up
```
jmprieur: roll-up? or roll-over?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -766,6 +766,28 @@ internal IWebUI CreateWebAuthenticationDialog(IPlatformParameters parameters)
...
+        /// in Azure AD: setting this parameter to true will send the public certificate to Azure AD 
+        /// along with the token request, so that Azure AD can use it to validate the subject name based on a trusted issuer policy.
+        /// This saves the application admin from the need to explicitly manage the certificate rollover
+        /// (either via portal or powershell/CLI operation) 
+        /// IMPORTANT: this flow isn’t enabled on the service at the time of this SDK release (<ADAL version>).
```
jmprieur: ADAL.Net 3.19

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -766,6 +766,28 @@ internal IWebUI CreateWebAuthenticationDialog(IPlatformParameters parameters)
...
+        /// <returns>It contains Access Token and the Access Token's expiration time. Refresh Token property will be null for this overload.</returns>
+#if ANDROID || iOS || WINDOWS_APP
```
jmprieur: This this is a net new API, I don't think we want an Obsolete attribute here. Just include this API in the case of platforms supporting it.

jmprieur:
A few questions on the  comment for the API.
I don't think we should keep the obsolete attribute. this API is new, it cannot already be obsolete. we only want to include this new API in the case of platforms supporting it (or not including it in the case of platforms supporting it)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
henrik-me: roll-over

henrik-me:
pls. remember to push to dev as well

trwalke:
@henrik-me will do
@jmprieur All set with new comment?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -766,6 +766,25 @@ internal IWebUI CreateWebAuthenticationDialog(IPlatformParameters parameters)
...
+        /// <param name="resource">Identifier of the target resource that is the recipient of the requested token.</param>
+        /// <param name="clientCertificate">The client certificate to use for token acquisition.</param>
+        /// <param name="sendX5c">This parameter enables application developers to achieve easy certificates roll-over
+        /// in Azure AD: setting this parameter to true will send the public certificate to Azure AD 
+        /// along with the token request, so that Azure AD can use it to validate the subject name based on a trusted issuer policy.
+        /// This saves the application admin from the need to explicitly manage the certificate rollover
+        /// (either via portal or powershell/CLI operation) 
+        /// IMPORTANT: this flow isn’t enabled on the service at the time of this SDK release (ADAL.Net 3.19).
+        /// <returns>It contains Access Token and the Access Token's expiration time. Refresh Token property will be null for this overload.</returns>
+        public async Task<AuthenticationResult> AcquireTokenAsync(string resource,
```
jmprieur: I'm confused. No conditional code any longer?
I would have thought that we'd expose this API only for .NET Framework and .Net Core?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -767,20 +767,17 @@ internal IWebUI CreateWebAuthenticationDialog(IPlatformParameters parameters)
         }
 
         /// <summary>
-        /// Acquires security token from the authority.
+        /// Acquires a security token from the authority while enabling the simplification Azure AD certificate roll-over.
```
jmprieur: simplication? or simplified?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
jmprieur: I mean the method should be enclosed into a #if !ANDROID && !iOS && ! WINDOWS_APP block, or something like that?

trwalke:
@jennyf19 I could not get the adalv3/dev to build on my machine last night so I was not sure if it was working or not, so I made a hotfix branch off of master. I will merge that hotfix into master when it is ready and then move the same hotfix changes to the new adalv3/dev. I was under the impression that hotfixes were added by branching off of master and merging back in. We can discuss today at the office.

rayluo:
@trwalke @jennyf19

> 1. could not get the adalv3/dev to build

Then let's get this part fixed first. I believe @jennyf19 successfully build that literally yesterday. Might just be some "git clean -xdf" thing needed in @trwalke 's local environment. @jennyf19 is an expert for that. :-)

> 2. will merge that hotfix into master when it is ready and then move the same hotfix changes to the new adalv3/dev. I was under the impression that hotfixes were added by branching off of master and merging back in.

In the context of a hotfix (or a release) branch, please do merge it back to BOTH the master and the development branch that you started with. We will have a detail discussion for this off-line in an upcoming session.

On a side note, Travis you started this branch at a rare good timing that both `adalv3/dev` and `master` are essentially the same, so you could have started off from `adalv3/dev` and then do this x5c work in a normal feature branch. That may feel less intimidating. I can help you to prune (i.e. [rebase](https://git-scm.com/book/en/v2/Git-Branching-Rebasing)) your current branch to be on top of `adalv3/dev`, after you somehow fix the first issue above. Otherwise, if we continue with this hotfix approach, we will at least need to add one more commit to bump the version number.

rayluo:
Thanks @jennyf19 for changing this PR's base branch from `master` to `adalv3/dev`. That addresses the topic at the end of my previous comment. Good job!

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/ClientKey.cs
```diff
@@ -43,6 +43,7 @@ public ClientKey(string clientId)
 
             this.ClientId = clientId;
             this.HasCredential = false;
+            this.SendX5c = false;
```
brentschmaltz: Why don't aren't these constructors calling each other using : this(....)?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
```diff
@@ -73,6 +73,8 @@ internal class JsonWebToken
 
         private readonly JWTPayload payload;
 
+        private bool sendX5C = false;
```
brentschmaltz: Could you just pass sendX5c to Encode and do without the local variable?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
```diff
@@ -113,9 +116,9 @@ private static string UrlEncodeSegment(byte[] segment)
             return Base64UrlEncoder.Encode(segment);
         }
 
-        private static string EncodeHeaderToJson(IClientAssertionCertificate credential)
+        private static string EncodeHeaderToJson(IClientAssertionCertificate credential, bool sendX5C)
         {
-            JWTHeaderWithCertificate header = new JWTHeaderWithCertificate(credential);
+            JWTHeaderWithCertificate header = new JWTHeaderWithCertificate(credential, sendX5C);
```
brentschmaltz: You don't really need the local variable header, you could write:
return JsonHelper.EncodeToJson(new JWTHeaderWithCertificate(credential, sendX5C));

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
```diff
@@ -209,12 +212,15 @@ internal sealed class JWTHeaderWithCertificate : JWTHeader
...
             private string _thumbPrint;
 
-            public JWTHeaderWithCertificate(IClientAssertionCertificate credential)
+            public JWTHeaderWithCertificate(IClientAssertionCertificate credential, bool sendX5C)
                 : base(credential)
             {
                 _thumbPrint = this.Credential.Thumbprint;
                 X509CertificatePublicCertValue = null;
 
+                if (!sendX5C)
```
brentschmaltz: Do we need the instance variable _thumbPrint? Why not just use the Property - X509CertificateThumbprint?
Change it to a {get;set;}

tests/Test.ADAL.NET.Unit/ClientAssertionTestImplementation.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Test.ADAL.NET.Unit
+{
+    /// <summary>
+    /// Test implementation of IClientAssertionCertificate.
+    /// </summary>
+    [DeploymentItem("valid_cert.pfx")]
+    public class ClientAssertionTestImplementation : IClientAssertionCertificate
+    {
+
```
brentschmaltz: don't need this extra line

tests/Test.ADAL.NET.Unit/JsonWebTokenTests.cs
```diff
@@ -51,6 +52,54 @@ public class JsonWebTokenTests
...
+                Content = new StringContent("{\"token_type\":\"Bearer\",\"expires_in\":\"3599\",\"access_token\":\"some-access-token\"}")
+            },
+            AdditionalRequestValidation = request =>
+            {
+                var requestContent = request.Content.ReadAsStringAsync().GetAwaiter().GetResult();
+                var formsData = EncodingHelper.ParseKeyValueList(requestContent, '&', true, null);
+
+                // Check presence of client_assertion in request
+                string encodedJwt;
+                Assert.IsTrue(formsData.TryGetValue("client_assertion", out encodedJwt), "Missing client_assertion from request");
```
brentschmaltz: You can inline this as: (...., out string encodedJwt)

tests/Test.ADAL.NET.Unit/JsonWebTokenTests.cs
```diff
@@ -51,6 +52,54 @@ public class JsonWebTokenTests
...
+
+                // Check presence of x5c cert claim. It should exist.
+                JwtSecurityTokenHandler handler = new JwtSecurityTokenHandler();
```
brentschmaltz: use var?

tests/Test.ADAL.NET.Unit/JsonWebTokenTests.cs
```diff
@@ -51,6 +52,54 @@ public class JsonWebTokenTests
...
+                var requestContent = request.Content.ReadAsStringAsync().GetAwaiter().GetResult();
+                var formsData = EncodingHelper.ParseKeyValueList(requestContent, '&', true, null);
+
+                // Check presence of client_assertion in request
+                string encodedJwt;
+                Assert.IsTrue(formsData.TryGetValue("client_assertion", out encodedJwt), "Missing client_assertion from request");
+
+                // Check presence of x5c cert claim. It should not exist.
+                JwtSecurityTokenHandler handler = new JwtSecurityTokenHandler();
+                var jsonToken = handler.ReadToken(encodedJwt) as JwtSecurityToken;
```
brentschmaltz: You can use: JwtSecurityTokenHandler.ReadJwtToken(...) and you don't have to cast as it returns as JwtSecurityToken.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/ClientKey.cs
```diff
@@ -95,6 +114,7 @@ public ClientKey(ClientAssertion clientAssertion)
 
         public bool HasCredential { get; private set; }
 
+        public bool SendX5c { get; private set; }
```
henrik-me: >public bool SendX5c { get; private set; } [](start = 8, length = 41)

though instead of setting this in the constructors: public string SendX5c { get; set; } = false; ( this is btw. the default value of a bool in C#)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/ClientKey.cs
henrik-me: @travis.walker@microsoft.com you can simplify by making the default constructor private and setting the 2 values which are common for all constructors. on top the new overload which sets the sendX5c can as Brent mentions be called using this as well. You can also consider removing the one with-out the bool param forcing it to be specified? 

---
In reply to: [164022484](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/959#discussion_r164022484) [](ancestors = 164022484)

brentschmaltz:
:shipit:

PR #960, opened 2018-01-26
fixing issues related to incorrect parameters send in broker requests

SomkaPe:
- Sending correct user id to android broker for the case of silent flow + user id of Unique Type
- Fixing issue with always sending skip cache parameter to IOS Broker
- sending caller.info.package and caller.info.uid  to android Broker as hot fix to "recent" breaking change from they side
- use AssemblyFileVersion instead of AssemblyVersion as value for ClientVersion broker parameter

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
@@ -110,7 +110,8 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Flows
             brokerParameters[BrokerParameter.Resource] = requestData.Resource;
             brokerParameters[BrokerParameter.ClientId] = requestData.ClientKey.ClientId;
             brokerParameters[BrokerParameter.CorrelationId] = this.CallState.CorrelationId.ToString();
-            brokerParameters[BrokerParameter.ClientVersion] = AdalIdHelper.GetAdalVersion();
+            brokerParameters[BrokerParameter.ClientVersion] = AdalIdHelper.GetAssemblyFileVersion();
+
             this.ResultEx = null;
 
             CacheQueryData.ExtendedLifeTimeEnabled = requestData.ExtendedLifeTimeEnabled;
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/AuthenticationRequest.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/AuthenticationRequest.cs
@@ -25,6 +25,7 @@
 //
 //------------------------------------------------------------------------------
 
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Flows;
 using System;
 using System.Collections.Generic;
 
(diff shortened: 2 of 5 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
```diff
@@ -477,6 +477,9 @@ private Bundle GetBrokerOptions(AuthenticationRequest request)
             brokerOptions.PutString(BrokerConstants.AccountLoginHint, username);
             brokerOptions.PutString(BrokerConstants.AccountName, username);
 
+            brokerOptions.PutString(BrokerConstants.CallerInfoPackage, mContext.PackageName);
```
jmprieur: What will happen with non-recent version of the broker (before the Android breaking change in the broker). Will these parameters be just ignored?

henrik-me: good point. is it possible to test that out somehow?Does the broker have a version or similar where we can test for this? 

---
In reply to: [164066081](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/960#discussion_r164066081) [](ancestors = 164066081)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/AuthenticationRequest.cs
```diff
@@ -68,10 +69,17 @@ public AuthenticationRequest(IDictionary<string, string> brokerPayload)
...
             }
 
-            if (brokerPayload.ContainsKey("username"))
+            if (brokerPayload.ContainsKey(BrokerParameter.Username))
             {
-                LoginHint = brokerPayload["username"];
-                BrokerAccountName = LoginHint;
+                if (brokerPayload[BrokerParameter.UsernameType].Equals(UserIdentifierType.UniqueId.ToString()))
+                {
+                    UserId = brokerPayload["username"];
```
henrik-me: >"username" [](start = 43, length = 10)

BrokerParameter.Username ? 
(same below)

henrik-me:
>                 Claims = brokerPayload["claims"];

are these consts in BrokerConstants? BrokerParameters or?

---
Refers to: src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/AuthenticationRequest.cs:94 in 0616795. [](commit_id = 0616795363c46281555dfed6c8edb53c944ecd9c, deletion_comment = False)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -110,7 +110,8 @@ protected AcquireTokenHandlerBase(RequestData requestData)
             brokerParameters[BrokerParameter.Resource] = requestData.Resource;
             brokerParameters[BrokerParameter.ClientId] = requestData.ClientKey.ClientId;
             brokerParameters[BrokerParameter.CorrelationId] = this.CallState.CorrelationId.ToString();
-            brokerParameters[BrokerParameter.ClientVersion] = AdalIdHelper.GetAdalVersion();
+            brokerParameters[BrokerParameter.ClientVersion] = AdalIdHelper.GetAssemblyFileVersion();
```
henrik-me: >AdalIdHelper.GetAssemblyFileVersion(); [](start = 62, length = 38)

we should have a task to set this from the build system instead of us having to specify this manually... a task in the technical debt user story perhaps

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
SomkaPe: yes, it will be ignored, As i understand  that change was done 6 or 7 broker releases ago, not sure that I can download it from Google play,

PR #970, opened 2018-01-31
Jennyf/hotfix2

jennyf19:
Due to missing ConfigureAwait(false) in one method, deadlock was created when calling UI context without specifying ConfigureAwait(false) - basically passing the context


SomkaPe:
:shipit:

Issue #969, opened 2018-01-30
3.19.0 UWP Initial Authentication Request Returns Cancel


aherrick:
Updated from 3.17.3 to 3.19.0.

In 3.19.0 when I make an initial request to Authenticate and AquireToken, it instantly returns a "Cancelled" response. The next time the request happens however it works as expected.

Reverting back to 3.17.3 works as expected and allows for Authentication.

See example code:

```
            var authResultResp = new AuthenticationResultResp();
            try
            {
                var authContext = new AuthenticationContext(authority);
                var platformParams = new PlatformParameters(PromptBehavior.Auto, false);

                authResultResp.AuthenticationResult = await authContext.AcquireTokenAsync(resource, clientId, new Uri(returnUri), platformParams);
            }
            catch (AdalException ex)
            {
                authResultResp.ErrorMsg = ex.ErrorCode;
            }
            catch (Exception ex)
            {
                authResultResp.ErrorMsg = ex.Message;
            }
```

jmprieur:
@aherrick thanks for the feedback
Which platform are you using? .NET Framework? Xamarin ? can you please share some repro code?

aherrick:
This is Xamarin Forms 2.5.0.122203 UWP. 

The code is above calling into `AquireTokenAsync` It throws `AdalException `on initial attempt.

On second attempt it works fine. With 3.17.3 it works fine on first try.

SomkaPe:
@aherrick  could you please provide details about AdalException - exception message, error code...

aherrick:
Stack trace:

```
   at Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Flows.AcquireTokenInteractiveHandler.VerifyAuthorizationResult()
   at Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Flows.AcquireTokenInteractiveHandler.<PreTokenRequest>d__10.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
[...]

PR #982, opened 2018-02-07
Somkape/uwp web ui

SomkaPe:
Running web UI on UI thread for uwp applications

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/winrt/DispatcherTaskExtensions.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/winrt/DispatcherTaskExtensions.cs
@@ -1,0 +1,54 @@
+﻿//----------------------------------------------------------------------
+//
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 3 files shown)


jennyf19 (approved):
:shipit:

jennyf19:
@SomkaPe What about making this change for dev as well?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/winrt/DispatcherTaskExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+using System;
+using System.Threading.Tasks;
+using Windows.UI.Core;
+
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
+{
+    internal static class DispatcherTaskExtensions
+    {
+        public static async Task<T> RunTaskAsync<T>(this CoreDispatcher dispatcher,
+            Func<Task<T>> func, CoreDispatcherPriority priority = CoreDispatcherPriority.High)
```
henrik-me: >CoreDispatcherPriority.High [](start = 66, length = 27)

does this have to be High?

SomkaPe: no, it should be normal, good catch, will change

henrik-me:
+1: also, thinking it may be challenging to write a test, however I suppose now at least you can verify that there will not be a UI lock. It may not make a lot of sense to do this utest, but pls. consider a creative way to validate.

---
In reply to: [363636278](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/982#issuecomment-363636278) [](ancestors = 363636278)

henrik-me (approved):
:shipit:

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/winrt/DispatcherTaskExtensions.cs
henrik-me: great. approving. assuming it will be addressed before commit.

---
In reply to: [167149844](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/982#discussion_r167149844) [](ancestors = 167149844)

jmprieur (approved):

jennyf19:
@SomkaPe Remember to reference issue #969

PR #983, opened 2018-02-07
Fix VSTS build warning - cannot find "TestMexMalformed.xml" in federa…

jennyf19:
…ted tests..this is not needed. have removed

Fix VSTS build error AdalIdHelper.GetAdalVersion always returns 0.0.0.0, we need to use GetAssemblyFileVerison(), which should return the match.

### DIFF
--- a/tests/Test.ADAL.NET.Unit/BrokerParametersTests.cs
+++ b/tests/Test.ADAL.NET.Unit/BrokerParametersTests.cs
@@ -83,7 +83,7 @@ namespace Test.ADAL.NET.Unit
             Assert.AreEqual(ClientId, brokerParams[BrokerParameter.ClientId]);
 
             Assert.AreEqual(acquireTokenInteractiveHandler.CallState.CorrelationId.ToString(), brokerParams[BrokerParameter.CorrelationId]);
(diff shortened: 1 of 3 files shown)


jennyf19:
@SomkaPe - this should fix your build issues

tests/Test.ADAL.NET.Unit/BrokerParametersTests.cs
```diff
@@ -83,7 +83,7 @@ public void AcquireTokenInteractiveHandlerConstructor_InitializeBrokerParameters
             Assert.AreEqual(ClientId, brokerParams[BrokerParameter.ClientId]);
 
             Assert.AreEqual(acquireTokenInteractiveHandler.CallState.CorrelationId.ToString(), brokerParams[BrokerParameter.CorrelationId]);
-            Assert.AreEqual(AdalIdHelper.GetAdalVersion(), brokerParams[BrokerParameter.ClientVersion]);
+            Assert.AreEqual(AdalIdHelper.GetAssemblyFileVersion(), brokerParams[BrokerParameter.ClientVersion]);
```
jmprieur: Why do we need to change the tests?  Did the behavior of the code for the broker and we did not update the tests yet?

jennyf19: The behavior in broker changed and tests were not updated.

PR #985, opened 2018-02-10
Fix dev

jmprieur:


build/credscan-exclusion.json
```diff
@@ -0,0 +1,33 @@
+{
```
jmprieur: Merging the ADAL.NET and MSAL.NET credscan's exclusion file for the common build.

Issue #989, opened 2018-02-13
Improved description of ADAL.NET nuget package enables better discovery


jmprieur:
From Visual Studio's NuGet package manager or Nuget.org, try to find ADAL.NET by searching for words like:
- "access token"
- "access a protected api"

**Expected:**
you find ADAL.NET (and MSAL.NET)

**Actual:**
you cannot find them

PR #990, opened 2018-02-13
Jmprieur/fix989

jmprieur:
This rewords the description of the ADAL.NET NuGet package so that searching for "secure API", or "token for protected API", or "access token" finds ADAL.Net
See #989


jmprieur:
We also need to apply the same to dev.

PR #1011, opened 2018-03-08
Trwalke/treat warnings as errors

trwalke:
Associated user story: https://identitydivision.visualstudio.com/DevEx/_workitems/edit/364068


trwalke:
Yes, consistent for all platforms.

PR #1017, opened 2018-03-19
Sagonzal/correlation

sangonzal:
Device code flow makes two API calls, which were previously using different `CorrelationIds`. They are "one business transaction" and should therefore use the same Id.

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/DeviceCodeResult.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/DeviceCodeResult.cs
@@ -77,5 +77,10 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
         /// Identifier of the target resource that would be the recipient of the token.
         /// </summary>
         public string Resource { get; internal set; }
+
+        /// <summary>
+        /// Identifier of the correlation id. Used in diagnostic purposes.
+        /// </summary>
+        public Guid CorrelationId { get; internal set; }
     }
 }
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
@@ -54,7 +54,9 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Flows
         protected AcquireTokenHandlerBase(RequestData requestData)
         {
             this.Authenticator = requestData.Authenticator;
-            this.CallState = CreateCallState(this.Authenticator.CorrelationId);
+            this.CallState = CreateCallState(requestData.CorrelationId != Guid.Empty
+                ? requestData.CorrelationId 
+                : this.Authenticator.CorrelationId);
             brokerHelper.CallState = this.CallState;
 
             var msg = string.Format(CultureInfo.CurrentCulture,
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
@@ -199,7 +199,8 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
                 TokenCache = this.TokenCache,
                 ExtendedLifeTimeEnabled = this.ExtendedLifeTimeEnabled,
                 Resource = deviceCodeResult.Resource,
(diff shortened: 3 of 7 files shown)


henrik-me:
@jmprieur any concerns about the addition to the public API?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/DeviceCodeResult.cs
```diff
@@ -77,5 +77,10 @@ internal DeviceCodeResult()
         /// Identifier of the target resource that would be the recipient of the token.
         /// </summary>
         public string Resource { get; internal set; }
+
+        /// <summary>
+        /// Identifier of the correlation id. Used in diagnostic purposes.
+        /// </summary>
+        public Guid CorrelationId { get; internal set; }
```
henrik-me: >public Guid CorrelationId { get; internal set; } [](start = 8, length = 48)

I'm not sure we should expose this but not sure how otherwise to keep the code clean... perhaps this is fine. Jean-Marc?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -54,7 +54,9 @@ internal abstract class AcquireTokenHandlerBase
         protected AcquireTokenHandlerBase(RequestData requestData)
         {
             this.Authenticator = requestData.Authenticator;
-            this.CallState = CreateCallState(this.Authenticator.CorrelationId);
+            this.CallState = CreateCallState(requestData.CorrelationId != Guid.Empty
+                ? requestData.CorrelationId 
+                : this.Authenticator.CorrelationId);
```
henrik-me: >this.Authenticator.CorrelationId [](start = 18, length = 32)

Will this work well for Authenticator scenarios? Ideally if the Correlation id is always set to Guild.Empty in all other cases but when we have the decice code flow?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/DeviceCodeResult.cs
jmprieur: what is the concern, @henrik-me ? in term of code not being clean?

jmprieur: I replied to quickly. No, @henrik-me : additive is fine provided we respect binary compatibility as well (in which case we are adding an extra compiler generated field. provided it's last that's fine)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
sangonzal: Correct, `requestData.CorrelationId` is only set in device code flow, and will be Guid.Empty for all other flows and so therefore the `CallState` will default to using `Authenticator.CorrelationId`.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/DeviceCodeResult.cs
henrik-me: Q was also ensure you are fine adding CorrelationId here. Anyway I do not have any real concern.

---
In reply to: [175547217](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1017#discussion_r175547217) [](ancestors = 175547217)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
henrik-me: Thanks for explaining. Great for backwards compatibility/consistency. 
The I suppose my only potential concern has nothing to do with your change it is really about the state we set. Do we validate the state is the same as we send?

---
In reply to: [175547615](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1017#discussion_r175547615) [](ancestors = 175547615)

henrik-me:
Please add utests (L0 effort)

henrik-me:
Consider also adding a test which ensures that for other flows it is not set

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
sangonzal: It seems that the CallState is not currently validated to make sure it is the same as the one set for any of the flows.

PR #1024, opened 2018-03-31
Addressing xml build warnings...this is internal and not public facing

jennyf19:


jennyf19:
I needed a test branch for VSTS and thought i'd take advantage and remove some warnings from core

jennyf19:
@jmprieur - I added in an XML comment for the AuthenticationResult internal method.  Thanks for the feedback.

PR #1038, opened 2018-04-16
Jennyf/core web view

jennyf19:
Initial PR for:
> Refactoring core 
> Enabling system webview support in core for msal
> Fix mock and unit tests references due to core refactoring

I did manual testing w/:
* Xamarin Android in Adal (works as expected w/embedded webview)
* Xamarin Android in MSAL (works as expected w/system browser)


core/src/Platforms/Android/SystemWebview/AuthenticationActivity.cs
```diff
@@ -61,7 +60,7 @@ public AuthenticationActivity()
         private string _requestUrl;
         private int _requestId;
         private bool _restarted;
-       // private CustomTabsActivityManager _customTabsActivityManager;
+        // private CustomTabsActivityManager _customTabsActivityManager;
```
jmprieur: you might want to remove this comment since it has been here already?

jennyf19: done...thanks!

henrik-me:
Thanks for adding the manual test info

core/src/Microsoft.Identity.Core.csproj
```diff
@@ -125,7 +125,17 @@
     <Reference Include="System.Xml" />
     <Compile Include="Platforms\Android\**\*.cs" />
   </ItemGroup>
-  
+  <ItemGroup Condition="'$(TargetFramework)' == 'MonoAndroid7'">
+    <PackageReference Include="Xamarin.Android.Support.CustomTabs">
+      <Version>25.4.0.2</Version>
+    </PackageReference>
+    <PackageReference Include="Xamarin.Android.Support.v7.AppCompat">
+      <Version>25.4.0.2</Version>
```
henrik-me: >25.4.0.2 [](start = 15, length = 8)

is this a consistent version used everywhere? Perhaps we should have a component with this version we include to ensure it gets easily set everywhere?

core/src/Properties/InternalsVisibleTo.cs
```diff
@@ -36,6 +36,7 @@
...
 [assembly: InternalsVisibleTo("AutomationApp, PublicKey=00240000048000009400000006020000002400005253413100040000010001002D96616729B54F6D013D71559A017F50AA4861487226C523959D1579B93F3FDF71C08B980FD3130062B03D3DE115C4B84E7AC46AEF5E192A40E7457D5F3A08F66CEAB71143807F2C3CB0DA5E23B38F0559769978406F6E5D30CEADD7985FC73A5A609A8B74A1DF0A29399074A003A226C943D480FEC96DBEC7106A87896539AD")]
 [assembly: InternalsVisibleTo("DesktopTestApp, PublicKey=00240000048000009400000006020000002400005253413100040000010001002D96616729B54F6D013D71559A017F50AA4861487226C523959D1579B93F3FDF71C08B980FD3130062B03D3DE115C4B84E7AC46AEF5E192A40E7457D5F3A08F66CEAB71143807F2C3CB0DA5E23B38F0559769978406F6E5D30CEADD7985FC73A5A609A8B74A1DF0A29399074A003A226C943D480FEC96DBEC7106A87896539AD")]
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
```
henrik-me: >DynamicProxyGenAssembly2 [](start = 31, length = 24)

interesting name... is this project coming in the next update to webview?

msal/src/Microsoft.Identity.Client/UIParent.cs
```diff
@@ -38,37 +39,35 @@ namespace Microsoft.Identity.Client
     /// </summary>
     public sealed class UIParent
     {
+        internal CoreUIParent CoreUIParent { get; private set; }
```
henrik-me: >internal CoreUIParent CoreUIParent { get; private set; } [](start = 8, length = 56)

not sure I understand why this is a component of the UIParent vs. a base class. The name indicates otherwise to me. Can you pls. educate me on the background for this choice?

core/src/Microsoft.Identity.Core.csproj
jennyf19: Yes it is. Do you want that included with this PR?

core/src/Properties/InternalsVisibleTo.cs
jennyf19: DynamicProxyGenAssembly2 is a temporary assembly build by mocking systems that use NSubsitute. It's created when the mock is needed and disposed of when the tests finish. Based on an error message, I needed to include this for the Test.MSAL.NET.Unit mocks.

msal/src/Microsoft.Identity.Client/UIParent.cs
jennyf19: I needed a class to hold the properties from ADAL and MSAL. I will need to move attributes of platform parameters from ADAL into CoreUIParent, so CoreUIParent will contain attributes of both (platform parameters and UIParent properties). 
If the word "Core" is weird, I can change it to something else.

core/src/Microsoft.Identity.Core.csproj
henrik-me: Can be done in the next PR. Pls. ensure you create a task for this to ensure it is not forgotten.

---
In reply to: [181953341](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1038#discussion_r181953341) [](ancestors = 181953341)

msal/src/Microsoft.Identity.Client/Internal/PlatformPlugin.cs
```diff
@@ -59,8 +60,13 @@ public static void InitializeByAssemblyDynamicLinking()
         {
 #if !FACADE
             CoreLoggerBase.Default = new MsalLogger(Guid.Empty, null);
-            InjectDependecies(
-                (IWebUIFactory) new UI.WebUIFactory(),
+            IWebUIFactory obj = null;
+#if ANDROID
+obj = new Microsoft.Identity.Core.UI.WebUIFactory();
```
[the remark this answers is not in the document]
jennyf19: yes...thanks! Fixed.

msal/src/Microsoft.Identity.Client/UIParent.cs
henrik-me: I'm thinking the xml comments of the class needs to explain the intention of the class. that will help me understand if it is more appropriate to name any of the classes differently. I'm not 100% sure what the intention of each of these are. Is the intention of CoreUIParent only to be a container for data/settings (properties)? Is there a reason not to put everything into Core? ... 

Aside: I think we talked about not using Core in our naming? Core and Base are generally names used to refer to certain inheritance hierarchies - which is probably what confused me the most in this case.

---
In reply to: [182136079](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1038#discussion_r182136079) [](ancestors = 182136079)

core/src/Properties/InternalsVisibleTo.cs
henrik-me: Where is NSubstitute used?

---
In reply to: [182133896](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1038#discussion_r182133896) [](ancestors = 182133896)

core/src/Microsoft.Identity.Core.csproj
henrik-me: Can be done in the next PR. Pls. ensure you create a task for this to ensure it is not forgotten.

---
In reply to: [182131252](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1038#discussion_r182131252) [](ancestors = 182131252)

msal/src/Microsoft.Identity.Client/UIParent.cs
henrik-me: I suppose I would consider the CoreUIParent the data/context object and the UIParent a factory class. Then the CoreUIParent (which could then be called UIParent) and the current UIParent could be called UIParentFactory ...

Any reason this is not all in the Core logic?

---
In reply to: [182141294](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1038#discussion_r182141294) [](ancestors = 182141294,182139799)

core/src/UI/CoreUIParent.cs
```diff
@@ -0,0 +1,84 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+#if ANDROID
+using System;
+using Android.App;
+#endif
```
henrik-me: minor: generally I suggest putting using statements inside the namespace.

core/src/Properties/InternalsVisibleTo.cs
jennyf19: We use it as a mocking framework in msal and adal.

msal/src/Microsoft.Identity.Client/UIParent.cs
jennyf19: It's not in core logic because UIParent is public and needs to remain in MSAL, and all the core classes are internal. CoreUIParent will hopefully be a single object representing properties of platform parameters and UIParent.

core/src/UI/CoreUIParent.cs
jennyf19: That wouldn't be consistent w/the other classes with have currently...why do you suggest putting them inside?

henrik-me: mostly to satisfy StyleCop rules... for most practical reasons it makes very little difference. Thus mostly a style reason (and most teams I worked on used that style). There is no real need to change these now - I just god used to seeing this the other way around... (and since the tools when enabled will force this behavior)

---
In reply to: [182148298](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1038#discussion_r182148298) [](ancestors = 182148298)

jennyf19: interesting. I read some documentation on inside vs outside...seems like inside is slightly better, but I think it's important to have consistency at this point.

Issue #1039, opened 2018-04-17
3.19.2: Trace events from Microsoft.IdentityModel.Clients.ActiveDirectory.AdalOption.AdalEventSource seem to be coming as Errors when they should be Information or Verbose


fpintos:
**Issue**
Events from AdalOption.AdalEventSource are reaching the event listener with name & level of "Error", when they should in fact be informational events.

This is likely due to [DefaultLog](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/130219b656485b4f536be62af1a18c4e123fe09d/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Features/WinCommon/AdalLogger.cs) not taking the level into account and just calling the event source Error().

**Expected**
Events to come with correct level and be filtered as expected by the event listener.

**Details/Repro**
I have an event listener hooked up to AdalOption.AdalEventSource that looks like this:
```csharp
    public class TracingEventListener : EventListener
    {
        protected override void OnEventWritten(EventWrittenEventArgs eventData)
        {
            var message = GetMessage(eventData);
            Trace.WriteLine($"level={eventData.Level},{message}");
        }

        protected static string GetMessage(EventWrittenEventArgs eventData)
        {
            if (eventData.Message != null)
            {
                return string.Format(eventData.Message, eventData.Payload.ToArray());
            }

            var sb = new StringBuilder($"eventId={eventData.EventId},eventName={eventData.EventName},");
            sb.Append('(');
            for (var i = 0; i < eventData.PayloadNames.Count; i++)
            {
                sb.Append(eventData.PayloadNames[i]);
                sb.Append('=');
                sb.Append((eventData.Payload[i] ?? "<null>"));
                if (i != eventData.PayloadNames.Count - 1)
                {
                    sb.Append(',');
                }
            }

            sb.Append(')');
            return sb.ToString();
        }
    }
```

And it is initialized with:
```csharp
var eventListener = new TracingEventListener();
eventListener.EnableEvents(AdalOption.AdalEventSource, EventLevel.Error);
```

When I run my application I get lots of events like this:
```
[...]

PR #1041, opened 2018-04-18
Filter logs for AdalEventSource based on log level in default log

jennyf19:
#1039
Customer did a check w/nuget package and AdalEventSource logging now filters based on log level in default log


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Features/WinCommon/Logger.cs
```diff
@@ -38,7 +38,22 @@ static Logger()
 
         internal override void DefaultLog(LogLevel logLevel, string message)
         {
-            AdalEventSource.Error(message);
+            if (logLevel == LogLevel.Information)
+            {
+                AdalEventSource.Information(message);
+            }
+            else if (logLevel == LogLevel.Verbose)
```
jmprieur: I might be a bit picky but I would have ordered them by severity? (Verbose first, and then information, warning, and error) ?

jennyf19: yes...great point - thank you

PR #1042, opened 2018-04-18
Filter logs for AdalEventSource based on log level in default log

jennyf19:
#1039 
Customer did a check w/nuget package and AdalEventSource logging now filters based on log level in default log


fpintos:
Thanks Jenny!

jennyf19:
@fpintos We'll have a release out later today w/the fix. 3.19.4

Issue #1016, opened 2018-03-19
iOS AuthenticationAgentUIViewController not dismissed when authentication task finishes


smstuebe:
Is there a reason, why `callbackMethod` is called before `DismissViewController` in `AuthenticationAgentUIViewController`? e.g.

https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/ec4c95bd7aa0f7ab5d239976943b9776eb5b0fa5/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationAgentUIViewController.cs#L155-L156

The Problem  I have is, that I have code like that (which is pretty common in Xamarin Apps) :

```csharp
try
{
    var auth = await _authenticationContext.AcquireTokenAsync(...);
    // do stuff
}
catch (Exception ex)
{
    await _userDialogs.AlertAsync("### error");
}
```

`_userDialogs` is an instance of [Alan Ritchie's UserDialogs ](https://github.com/aritchie/userdialogs) which uses `UIApplication.SharedApplication.GetTopViewController()` to get the top ViewController. And so it happens that **after** `AcquireTokenAsync` returns or throws an exception, the navigation stack is in an undefined state and this results in Errors like these: 

>  Warning: Attempt to present <UIAlertController: 0x104602800> on <AuthenticationAgentUINavigationController: 0x10be47000> whose view is not in the window hierarchy!

In my opinion it should clean up everything, dismiss the VC completely and than return. What do you think?

jennyf19:
@smstuebe Thanks for bringing this to our attention. We'll take a look and get back to you.

henrik-me:
@smstuebe . My general thought on this is that we would not do any cleanup before we are done with all the work we need to do which includes doing the callback. 

It will take time for us to fully validate if this will have any impact on other customer scenarios and we have no current plans on changing this behavior. 

Did you try changing this value to validate if it actually works for you to change the order?

Does the library you mention have overrides or workarounds for this kind of behavior. I'm thinking there might be other controls which behaves similarly. E.g. Can you manually pop the window off the navigation stack?

Also will it be possible for you to be more specific in your exception handling and thus be able to handle this scenario?

jennyf19:
@smstuebe - I did some testing w/Xamarin iOS and there is no reason to not shut down the UI as soon as it is done. I will make the change you suggested.

henrik-me:
[...]

PR #1044, opened 2018-04-19
Changing order of callback and dismissing the view controller

jennyf19:
Addressing issue #1016 
Customer reports issues with navigation stack when authentication is cancelled and we perform the callback first and then `DismissViewController`. It makes sense to shut down the UI first, especially for customers using more complex view controllers

Manually tested:
> Xamarin iOS test app

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationAgentUIViewController.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationAgentUIViewController.cs
@@ -93,8 +93,8 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
                 if (requestUrlString.StartsWith(callback, StringComparison.OrdinalIgnoreCase) || 
                     requestUrlString.StartsWith(BrokerConstants.BrowserExtInstallPrefix, StringComparison.OrdinalIgnoreCase))
                 {
-                    callbackMethod(new AuthorizationResult(AuthorizationStatus.Success, request.Url.ToString()));
-                    this.DismissViewController(true, null);
+                    this.DismissViewController(true, () =>
+                        callbackMethod(new AuthorizationResult(AuthorizationStatus.Success, request.Url.ToString())));
                     return false;
                 }
 
@@ -123,8 +123,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
                     AuthorizationResult result = new AuthorizationResult(AuthorizationStatus.ErrorHttp);
                     result.Error = AdalError.NonHttpsRedirectNotSupported;
                     result.ErrorDescription = AdalErrorMessage.NonHttpsRedirectNotSupported;
-                    callbackMethod(result);
-                    this.DismissViewController(true, null);
+                    this.DismissViewController(true, () => callbackMethod(result));
                     return false;
                 }
 
@@ -150,8 +149,8 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
 
         private void CancelAuthentication(object sender, EventArgs e)
         {
-            this.DismissViewController(true, null);
-            callbackMethod(new AuthorizationResult(AuthorizationStatus.UserCancel, null));
+            this.DismissViewController(true, () =>
+                callbackMethod(new AuthorizationResult(AuthorizationStatus.UserCancel, null)));
         }
 
         public override void DismissViewController(bool animated, Action completionHandler)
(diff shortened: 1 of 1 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationAgentUIViewController.cs
```diff
@@ -152,8 +150,8 @@ public override void ViewDidLoad()
 
         private void CancelAuthentication(object sender, EventArgs e)
         {
-            callbackMethod(new AuthorizationResult(AuthorizationStatus.UserCancel, null));
             this.DismissViewController(true, null);
+            callbackMethod(new AuthorizationResult(AuthorizationStatus.UserCancel, null));
```
henrik-me: isn't the override on the DismissViewController a callback to a method which will be ensure to be completed after dismissal of the viewcontroller? (see below) - I think the right thing to do is to :

this.DismissViewController(true, new AuthorizationResult( ...)

(wrapped into an action perhaps via the callbackMethod)

henrik-me (changes requested):
:clock1:

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationAgentUIViewController.cs
jennyf19: I have a few questions about this for you...I'll swing by tomorrow. 🐈

gameleon-dev:
The `DismissViewController` method is not synchronous. Which means there is no guarantee the view controller has been properly removed from the hierarchy even if you switch the method calls in the code.

The proper way to handle this would be to either use the completion handler:

```
DismissViewController(true, () =>
{
    callbackMethod(new AuthorizationResult(AuthorizationStatus.UserCancel, null));
});
```

Or use the Async version of the dismiss method:

```
await this.DismissViewControllerAsync(true);
callbackMethod(new AuthorizationResult(AuthorizationStatus.UserCancel, null));
```

Both these methods will ensure that the viewcontroller has been removed from the view hierachy.

gameleon-dev:
Looking at the controller it seems there are a two more snippets where this might be an issue:

https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/82565905ca0213e7963a275c40382a8560f9d141/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationAgentUIViewController.cs#L95-L101

and 

https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/62a82c78b2a8504065f342309f8015dde8b9eccc/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationAgentUIViewController.cs#L128-L129

henrik-me:
Thanks for sharing the info @gameleon-dev. We have had a few internal discussions on this. We'll take a look at it. @jennyf19 @jmprieur

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationAgentUIViewController.cs
gameleon-dev: I would propose to change this line to:

`this.DismissViewController(true, ()=> callbackMethod(new AuthorizationResult(AuthorizationStatus.UserCancel, null)));`

Moving the callback method to the completionHandler would ensure it will not fire until the controller is fully dismissed and removed from the view hierarchy.

The same would apply to the following code snippets:

https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/40e622f5aa279bf7c5852cdca8bd572580039afa/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationAgentUIViewController.cs#L96-L97

https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/40e622f5aa279bf7c5852cdca8bd572580039afa/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationAgentUIViewController.cs#L126-L127

These would have to be changed to 

`
                    this.DismissViewController(true, ()=>
                        callbackMethod(new AuthorizationResult(AuthorizationStatus.Success, request.Url.ToString())));`

and 

`
                    this.DismissViewController(true, ()=> callbackMethod(result));`

respectively.

gameleon-dev:
Apologies for the delay on this one. I have made the proposed changes in #1095

jennyf19:
@henrik-me - I pulled in @gameleon-dev 's changes and updated the branch. Manually tested w/iOS simulator.

---
In reply to: [383859682](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1044#issuecomment-383859682) [](ancestors = 383859682)

henrik-me:
Thanks, @jennyf19 if I recall correctly you made some similar changes to the dev branch and thus we should be good to go. Approving.

henrik-me (approved):

PR #1050, opened 2018-05-11
Henrikm/buildandutestupdates

henrik-me:
minor doc bug fix and adding a few simple scripts to build and restore. also added the beginning of a pre-req validation script. Made an update to the target android version... I believe it's ok, but it may have side effects and should potentially be 26 and not 27 (as that is targeting v.8 which is also updated in this change). Made a few updates to address some of the issues reported by the build of deprecated Xamarin Android forms PageRenderer.


verify.cmd
```diff
@@ -0,0 +1,49 @@
...
+echo Available SDK's:
+echo %ProgramFiles(x86)%\Android\android-sdk\platforms
+echo .
+echo Open SDK Manager:
+echo  %ProgramFiles(x86)%\Android\android-sdk\SdkManager.exe
+echo .
+echo Validating available SDK's ...
+if not exist android-25 echo Please install Android SDK v25 (Open SDK Manager from above location)
+if not exist android-26 echo Please install Android SDK v26 (Open SDK Manager from above location)
+if not exist android-27 echo Please install Android SDK v27 (Open SDK Manager from above location)
```
jennyf19: cool...can we install automatically?

strongname.cmd
```diff
@@ -0,0 +1,21 @@
...
+set bconfig=debug
+if '%1' NEQ '' (set bconfig=%1%)
+
+set gotoFolder=adal\src\Microsoft.IdentityModel.Clients.ActiveDirectory\bin\%bconfig%\net45
+echo config: %gotoFolder%
+pushd %gotoFolder%
+
+sn.exe -Vr Microsoft.IdentityModel.Clients.ActiveDirectory.dll
+sn.exe -Vr Microsoft.Identity.Core.dll
+
```
jennyf19: why are you disabling strongname for just this dll?

jennyf19 (changes requested):
Looks good...builds much faster now.
I pulled your branch and did a build w/Core..it was fine, but I did have errors when building ADAL alone w/msbuild (restore/build), but they are not related to the PR (i think)...no issues w/MSAL alone.
I also deployed the AdalAndroidTestApp successfully to a device...but the XForms.Android app returned a null activity and couldn't load. does it work for you w/the changes?

strongname.cmd
henrik-me: that would enable the tests to run. Ideally it should be enough to have the dll name and the strongname identifier marked. I will continue testing... and perhaps will identify that I need to do more (so far this seems to work on my box).

verify.cmd
henrik-me: Next step will be to kick off the script which will install. For now I'm building out the awareness (and I just added the link to where to install from)

henrik-me:
I will take a new look at XForms app. It's likely that the last changes broke something...

verify.cmd
jennyf19: awesome

strongname.cmd
jennyf19: sounds good.

jennyf19 (approved):

jennyf19:
:shipit: 😸

henrik-me:
Added build scripts for vs dev cmd 2015, vs dev cmd 2017 and std. command-line and also modified various project references and targets. Made a set of additional changes.

PR #1057, opened 2018-05-18
Jennyf/update app

jennyf19:
Update to XForms iOS and Android Apps in MSAL to handle new Webview options. 
>Developer can do the following:
>1. System browser only
>2. Embedded Webview only
>3. Helper method, then launch either system browser, embedded webview, or custom error message

These options are [commented](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/compare/jennyf/updateApp?expand=1#diff-900ecb39a395f0a397c3d0abef33ffbcR52) out in the respective apps

**Tested with:**
* Android phone w/& w/out Chrome installed on device
* iOS simulator

### DIFF
--- a/msal/src/Microsoft.Identity.Client/UIParent.cs
+++ b/msal/src/Microsoft.Identity.Client/UIParent.cs
@@ -50,9 +50,8 @@ namespace Microsoft.Identity.Client
         }
 
 #if iOS
-        public UIParent(bool useEmbeddedWebview)
+        public UIParent(bool useEmbeddedWebview) : this()
         {
-            CoreUIParent = new CoreUIParent();
             CoreUIParent.UseEmbeddedWebview = useEmbeddedWebview;
         }
 #endif
@@ -73,7 +72,7 @@ namespace Microsoft.Identity.Client
             CoreUIParent = new CoreUIParent(Activity);
         }
 
-        public UIParent(Activity activity, bool useEmbeddedWebview) :this(activity)
+        public UIParent(Activity activity, bool useEmbeddedWebview) : this(activity)
         {
             CoreUIParent.UseEmbeddedWebview = useEmbeddedWebview;
         }
--- a/msal/src/Microsoft.Identity.Client/UIParent.cs
+++ b/msal/src/Microsoft.Identity.Client/UIParent.cs
@@ -52,6 +52,7 @@ namespace Microsoft.Identity.Client
 #if iOS
         public UIParent(bool useEmbeddedWebview)
         {
+            CoreUIParent = new CoreUIParent();
             CoreUIParent.UseEmbeddedWebview = useEmbeddedWebview;
         }
 #endif
--- a/msal/tests/dev apps/XForms/XForms.Android/MainActivity.cs
+++ b/msal/tests/dev apps/XForms/XForms.Android/MainActivity.cs
@@ -49,7 +49,7 @@ namespace XForms.Droid
 
             App.UIParent = new UIParent(Xamarin.Forms.Forms.Context as Activity);
 
-            #region Web browsers for MSAL.NET and Android
+            #region Web browsers for MSAL.NET Android
             // To activate embedded webview, remove '//' below
             //App.UIParent = new UIParent(Xamarin.Forms.Forms.Context as Activity, true);
 
--- a/msal/tests/dev apps/XForms/XForms.Android/MainActivity.cs
+++ b/msal/tests/dev apps/XForms/XForms.Android/MainActivity.cs
@@ -61,11 +61,13 @@ namespace XForms.Droid
             /*bool useEmbeddedWebview = UIParent.IsSystemWebviewAvailable();
             if (useEmbeddedWebview)
             {
-                App.UIParent = new UIParent(Xamarin.Forms.Forms.Context as Activity, true);
+                // Chrome present on device, use system browser
+                App.UIParent = new UIParent(Xamarin.Forms.Forms.Context as Activity);
             }
             else
             {
-                App.UIParent = new UIParent(Xamarin.Forms.Forms.Context as Activity);
+                // Chrome not present on device, use embedded webview
+                App.UIParent = new UIParent(Xamarin.Forms.Forms.Context as Activity, true);
             }*/
             #endregion
         }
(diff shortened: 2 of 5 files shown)


msal/tests/dev apps/XForms/XForms.Android/XForms.Android.csproj
```diff
@@ -28,7 +28,7 @@
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
     <OutputPath>bin\Debug\</OutputPath>
-    <DefineConstants>TRACE;DEBUG</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;ANDROID</DefineConstants>
```
henrik-me: ><DefineConstants>TRACE;DEBUG;ANDROID</DefineConstants> [](start = 4, length = 54)

move the definition of ANDROID the PropertyGroup above as that will ensure all configurations has it set.

henrik-me: same for iOS

---
In reply to: [189373909](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1057#discussion_r189373909) [](ancestors = 189373909)

msal/src/Microsoft.Identity.Client/UIParent.cs
```diff
@@ -52,6 +52,7 @@ public UIParent()
 #if iOS
         public UIParent(bool useEmbeddedWebview)
         {
+            CoreUIParent = new CoreUIParent();
```
henrik-me: >CoreUIParent = new CoreUIParent(); [](start = 12, length = 34)

call this after the constructor as that will do what the default constructor does

henrik-me:
Approving. Pls. see comments

msal/tests/dev apps/XForms/XForms.Android/XForms.Android.csproj
jennyf19: I was having issues getting the individual apps to pick up the ifdefs in UIParent, and found this as a solution...but now, i tried without and they work fine, so am going to remove this.

---
In reply to: [189373909](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1057#discussion_r189373909) [](ancestors = 189373909)

msal/src/Microsoft.Identity.Client/UIParent.cs
jennyf19: done. thanks.

---
In reply to: [189374400](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1057#discussion_r189374400) [](ancestors = 189374400)

msal/tests/dev apps/XForms/XForms.Android/MainActivity.cs
```diff
@@ -50,7 +46,34 @@ protected override void OnCreate(Bundle bundle)
 
             global::Xamarin.Forms.Forms.Init(this, bundle);
             LoadApplication(new App());
+
             App.UIParent = new UIParent(Xamarin.Forms.Forms.Context as Activity);
```
jmprieur: @jennyf19 @henrik-me 
I'm confused about what we are trying to explain in this section, and why the code does not use the help function (I'm assuming this is IsSystemWebviewAvailable() from the beginning?

jennyf19: This section describes the three ways developers can use the webview:
1. Use only system browser (this is default and is what we currently use)
2. Use only embedded Webview (noted in line 56)
3. Use helper method, noted in line 65-75

I left system browser as the default. Let me know if this is not clear or there is a different way of explaining this.

jmprieur: thanks for the explanation, @jennyf19 .
A lot of decisions are still left to the developer. If we get back to *why* you did this work, I believe that this is because there are devices where a demanded system web browser (Chrome tabs) is not available? So I wonder if the library itself should not use option 3 by itself - which, I agree is a change, but probably not in the cases where this works today?, and then let the developers override this choice? Or am I missing something?

jennyf19: @jmprieur - definitely don't want to add more complexity or work for the developers, but I don't think the library should decide for them to switch to embedded Webview, as this lowers the security bar, so the developer should be aware and want this functionality.
If the developer uses the helper method as I did in the sample, the user experience is seamless...embedded Webview will launch from the start. Let's discuss further.

henrik-me:
>     public sealed class UIParent

IMO: We should expose UseEmbeddedWebView as a property for the developer to flip. Perhaps even out to the AquireToken or PublicClient object....

---
Refers to: msal/src/Microsoft.Identity.Client/UIParent.cs:40 in 1babb49. [](commit_id = 1babb49734f38b6a4182a106952add0ac451f886, deletion_comment = False)

msal/tests/dev apps/XForms/XForms.Android/MainActivity.cs
henrik-me: Current default should be embedded webview imo as that is parity with ADAL, also the use of system browser may fail on devices where this is not setup/installed correctly. IMO we should ensure that the user on the devices still will be able to use the product or are properly notified on what they need to do to solve the problem (we can show a dialog or we can fall back to using the embedded webview in those cases). We can log and send telemetry that we did this (also we can have an option for developers to not allow this behavior.

jennyf19: @henrik-me System browser is more secure...Auth in embedded webview is not allowed for Google, so seems like we'd want MSAL to use the more secure model. 
The idea of the helper is that the developer can decide how the experience will be handled. Let's discuss this further next week w/ @jmprieur

jennyf19:
@henrik-me I was aiming to not change the api surface...am up to discussing this further.

msal/tests/dev apps/XForms/XForms.Android/MainActivity.cs
henrik-me: Security and whether the component is there or not... Customers will still have an issue thus the idea could be to show the user information about the missing components or allow the app developers to show the right message. Anyway, this should be aligned across libraries. 
Also can be different for Android vs. iOS as iOS always has a system browser. On iOS there is currently a debate on moving to SFAuthenticanSession. Is this planned?

jennyf19: @henrik-me Moving to SFAuthenticationSession is currently not planned...I can add a task for it. I agree about having similar experiences across libraries.

henrik-me:
IMO we should discuss and also it should be aligned across libraries

PR #1061, opened 2018-05-23
experiment to fix myget issue

jennyf19:
Build is passing in VSTS and successfully pushing to myGet, but is showing up red still...seeing if this will fix it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Active Directory Authentication Library (ADAL) provides easy to use authenticati
 
  Stable (`master` branch)    | Nightly (`dev` branch)
 -----------------------------|-------------------------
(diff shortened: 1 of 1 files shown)


henrik-me:
creative solution.... wasn't it pointing to the wrong build?

jennyf19:
that was my original thinking, but everything looks fine. In VSTS, it's using the adalv3/dev branch and the condition is set correctly. Also, myGet is on 3.19.6 now.

---
In reply to: [391538560](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1061#issuecomment-391538560) [](ancestors = 391538560)

PR #1066, opened 2018-05-24
Jennyf/nuget

jennyf19:
Updating MSAL Xamarin.Forms to 27.0.2 and MonoAndroid8.1
Renaming MSAL exceptions as Core exceptions
Adding core dll's to adal nuget


jennyf19:
Dev nightly failed because i didn't remove the changes for core msal dll's, which is not working currently. Cannot run a new build because i changed the monoandroid paths back to 7 for the other PRs.

Issue #936, opened 2018-01-05
Null reference when clearing token cache in ADAL before acquiring a token at least once


trwalke:
Clearing the token cache before calling acquiretokenasync at least once will result in a an unhandled null reference exception.
Reproduction steps:
1. Run the ADAL Automation app
2. Hit the clear cache button first.

The app will halt immediately and a null reference exception will be thrown

jmprieur:
This is fixed

PR #1075, opened 2018-06-02
Instantiating CoreLoggerBase.Default in TokenCache

trwalke:
ADAL crashes when clearing the cache before attempting to acquire a token. This is because the CoreLoggerBase.Default is only instantiated during acquiretokenasync calls and the TokenCache has the ability to clear its cache and log it before these calls are made. This proposal involves instantiating CoreLoggerBase.Default on the construction of the ADAL TokenCache to avoid null references when logging from the TokenCache.

This was tested by following the reproduction steps outlined in #936
https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/936


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -85,6 +90,8 @@ public TokenCache(byte[] state)
             : this()
         {
             this.Deserialize(state);
+            if (CoreLoggerBase.Default == null)
+                CoreLoggerBase.Default = new AdalLogger(Guid.Empty);
```
henrik-me: not needed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -76,6 +79,8 @@ static TokenCache()
         public TokenCache()
         {
             this.tokenCacheDictionary = new ConcurrentDictionary<AdalTokenCacheKey, AdalResultWrapper>();
+            if (CoreLoggerBase.Default == null)
+                CoreLoggerBase.Default = new AdalLogger(Guid.Empty);
```
henrik-me: should not be needed?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -68,6 +68,9 @@ static TokenCache()
                 BeforeAccess = StorageDelegates.BeforeAccess,
                 AfterAccess = StorageDelegates.AfterAccess
             };
+
+            if (CoreLoggerBase.Default == null)
+                CoreLoggerBase.Default = new AdalLogger(Guid.Empty);
```
henrik-me: should be enough?

henrik-me:
unit tests? other automated tests?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
trwalke: @henrik-me For the Automation, yes, this is enough. The others were just to ensure all bases are covered.

PR #1082, opened 2018-06-07
Jennyf/wk web view

jennyf19:
**Manual Testing With**

>iOS simulator
>iOS device


core/src/Platforms/iOS/EmbeddedWebview/AuthenticationAgentUIViewController.cs
```diff
@@ -64,90 +63,93 @@ public override void ViewDidLoad()
...
-                 && !request.Url.Scheme.Equals(Uri.UriSchemeHttps, StringComparison.OrdinalIgnoreCase))
-                {
-                    AuthorizationResult result = new AuthorizationResult(AuthorizationStatus.ErrorHttp);
-                    result.Error = MsalError.NonHttpsRedirectNotSupported;
-                    result.ErrorDescription = MsalErrorMessage.NonHttpsRedirectNotSupported;
-                    callbackMethod(result);
-                    this.DismissViewController(true, null);
-                    return false;
-                }
+                wkWebView.LoadRequest(new NSUrlRequest(new NSUrl(this.url)));
```
henrik-me: >wkWebView.LoadRequest(new NSUrlRequest(new NSUrl(this.url))); [](start = 16, length = 61)

here you are doing what you do above... why? 

(above the if(LoadView()...)

core/src/Platforms/iOS/EmbeddedWebview/AuthenticationAgentUIViewController.cs
```diff
@@ -64,90 +63,93 @@ public override void ViewDidLoad()
...
             View.BackgroundColor = UIColor.White;
 
-            webView = new UIWebView(View.Bounds);
-            webView.ShouldStartLoad = (wView, request, navType) =>
+            WKWebViewConfiguration webConfiguration = new WKWebViewConfiguration();
+
+            wkWebView = new WKWebView(View.Bounds, webConfiguration);
+            NSUrl url = wkWebView.Url;
+
+            wkWebView.LoadRequest(new NSUrlRequest(url));
```
henrik-me: >wkWebView.LoadRequest(new NSUrlRequest(url)); [](start = 12, length = 45)

perhaps the actual question is.... why is this to be loaded into the webview?

core/src/Platforms/iOS/EmbeddedWebview/AuthenticationAgentUIViewController.cs
```diff
@@ -64,90 +63,93 @@ public override void ViewDidLoad()
...
+
+            if (LoadView(url) == true)
             {
-                if (request == null)
-                {
-                    return true;
-                }
+                wkWebView.StopLoading();
```
henrik-me: >wkWebView.StopLoading(); [](start = 16, length = 24)

why is this needed? What loading is being stopped?

core/src/Platforms/iOS/EmbeddedWebview/AuthenticationAgentUIViewController.cs
jennyf19: I'll look into this. Was trying to keep what we had, but i'll go back and double check these things you brought up.

henrik-me:
>         }

Should this be exposed through the helper class instead of on the UIParent?

---
Refers to: msal/src/Microsoft.Identity.Client/UIParent.cs:104 in ba9cf8d. [](commit_id = ba9cf8d34799891cb8fb8568e0216abadd2c025b, deletion_comment = False)

core/src/Platforms/iOS/EmbeddedWebview/AuthenticationAgentUIViewController.cs
```diff
@@ -64,90 +63,93 @@ public override void ViewDidLoad()
...
+            if (requestUrlString.StartsWith(BrokerConstants.BrowserExtPrefix, StringComparison.OrdinalIgnoreCase))
             {
-                // If the title is too long, iOS automatically truncates it and adds ...
-                this.Title = webView.EvaluateJavascript(@"document.title") ?? "Sign in";
-            };
+                DispatchQueue.MainQueue.DispatchAsync(() => CancelAuthentication(null, null));
 
-            View.AddSubview(webView);
+                // Build the HTTPS URL for launching with an external browser
+                var httpsUrlBuilder = new UriBuilder(requestUrlString)
```
henrik-me: is this the broker scenarios? What is the purpose of this?

core/src/Platforms/iOS/EmbeddedWebview/AuthenticationAgentUIViewController.cs
```diff
@@ -64,90 +63,93 @@ public override void ViewDidLoad()
...
             {
-                if (request == null)
-                {
-                    return true;
-                }
+                wkWebView.StopLoading();
 
-                string requestUrlString = request.Url.ToString();
+                // If the title is too long, iOS automatically truncates it and adds ...
+                this.Title = wkWebView.EvaluateJavaScriptAsync(@"document.title").ToString();
```
bgavrilMS: EvaluateJavaScriptAsync returns a Task<string>, so calling toString() on it will return the Task.ToString() implementation, which is "System.Threading.Tasks.Task`1[System.String]"

What you want to do is to either: 

a) await the result
b) Wait(), potentially with a timeout
c) Use the non-async version, if there is one.

msal/tests/dev apps/XForms/XForms.Android/MainActivity.cs
```diff
@@ -47,13 +47,13 @@ protected override void OnCreate(Bundle bundle)
...
 
-            App.UIParent = new UIParent(Xamarin.Forms.Forms.Context as Activity);
+            App.UIParent = new UIParent(this);
 
             #region Web browsers for MSAL.NET Android
             // To activate embedded webview, remove '//' from line 56 below, 
             // and comment out line 50 above -> App.UIParent = new UIParent(Xamarin.Forms.Forms.Context as Activity);
 
-            //App.UIParent = new UIParent(Xamarin.Forms.Forms.Context as Activity, true);
+            //App.UIParent = new UIParent(this, true);
```
bgavrilMS: This is why dead code is difficult to maintain...

core/src/Platforms/iOS/EmbeddedWebview/AuthenticationAgentUIViewController.cs
```diff
@@ -64,90 +63,93 @@ public override void ViewDidLoad()
...
-            webView = new UIWebView(View.Bounds);
-            webView.ShouldStartLoad = (wView, request, navType) =>
+            WKWebViewConfiguration webConfiguration = new WKWebViewConfiguration();
+
+            wkWebView = new WKWebView(View.Bounds, webConfiguration);
+            NSUrl url = wkWebView.Url;
+
+            wkWebView.LoadRequest(new NSUrlRequest(url));
+
+            if (LoadView(url) == true)
```
bgavrilMS: no need for == true

core/src/Platforms/iOS/EmbeddedWebview/AuthenticationAgentUIViewController.cs
```diff
@@ -64,90 +63,93 @@ public override void ViewDidLoad()
...
             {
-                if (request == null)
-                {
-                    return true;
-                }
+                wkWebView.StopLoading();
 
-                string requestUrlString = request.Url.ToString();
+                // If the title is too long, iOS automatically truncates it and adds ...
```
bgavrilMS: In the previous implementation, if the title was empty we used "Sing In". Why aren't we doing this anymore?

core/src/Platforms/iOS/EmbeddedWebview/AuthenticationAgentUIViewController.cs
jennyf19: yes...good catch...thanks :)

core/src/Platforms/iOS/EmbeddedWebview/AuthenticationAgentUIViewController.cs
```diff
@@ -64,90 +63,93 @@ public override void ViewDidLoad()
...
-                    callbackMethod(result);
-                    this.DismissViewController(true, null);
-                    return false;
-                }
+                wkWebView.LoadRequest(new NSUrlRequest(new NSUrl(this.url)));
 
-                return true;
-            };
+                // if this is false, page will be 'zoomed in' to normal size
+                //webView.ScalesPageToFit = true;
```
bgavrilMS: dead code, either use it or remove it.

jennyf19: will do. thanks

msal/tests/dev apps/XForms/XForms.Android/MainActivity.cs
henrik-me: This is the test app. And for different scenarios you would comment/uncomment. To Bogdans comment: Making some ui that switches between the scenarios instead of this? It will require some changes to the app but perhaps it's worth it... 

---
In reply to: [194104305](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1082#discussion_r194104305) [](ancestors = 194104305)

Issue #1085, opened 2018-06-11
Can we please bring back method overload AcquireTokenAsync(string, string, UserCredential)?


TimLovellSmith:
Now when I try to use recent Microsoft Azure Rest SDK versions with higher versions of ADAL, like the latest, 3.19.6 I get this compatibility error.

"Method not found: 'System.Threading.Tasks.Task`1<Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationResult> Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationContext.AcquireTokenAsync(System.String, System.String, Microsoft.IdentityModel.Clients.ActiveDirectory.UserCredential)'."

These binary breaking changes are very painful to work through!

jmprieur:
hello, @TimLovellSmith : 
- what was the previous version you were using? would it be ADAL v2.x ?
- The method is still there (as an extension method): https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/62a82c78b2a8504065f342309f8015dde8b9eccc/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Features/WinCommon/AuthenticationContextIntegratedAuthExtensions.cs#L48
- It's available in .NET 4.5 and .NET Standard 1.3

Also which platform are you targeting (Xamarin.Android?). We've seen 2 cases where NuGet restore (the NuGet UI through Visual Studio, actually) did not restore the right platform.

cc: @henrik-me @bgavrilMS @jennyf19

TimLovellSmith:
@jmprieur Unfortunately, being there as an extension method just isn't binary compatible, and the call to it is not from our code, it is from a DLL we depend on. An extension method is basically syntactic sugar for calling a static method, where the static method is on a different class. When you look at the compiled IL it is trying to call an instance method on the AuthenticationContext class, not a static method.
I'm targeting Windows desktop .net 4.6.2.

jmprieur:
@TimLovellSmith 
you should expect breaking changes when moving from ADAL 2.x to ADAL 3.x as the major number was incremented. ADAL uses semantinc versioning (See https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/wiki/Semantic-versioning.-API-change-management)

I agree that changing a method to an extension method is a binary breaking change. (I wonder why this was done ...)

Can you tell us more about your scenario? 
[...]

PR #1086, opened 2018-06-12
Fix binary compat with ADAL 2.x for calls to AuthenticationContext.AcquireTokenAsync

TimLovellSmith:
Because I believe if it is not necessary to break binary compat, it is just better not to! For discussion see #1085.


henrik-me:
Thanks for pointing out the issues with binary compatibility.

"fixing" this will need more work. Also there are more breaking changes between v2 and v3 than this. 
At this point this change would break existing v3.
Aside: the current dev branch is for vNext of ADAL and MSAL. (Which may even break v3 and thus we may be able to accomodate this change)

TimLovellSmith:
@henrik-me Hi Henrik, which one is the v3 branch? And how does this break v3?

henrik-me:
adalv3dev is the current v3 branch. because the change is going into v3 (or later) then IMO this is as breaking as when the change went the other way between 2 and 3.

TimLovellSmith:
@henrik-me What exactly is the breaking change that you think I'm making?

henrik-me:
See discussion in #1085. Closing this PR. It's something we may consider for v4 as a general element of removing extension methods.

PR #1087, opened 2018-06-12
Hotfix 3.19.7

rayluo:
This hotfix will:

* No longer clear SecurePassword after the username password grant flow
* Keep whitespaces inside a SAML token. This will bring better interoperability with other federated IDPs.

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/WsTrust/WsTrustResponse.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/WsTrust/WsTrustResponse.cs
@@ -95,7 +95,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.WsTrust
         {
             try
             {
-                return XDocument.Load(responseStream, LoadOptions.None);
+                return XDocument.Load(responseStream, LoadOptions.PreserveWhitespace);
             }
             catch (XmlException ex)
             {
@@ -148,9 +148,17 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.WsTrust
                     {
                         continue;
                     }
-                    
-                    tokenResponseDictionary.Add(tokenTypeElement.Value,
-                        requestedSecurityToken.FirstNode.ToString(SaveOptions.DisableFormatting));
+
+                    var token = new System.Text.StringBuilder();
+                    foreach (var node in requestedSecurityToken.Nodes())
+                    {
+                        // Since we moved from XDocument.Load(..., LoadOptions.None) to Load(..., LoadOptions.PreserveWhitespace),
+                        // requestedSecurityToken can contain multiple nodes, and the first node is possibly just whitespaces e.g. "\n   ",
+                        // so we concatenate all the sub-nodes to include everything
+                        token.Append(node.ToString(SaveOptions.DisableFormatting));
+                    }
+
+                    tokenResponseDictionary.Add(tokenTypeElement.Value, token.ToString());
                 }
             }
             catch (XmlException ex)
--- a/tests/Test.ADAL.NET.Unit/WsTrustTests.cs
+++ b/tests/Test.ADAL.NET.Unit/WsTrustTests.cs
@@ -29,6 +29,7 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.WsTrust;
 using System.IO;
 using System.Xml.Linq;
+using System.Text;
 
 namespace Test.ADAL.NET.Unit
 {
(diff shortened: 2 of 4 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/WsTrust/WsTrustResponse.cs
```diff
@@ -148,9 +148,17 @@ internal static WsTrustResponse CreateFromResponseDocument(XDocument responseDoc
...
+                    var token = new System.Text.StringBuilder();
+                    foreach (var node in requestedSecurityToken.Nodes())
+                    {
+                        // Since we moved from XDocument.Load(..., LoadOptions.None) to Load(..., LoadOptions.PreserveWhitespace),
+                        // requestedSecurityToken can contain multiple nodes, and the first node is possibly just whitespaces e.g. "\n   ",
+                        // so we concatenate all the sub-nodes to include everything
+                        token.Append(node.ToString(SaveOptions.DisableFormatting));
+                    }
+
+                    tokenResponseDictionary.Add(tokenTypeElement.Value, token.ToString());
```
jmprieur: this is a dictionary? are we sure we won't have several tokens of the same type?

jmprieur (commented):
LGTM. just a question.

jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/WsTrust/WsTrustResponse.cs
```diff
@@ -148,9 +148,17 @@ internal static WsTrustResponse CreateFromResponseDocument(XDocument responseDoc
                     {
                         continue;
                     }
-                    
-                    tokenResponseDictionary.Add(tokenTypeElement.Value,
-                        requestedSecurityToken.FirstNode.ToString(SaveOptions.DisableFormatting));
+
```
bgavrilMS: Are there no tests for this method? Could you write some - it shouldn't be too difficult since it's a static method that takes in just an xml and a version. Boy scout rule :)

bgavrilMS (changes requested):
Logic added is non-trivial, some unit tests would be great.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/WsTrust/WsTrustResponse.cs
rayluo: Test case added. Am I a good boy now? :)

tests/Test.ADAL.NET.Unit/WsTrustTests.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        [TestMethod]
+        [TestCategory("WsTrustTests")]
+        public void TestCreateFromResponseDocument_WhenInputContainsWhitespace_ShouldPreserveWhitespace()
+        {
+            string sample = File.ReadAllText("WsTrustResponse.xml");
+            string characteristic = "\n        <saml:Assertion";
+            StringAssert.Contains(sample, characteristic);
+            WsTrustResponse resp = WsTrustResponse.CreateFromResponseDocument(
+                XDocument.Parse(sample, LoadOptions.PreserveWhitespace), WsTrustVersion.WsTrust2005);
```
bgavrilMS: You can invoke and test ReadDocumentFromResponse as well to read the actual file instead of doing XDocument.Parse

tests/Test.ADAL.NET.Unit/WsTrustTests.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Xml.Linq;
+
+namespace Test.ADAL.NET.Unit
+{
+    [TestClass]
+    [DeploymentItem("WsTrustResponse.xml")]
+    public class WsTrustTests
+    {
+        [TestMethod]
+        [TestCategory("WsTrustTests")]
```
bgavrilMS: This file under test has a single t:RequestedSecurityToken, but in your code you have a foreach that deals with the case of multiple t:RequestedSecurityToken. It would be good for a test to also look at this case.

bgavrilMS (approved):
Added some more comments, approving as to not block the hot fix

tests/Test.ADAL.NET.Unit/WsTrustTests.cs
rayluo: Hmm, if you are referring to [line 153 of this PR](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1087/files#diff-546f2adc268992cd2106ba3513ae23e2R153), that line:

> foreach (var node in requestedSecurityToken **.Nodes()** )

actually deals with the case of multiple nodes inside ONE "RequestedSecurityToken". (I'm sorry if that trailing `.Nodes()` part might not be considered obvious.)

Besides, the test sample has been around [for 2 years](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/commits/v3.17.3/tests/Test.ADAL.NET.Unit/WsTrustResponse.xml), and I believe it was modeled from real world case.

So, do we still need to have extra test sample here?

tests/Test.ADAL.NET.Unit/WsTrustTests.cs
rayluo: OK I see your point on also testing the changed method `ReadDocumentFromResponse()`. The upcoming new commit will do it, but in a slightly different way. Instead of testing both `ReadDocumentFromResponse()` and `CreateFromResponseDocument()` in one test case, a new test case will be added to test the higher level `CreateFromResponse()` which invokes both lower level helpers. Hope that helps.

PR #1088, opened 2018-06-12
Enable msal packaging by laying out assemblies via a build target

bgavrilMS:
Entry point is from VSTS or via msbuild /t:LayoutNuget


core/src/Microsoft.Identity.Core.csproj
```diff
@@ -9,21 +9,12 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <TargetFrameworks>net45;netstandard1.1;netstandard1.3;win81;MonoAndroid8.1;Xamarin.iOS10</TargetFrameworks>
```
bgavrilMS: This was used for msbuild /t:pack, but that route won't work, so I moved all this package metadata to a nuspec file

msal/build/VerifyAssemblySigning.proj
```diff
@@ -1,50 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
bgavrilMS: We don't need to do assembly signing verification - the signing is done as part of the VSTS sign task

msal/build/prepare-versioning.ps1
```diff
@@ -1,15 +0,0 @@
-Write-Host "=========================="
```
bgavrilMS: Versioning can be done without hacking it

msal/src/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -8,39 +8,43 @@
...
+    <RemoveDir Directories="$(NugetPackageRootFolder)" />
+    <Copy SourceFiles="@(FilesToPack)" DestinationFiles="$(NugetPackageRootFolder)\lib\%(RecursiveDir)%(FileName)%(Extension)" />
+    <Copy SourceFiles="$(NuspecFilePath)" DestinationFolder="$(NugetPackageRootFolder)" />
+  </Target>
+  
 
   <PropertyGroup>
     <TargetFrameworks>net45;netstandard1.1;netstandard1.3;win81;MonoAndroid8.1;Xamarin.iOS10</TargetFrameworks>
-    <GenerateDocumentationFile Condition=" '$(Configuration)' == 'Release' ">true</GenerateDocumentationFile>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
bgavrilMS: It takes 0.2 seconds to generate those xml files :)

core/src/Microsoft.Identity.Core.csproj
jennyf19: Thanks for all the additional comments and explanation throughout...very helpful

jmprieur: How does the compiler know that it needs to generate DLLs for core for all these platforms if the TargetFrameworks property does not appear?

msal/build/VerifyAssemblySigning.proj
jmprieur: I believe that Duncan had had it because at that time the code sign-in task could be successful but the signing not be done (in some weird cases)

msal/src/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
jmprieur: and they are useful for IntelliSense

core/src/Microsoft.Identity.Core.csproj
bgavrilMS: The <TargetFrameworks> was not removed, you can see it lower down. The GitHub diff view mistakenly thinks I deleted it and added it back lower down (now on line 12 in the Unified Diff view)

PR #1089, opened 2018-06-12
Some binary compatibility fix between ADAL 2.x and ADAL 3.x 

TimLovellSmith:
for calls to AuthenticationContext.AcquireTokenAsync(string, string, UserCredential).


bgavrilMS:
As discussed, the assemblies are strongly named so a recompilation will be necessary. It would be better to recompile the app and migrate to ADAL v3.

TimLovellSmith:
@bgavrilMS Recompilation won't actually be necessary when the *reference* to the assembly isn't referring by its strong name right? If trying to use ADALv3 instead of ADALv2 was failing due to strong naming issues the issue I am seeing would not be MissingMethodException!

PR #1093, opened 2018-06-18
Unified ADAL MSAL build - update the ADAL build, versioning and packaging

bgavrilMS:
1. Version will be controlled through the csproj
2. Separate versioning for core, adal and msal
3. Packaging


adal/build/LayoutNugetPackage.proj
```diff
@@ -1,290 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
bgavrilMS: This is too complex - we don't need to list every file that gets copied, it is sufficient to copy the output.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory.Platform/Microsoft.IdentityModel.Clients.ActiveDirectory.Platform.csproj
```diff
@@ -10,24 +10,28 @@
...
     <UseFullSemVerForNuGet>false</UseFullSemVerForNuGet>
     <DebugType>full</DebugType>
     <RootNamespace>Microsoft.IdentityModel.Clients.ActiveDirectory</RootNamespace>
-	<EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+	  <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+  </PropertyGroup>
+
+  <PropertyGroup Label="GenerateAssemblyInfo inputs">
+    <!-- This should be passed from the VSTS build -->
+    <AdalClientSemVer Condition="'$(AdalClientSemVer)' == ''">1.0.0-localbuild</AdalClientSemVer>
```
bgavrilMS: The build generates the AssemblyInfo attributes related to version based on this. The release build also passes this version to the packaging task.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Properties/AssemblyInfo.cs
```diff
@@ -29,16 +29,9 @@
 using System.Reflection;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyProduct("Active Directory Authentication Library")]
```
bgavrilMS: Generated to csproj

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Properties/AssemblyInfo.cs
```diff
@@ -29,16 +29,9 @@
...
-[assembly: AssemblyConfiguration("")]
-[assembly: AssemblyCulture("")]
-[assembly: AssemblyCompany("Microsoft Corporation")]
-[assembly: AssemblyCopyright("Copyright (c) Microsoft Corporation. All rights reserved.")]
-[assembly: AssemblyTrademark("")]
-[assembly: AssemblyMetadata("Serviceable", "True")]
+// Version and Metadata are set at build time from msbuild properties defined in the csproj
 
-[assembly: AssemblyFileVersion("3.17.3.0")]
+[assembly: AssemblyMetadata("Serviceable", "True")]
```
jmprieur: What is it for?

msal/src/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -151,21 +151,20 @@
...
-      <NugetPackageRootFolder Condition="'$(NugetPackageRootFolder)' == ''">$([System.IO.Path]::Combine($(MSBuildThisFileDirectory), "..", "..", "layout"))</NugetPackageRootFolder>
-      <PackagePdb Condition="'$(PackagePdb)'==''">false</PackagePdb>
+      <NugetPackageRootFolder Condition="'$(NugetPackageRootFolder)' == ''">$([System.IO.Path]::Combine($(MSBuildThisFileDirectory), "..",".nugetlayout"))</NugetPackageRootFolder>
       <NuspecFilePath>$([System.IO.Path]::Combine($(MSBuildThisFileDirectory), "..", "..", "build", "Microsoft.Identity.Client.nuspec"))</NuspecFilePath>
     </PropertyGroup>
     <ItemGroup>
       <FilesToPack Include="bin\$(Configuration)\**\Microsoft.Identity.Client.dll" />
       <FilesToPack Include="bin\$(Configuration)\**\Microsoft.Identity.Client.xml" />
-      <FilesToPack Include="bin\$(Configuration)\**\Microsoft.Identity.Client.pdb" Condition="'$(PackagePdb)' == 'true' " />
+      <FilesToPack Include="bin\$(Configuration)\**\Microsoft.Identity.Client.pdb" />
```
jmprieur: Did we decide to systematically package the PDBs? (I'm all for it, BTW)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory.Platform/Microsoft.IdentityModel.Clients.ActiveDirectory.Platform.csproj
```diff
@@ -10,24 +10,28 @@
...
     <Product>Microsoft.IdentityModel.Clients.ActiveDirectory.Platform ($(TargetFramework))</Product>
-    <GenerateDocumentationFile Condition=" '$(Configuration)' == 'Release' ">true</GenerateDocumentationFile>
-    <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
-    <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
-    <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
-    <GenerateAssemblyInformationalVersionAttribute>false</GenerateAssemblyInformationalVersionAttribute>
-    <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
-    <GenerateAssemblyDescriptionAttribute>false</GenerateAssemblyDescriptionAttribute>
-    <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
jmprieur: Thanks!
the debugging experience with the XML comments and PDB is way better in VS

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Properties/AssemblyInfo.cs
bgavrilMS: I believe this is related to Windows Security Updates - https://github.com/dotnet/cli/issues/3345

bgavrilMS: Not sure if it applies to our stuff, I believe it is related to GACed assemblies: 


_The CLR identifies Microsoft published NuGet libraries via a special custom attribute created for this purpose. When a library with this attribute is loaded, the CLR records the load event as a marker on the machine, specifically the full assembly name. The marker information never leaves the machine, nor is it sent to Microsoft in any form.

At the point that we release a security update for a NuGet library, Microsoft Update will use the presence of the marker for that library as an indication that a given machine is a candidate for that update. The check for markers is done on the client machine. Markers that don’t have an associated security update are never consulted or used.

Microsoft Update installs updated assemblies into the Global Assembly Cache (GAC), to make them available to all users and apps on the machine that use a given .NET NuGet library. The update includes publisher policy, which is also installed to the GAC. Publisher policy redirects older, vulnerable, versions of libraries to new, updated, versions. The policy is intended to be as targeted as possible, and will not apply to older versions that are not vulnerable, since that is not necessary._

msal/src/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
bgavrilMS: The packaging task are creating 2 packages - one with DLL and XML, and one with DLL, PDB and XML. I figured this behaviour is what we always want and there's no need to control copying the pdbs.

bgavrilMS:
Manually merged

PR #1096, opened 2018-06-21
Fix a minor globalization issue and trivial missing comments

bgavrilMS:
Make FxCop happy...


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/PlatformParameters.cs
```diff
@@ -39,6 +39,9 @@ public class PlatformParameters : IPlatformParameters
     {
         internal CoreUIParent CoreUIParent { get; private set; }
 
+        /// <summary>
+        /// Constructor
```
jmprieur: We are not going to remove it, but I question the value of this default constructor. No parent activity, does not set the broker and prompt behavior. I don't understand the scenarios?

PR #1101, opened 2018-06-22
Jennyf/wk web view update

jennyf19:
**Manual testing with:**
> iOS device and simulator
> iOS device w/broker


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/OAuth2/TokenResponse.cs
```diff
@@ -140,7 +140,9 @@ internal static TokenResponse CreateFromBrokerResponse(IDictionary<string, strin
                 CorrelationId = responseDictionary["correlation_id"],
                 Resource = responseDictionary["resource"],
                 ExpiresOn = long.Parse(responseDictionary["expires_on"].Split('.')[0], CultureInfo.CurrentCulture),
-                ClientInfo = responseDictionary["client_info"]
+                ClientInfo = responseDictionary.ContainsKey("client_info")
+                ? Authenticator.EnsureUrlEndsWithForwardSlash(EncodingHelper.UrlDecode(responseDictionary["client_info"]))
```
henrik-me: >Authenticator.EnsureUrlEndsWithForwardSlash(EncodingHelper.UrlDecode [](start = 18, length = 68)

I'm wondering how we are finding this to be needed now. was this fixed in ADAL v3?

henrik-me:
unit tests? other tests? a manual test case?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/OAuth2/TokenResponse.cs
jennyf19: It was added as part of the cache changes, but broker does not return client_info, so we need to add a check here for that. The next broker release will contain client_info.

core/src/Platforms/iOS/EmbeddedWebview/AuthenticationAgentUIViewController.cs
```diff
@@ -25,28 +25,23 @@
...
-using CoreGraphics;
 using Foundation;
+using Microsoft.Identity.Core.Platforms.iOS.EmbeddedWebview;
+using System;
 using UIKit;
-using Microsoft.Identity.Core.Helpers;
+using WebKit;
+using static Microsoft.Identity.Core.Platforms.iOS.EmbeddedWebview.WKWebNavigationDelegate;
 
 namespace Microsoft.Identity.Core.UI.EmbeddedWebview
```
henrik-me: >Microsoft.Identity.Core.UI.EmbeddedWebview [](start = 10, length = 42)

seems like the two namespaces for the WKWebNavigationDeletate and this class is not the same. Any reason why?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/OAuth2/TokenResponse.cs
henrik-me: I was more thinking about the actual change to now url ends with forward slash change. what is the reason for that?

---
In reply to: [197549673](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1101#discussion_r197549673) [](ancestors = 197549673)

henrik-me: .. should the same change be done in current ADAL?

---
In reply to: [197550823](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1101#discussion_r197550823) [](ancestors = 197550823,197549673)

jennyf19: Will fix this...just need to do a null check. thanks.

jennyf19:
I listed the manual tests above. I tested w/iOS simulator and device and iOS device w/broker (authenticator app) w/adal

---
In reply to: [399560036](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1101#issuecomment-399560036) [](ancestors = 399560036)

henrik-me:
I know I were cryptic above. My comment really more reflects how are we in the future going to ensure we have proper regression testing done on the system browser/embedded webview scenarios? Which test automation has been planned for this feature or which test case has been created to ensure we properly test this regularly in the future

---
In reply to: [399562902](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1101#issuecomment-399562902) [](ancestors = 399562902,399560036)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/OAuth2/TokenResponse.cs
jennyf19: no because client_info was added was for cache unification, and current broker does not return client_info anyway. Once broker does return client_info we will be on next version of adal anyway.

---
In reply to: [197551638](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1101#discussion_r197551638) [](ancestors = 197551638,197550823,197549673)

jennyf19:
I see. I will create something together for test automation. I'll focus on adal first, since that will only use embedded webview anyway.

---
In reply to: [399563890](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1101#issuecomment-399563890) [](ancestors = 399563890,399562902,399560036)

core/src/Platforms/iOS/EmbeddedWebview/AuthenticationAgentUIViewController.cs
jennyf19: fixed...thanks.

henrik-me (approved):
:shipit:

adal/tests/Test.ADAL.NET.Integration/DeviceProfileTests.cs
```diff
@@ -48,6 +50,15 @@ public void Initialize()
             AdalHttpMessageHandlerFactory.AddMockHandler(MockHelpers.CreateInstanceDiscoveryMockHandler(TestConstants.GetDiscoveryEndpoint(TestConstants.DefaultAuthorityCommonTenant)));
         }
 
+        [TestCleanup()]
+        public void Cleanup()
+        {
+            if (context != null && context.TokenCache != null)
```
henrik-me, line 56: i asked initially to get this introduced... we may have to let each test handle cleaning up token cache or detect how to clean only the default handler.

PR #1103, opened 2018-06-26
move callback invocation to completionHandler of DismissViewController

jennyf19:
**Manually tested with:**
>ADAL iOS test app on simulator
>MSAL iOS test app on simulator w/embedded webview

#1016

### DIFF
--- a/core/src/Platforms/iOS/EmbeddedWebview/WKWebNavigationDelegate.cs
+++ b/core/src/Platforms/iOS/EmbeddedWebview/WKWebNavigationDelegate.cs
@@ -73,7 +73,8 @@ namespace Microsoft.Identity.Core.UI.EmbeddedWebview
                    requestUrlString.StartsWith(BrokerConstants.BrowserExtInstallPrefix, StringComparison.OrdinalIgnoreCase))
             {
                 AuthenticationAgentUIViewController.callbackMethod(new AuthorizationResult(AuthorizationStatus.Success, requestUrlString));
-                AuthenticationAgentUIViewController.DismissViewController(true, null);
+                AuthenticationAgentUIViewController.DismissViewController(true, () =>
+                    AuthenticationAgentUIViewController.callbackMethod(new AuthorizationResult(AuthorizationStatus.Success, requestUrlString)));
                 decisionHandler(WKNavigationActionPolicy.Cancel);
                 return;
             }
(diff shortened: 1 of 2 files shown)


core/src/Platforms/iOS/EmbeddedWebview/WKWebNavigationDelegate.cs
```diff
@@ -73,7 +73,8 @@ public override void DecidePolicy(WKWebView webView, WKNavigationAction navigati
                    requestUrlString.StartsWith(BrokerConstants.BrowserExtInstallPrefix, StringComparison.OrdinalIgnoreCase))
             {
                 AuthenticationAgentUIViewController.callbackMethod(new AuthorizationResult(AuthorizationStatus.Success, requestUrlString));
```
henrik-me: >AuthenticationAgentUIViewController.callbackMethod(new AuthorizationResult(AuthorizationStatus.Success, requestUrlString)); [](start = 16, length = 123)

remove ? Seems like you have two calls to callBackMethod #Resolved

core/src/Platforms/iOS/EmbeddedWebview/WKWebNavigationDelegate.cs
```diff
@@ -73,7 +73,8 @@ public override void DecidePolicy(WKWebView webView, WKNavigationAction navigati
...
-                AuthenticationAgentUIViewController.DismissViewController(true, null);
+                AuthenticationAgentUIViewController.DismissViewController(true, () =>
+                    AuthenticationAgentUIViewController.callbackMethod(new AuthorizationResult(AuthorizationStatus.Success, requestUrlString)));
                 decisionHandler(WKNavigationActionPolicy.Cancel);
```
henrik-me: >Cancel [](start = 57, length = 6)

Should this be cancel and not Ok/Success or something similar? #Resolved

henrik-me (changes requested):
:clock1:

core/src/Platforms/iOS/EmbeddedWebview/WKWebNavigationDelegate.cs
jennyf19: done...thanks. Forgot to remove that line.  #Resolved

core/src/Platforms/iOS/EmbeddedWebview/WKWebNavigationDelegate.cs
jennyf19: Success is what we've had, so I don't think we should change it. we can discuss off-line though. #Resolved

henrik-me (approved):
:shipit:

PR #1105, opened 2018-06-28
Add Api Backwards Compat tests

bgavrilMS:


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "1.0.0-beta5",
+    "version": "2.0.7",
```
bgavrilMS, line 3: This global.json is an artifact from the PCL world, which is needed to build in VS2015 (I think). However, if you use the "dotnet" cmd line, it complains about this SDK. Updated it to a more recent version. This does not have any impact on the libraries.

jmprieur (approved):
LGTM

bgavrilMS:
Note that the tests fail because they have found a discrepancy - we have forgotten to merge a hotfix commit from master to adalv3/dev. Please see https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1106 for this.

global.json
jennyf19: Great

jennyf19 (approved):
:shipit: 
Please commit changes to dev branch as well 
Is there a reason the mobile platforms and netstandard1.3 are not included?

bgavrilMS:
Note: for dev branch, we don't really have a baseline out there and it would be too much work to make the tool work. I think we should simply manually check binary compat on our first release and then start running these tests. 

Unfortunately reflecting over the mobile platform DLLs is a big pain because the CLR doesn't know where to find the frameworks and we don't really control assembly loading. For now, we can only check et45, netcore45 and netstandard11. I'm still looking at a solution of the other platforms.

jennyf19:
@bgavrilMS Sounds good...thanks for answering my questions. Really glad we have this now for adalv3/dev.

PR #1106, opened 2018-06-29
Cherry pick the latest HotFix from master into adalv3/dev

bgavrilMS:
adalv3/dev is not in a good state because it has diverged from master. The APICheck tests are failing for this reason.


jennyf19 (approved):
🚢 Thanks @bgavrilMS

PR #1108, opened 2018-07-02
Somkape/unif

SomkaPe:
- return one Account per user
- making .net msal ios persistence compatible with Ios msal lib implementation


core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -238,29 +237,46 @@ internal class CacheFallbackOperations
         }
 
         public static AdalResultWrapper FindMsalEntryForAdal(ITokenCacheAccessor tokenCacheAccessor, string authority,
-            string clientId, string upn)
+            string clientId, string upn, RequestContext requestContext)
```
bgavrilMS, line 218: We should get into the habit of adding comments to all public and internal methods (public for customers, internal for the team).  #Pending

bgavrilMS, line 218: Having so many parameters + static is a code smell - consider refactoring to an instance class with state.  #Pending

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -238,29 +237,46 @@ internal class CacheFallbackOperations
...
+            List<MsalAccountCacheItem> accounts = new List<MsalAccountCacheItem>();
+            foreach (string accountStr in tokenCacheAccessor.GetAllAccountsAsString())
+            {
+                var accountItem = JsonHelper.TryToDeserializeFromJson<MsalAccountCacheItem>(accountStr, requestContext);
+                if (accountItem != null && accountItem.Environment.Equals(environment))
+                {
+                    accounts.Add(accountItem);
+                }
+            }
+            if (accounts.Count > 0)
```
bgavrilMS, line 231: accounts.Any() is the preffered way to check if a collection is non empty, because it avoids iterating the IEnumerable #Pending

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -238,29 +237,46 @@ internal class CacheFallbackOperations
...
 
         public static AdalResultWrapper FindMsalEntryForAdal(ITokenCacheAccessor tokenCacheAccessor, string authority,
-            string clientId, string upn)
+            string clientId, string upn, RequestContext requestContext)
         {
-            foreach (var rtString in tokenCacheAccessor.GetAllRefreshTokensAsString())
-            {
-                var rtCacheItem =
-                    JsonHelper.DeserializeFromJson<MsalRefreshTokenCacheItem>(rtString);
+            var environment = new Uri(authority).Host;
```
bgavrilMS: Using "var" in production code makes the code difficult to read. Consider using the actual type. For example in this case I had to google the fact Uri.Host is actually a string  #Pending

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -238,29 +237,46 @@ internal class CacheFallbackOperations
...
 
-                var accountStr = tokenCacheAccessor.GetAccount(rtCacheItem.GetAccountItemKey());
-
-                var accountItem =
-                    JsonHelper.DeserializeFromJson<MsalAccountCacheItem>(accountStr);
+            List<MsalAccountCacheItem> accounts = new List<MsalAccountCacheItem>();
+            foreach (string accountStr in tokenCacheAccessor.GetAllAccountsAsString())
+            {
+                var accountItem = JsonHelper.TryToDeserializeFromJson<MsalAccountCacheItem>(accountStr, requestContext);
+                if (accountItem != null && accountItem.Environment.Equals(environment))
```
bgavrilMS: string comparison - use culture information #Pending

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -238,29 +237,46 @@ internal class CacheFallbackOperations
...
+                {
+                    var rtCacheItem =
+                        JsonHelper.TryToDeserializeFromJson<MsalRefreshTokenCacheItem>(rtString, requestContext);
 
-                //TODO - authority check needs to be updated for alias check
-                if (new Uri(authority).Host.Equals(rtCacheItem.Environment) && rtCacheItem.ClientId.Equals(clientId) &&
-                    accountItem.PreferredUsername.Equals(upn))
-                    return new AdalResultWrapper
+                    //TODO - authority check needs to be updated for alias check
+                    if (rtCacheItem != null && environment.Equals(rtCacheItem.Environment)
```
bgavrilMS: string comparison without culture - please review all string comparisons, there are quite a few. #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
```diff
@@ -29,6 +29,7 @@
 using System.Collections.ObjectModel;
 using System.Runtime.Serialization;
 using Microsoft.Identity.Core.Cache;
+using Microsoft.Identity.Core.Helpers;
 
 namespace Microsoft.Identity.Core
```
bgavrilMS, line 34: Please document these classes since you are in the position to understand what they all do #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
```diff
@@ -46,56 +47,90 @@ internal class TokenCacheAccessor : ITokenCacheAccessor
         internal readonly IDictionary<string, string> AccountCacheDictionary =
```
bgavrilMS, line 47: I know you haven't touched this, this code per se, but I wanted to raise the point of internal methods / members / properties inside internal classes. 

```
internal class Foo { 
    internal void Bar() {}    
    public void Buz() {}
} 
```
There is no difference between the visiblity of Bar and Buz. Still, we seem to be mixing up the 2, for example the dictionaries are internal and the methods public. We should consistent (personally I prefer public, so if we ever decide to make a class public, you only have to make 1 change)

 #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
```diff
@@ -46,56 +47,90 @@ internal class TokenCacheAccessor : ITokenCacheAccessor
...
             new ConcurrentDictionary<string, string>();
 
-        public void SaveAccessToken(string cacheKey, string item)
+        public void SaveAccessToken(MsalAccessTokenCacheItem item)
         {
-            AccessTokenCacheDictionary[cacheKey] = item;
+            AccessTokenCacheDictionary[item.GetKey().ToString()] = JsonHelper.SerializeToJson(item);
```
bgavrilMS, line 52: These dictionaries should be exposed as read-only properties (i.e. get, private set) not as members... #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
```diff
@@ -46,56 +47,90 @@ internal class TokenCacheAccessor : ITokenCacheAccessor
...
         }
 
-        public void SaveRefreshToken(string cacheKey, string item)
+        public void SaveRefreshToken(MsalRefreshTokenCacheItem item)
         {
-            RefreshTokenCacheDictionary[cacheKey] = item;
+            RefreshTokenCacheDictionary[item.GetKey().ToString()] = JsonHelper.SerializeToJson(item);
```
bgavrilMS, line 57: Is this fast enough for web servers etc. ? #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
```diff
@@ -46,56 +47,90 @@ internal class TokenCacheAccessor : ITokenCacheAccessor
...
-            return RefreshTokenCacheDictionary[refreshTokenKey];
+            return AccessTokenCacheDictionary[strKey];
         }
 
-        public void DeleteAccessToken(string cacheKey)
+        public string GetRefreshToken(MsalRefreshTokenCacheKey refreshTokenKey)
         {
-            AccessTokenCacheDictionary.Remove(cacheKey);
+            var strKey = refreshTokenKey.ToString();
+            if (!RefreshTokenCacheDictionary.ContainsKey(strKey))
```
bgavrilMS, line 84: This is not thread safe. You could have a thread be suspended immediately after line 84, another thread adding a refresh token with that key. You can use TryGetValue to at least minimize the number of calls to the dictionary, but it won't solve the multi-threading issue. See details https://stackoverflow.com/questions/16929231/is-trygetvalue-thread-safe-with-itself

It might be that having a race condition here is not a big problem, in which case you should add comments explaining this. #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
```diff
@@ -46,56 +47,90 @@ internal class TokenCacheAccessor : ITokenCacheAccessor
...
+            }
+
+            return RefreshTokenCacheDictionary[strKey];
         }
 
-        public void DeleteRefreshToken(string cacheKey)
+        public string GetIdToken(MsalIdTokenCacheKey idTokenKey)
         {
-            RefreshTokenCacheDictionary.Remove(cacheKey);
+            var strKey = idTokenKey.ToString();
```
bgavrilMS, line 94: There's quite a lot of repetetive code in these methods that can be eliminated with a privat method.

core/src/Features/Cache/TokenCacheAccessor.cs
```diff
@@ -29,6 +29,7 @@
 using System.Collections.ObjectModel;
 using System.Runtime.Serialization;
 using Microsoft.Identity.Core.Cache;
+using Microsoft.Identity.Core.Helpers;
 
 namespace Microsoft.Identity.Core
 {
```
bgavrilMS, line 35: This class needs documentation, especially the critical fact that it has to be thread safe.  #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
```diff
@@ -46,56 +47,90 @@ internal class TokenCacheAccessor : ITokenCacheAccessor
...
         }
 
-        public void DeleteAccount(string cacheKey)
+        public void DeleteAccessToken(MsalAccessTokenCacheKey cacheKey)
         {
-            AccountCacheDictionary.Remove(cacheKey);
+            AccessTokenCacheDictionary.Remove(cacheKey.ToString());
         }
 
+        public void DeleteRefreshToken(MsalRefreshTokenCacheKey cacheKey)
```
bgavrilMS, line 119: Also not thread safe and could end up throwing an exception if a second thread manages to remove the key. Use TryRemove if you want to avoid exceptions.  #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
```diff
@@ -123,7 +158,7 @@ public ICollection<string> GetAllAccountsAsString()
                 new ReadOnlyCollection<string>(
                    AccountCacheDictionary.Values.ToList());
         }
-
+        /*
```
bgavrilMS, line 161: Please delete dead code https://refactoring.guru/smells/dead-code

core/src/Features/Cache/TokenCacheAccessor.cs
```diff
@@ -151,58 +186,19 @@ public ICollection<string> GetAllAccountKeys()
...
-        public string GetAccount(string accountKey)
+        public void SetSecurityGroup(string securityGroup)
         {
-            if (!AccountCacheDictionary.ContainsKey(accountKey))
-            {
-                return null;
-            }
-
-            return AccountCacheDictionary[accountKey];
+            throw new System.NotImplementedException();
```
bgavrilMS, line 201: Please add a comment on why this is not implemented, e.g. supports a more advanced scenario that will be coded later etc.

core/src/Features/Cache/TokenCacheAccessor.cs
```diff
@@ -46,56 +47,90 @@ internal class TokenCacheAccessor : ITokenCacheAccessor
         internal readonly IDictionary<string, string> AccountCacheDictionary =
             new ConcurrentDictionary<string, string>();
 
-        public void SaveAccessToken(string cacheKey, string item)
+        public void SaveAccessToken(MsalAccessTokenCacheItem item)
```
bgavrilMS, line 50: Please initialise all internal dictionaries using an invariant culture EqualityComparer.   #Pending

core/src/Helpers/ExtensionMethods.cs
```diff
@@ -157,14 +157,14 @@ internal static string AsSingleString(this IEnumerable<string> input)
             return String.Join(" ", input);
         }
 
-        internal static SortedSet<string> AsSet(this string singleString)
+        internal static SortedSet<string> AsLowerCaseSortedSet(this string singleString)
```
bgavrilMS, line 160: Unlike Java, .net does not use the concept of "toLower" to force case insensitive comparisons. The reason is that not all cultures (think China, Japan etc.) have lowercase and uppercase letters, this more of a Latin / Cyrillic alphabet concept. 

In .net, you specify the culture, which can either be a specific culture specific or "ordinal", i.e. language agnostic (based on En language). You can also specify "ignore case"  -  https://docs.microsoft.com/en-us/dotnet/standard/base-types/comparing 

All data structures that involve equality (Sets, Dictionaries), have constructors that take in an IComparer, to which you can pass: 

```
new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```

core/src/Cache/CredentialType.cs
```diff
@@ -29,8 +29,8 @@ namespace Microsoft.Identity.Core.Cache
 {
     internal enum CredentialType
     {
-        IdToken,
-        AccessToken,
-        RefreshToken
+        idtoken,
```
bgavrilMS, line 32: Actually the .net coding style states that enum values should be uppercase https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions #Pending

core/src/Cache/ITokenCacheAccessor.cs
```diff
@@ -63,13 +63,16 @@ internal interface ITokenCacheAccessor
 
         ICollection<string> GetAllAccountsAsString();
 
+        /*
```
bgavrilMS, line 66: dead code #Pending

core/src/Cache/ITokenCacheAccessor.cs
```diff
@@ -31,29 +31,29 @@ namespace Microsoft.Identity.Core.Cache
 {
     internal interface ITokenCacheAccessor
```
bgavrilMS, line 32: Please add comments. If you add comments on the interface methods, you do not have to add them on the implemeting class. #Pending

core/src/Instance/AadAuthority.cs
```diff
@@ -57,11 +57,13 @@ public AadAuthority(string authority, bool validateAuthority) : base(authority,
         protected void UpdateCanonicalAuthority()
         {
             UriBuilder uriBuilder = new UriBuilder(CanonicalAuthority);
+            // will not be needed after authority aliasses will be added
```
bgavrilMS, line 60: dead code - remove or uncomment #Pending

core/src/Platforms/android/TokenCacheAccessor.cs
```diff
@@ -70,43 +71,52 @@ public TokenCacheAccessor(RequestContext requestContext) : this()
...
-            editor.PutString(cacheKey, item);
+            editor.PutString(item.GetKey().ToString(), JsonHelper.SerializeToJson(item));
             editor.Apply();
         }
 
-        public string GetRefreshToken(string refreshTokenKey)
+        public void SaveIdToken(MsalIdTokenCacheItem item)
         {
-            return _refreshTokenSharedPreference.GetString(refreshTokenKey, null);
+            ISharedPreferencesEditor editor = _idTokenSharedPreference.Edit();
```
bgavrilMS, line 90: Too much duplicated code... #Pending

core/src/Platforms/android/TokenCacheAccessor.cs
```diff
@@ -30,6 +30,7 @@
 using Microsoft.Identity.Client;
 using Microsoft.Identity.Core.Cache;
 using System.Collections.ObjectModel;
+using Microsoft.Identity.Core.Helpers;
 
 namespace Microsoft.Identity.Core
 {
```
bgavrilMS, line 36: Please add comemnts, especially if this is multi-threaded or not (I assume not?) #Pending

core/src/Platforms/android/TokenCacheAccessor.cs
```diff
@@ -134,7 +152,7 @@ public ICollection<string> GetAllAccountsAsString()
         {
             return _accountSharedPreference.All.Values.Cast<string>().ToList();
         }
-
+        /*
```
bgavrilMS, line 155: dead code. #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -32,90 +32,248 @@
...
 {
     internal class TokenCacheAccessor : ITokenCacheAccessor
     {
-        private const string AccessTokenServiceId = "com.microsoft.identity.client.accessToken";
-        private const string RefreshTokenServiceId = "com.microsoft.identity.client.refreshToken";
-        private const string IdTokenServiceId = "com.microsoft.identity.client.idToken";
-        private const string AccountServiceId = "com.microsoft.identity.client.Account";
+        public const string CacheKeyDelimiter = "-";
+
+        static Dictionary<string, int> AuthorityTypeToAttrType = new Dictionary<string, int>()
```
bgavrilMS, line 43: Please specify culture (invariant, case insenstive)

bgavrilMS: What are these magic numbers? #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -32,90 +32,248 @@
...
+        {
+            {AuthorityType.AAD.ToString(), 1001},
+            {AuthorityType.MSA.ToString(), 1002},
+            {AuthorityType.MSSTS.ToString(), 1003},
+            {AuthorityType.OTHER.ToString(), 1004},
+        };
+
+        enum CredentialAttrType
+        {
+            AccessToken = 2001,
```
bgavrilMS, line 53: What are these magic numbers? The compiler will generate values for you. #Pending

core/src/Cache/CacheFallbackOperations.cs
henrik-me: I believe this class has been considered a "factory" class or perhaps rather a helper class to create objects... I suppose at this point in time it's even misleading that the class is not marked as static. Clearly some technical debt to be cleaned. IMO, a work item should be created to track refactoring.
@Petro: What is the intent of this file containing these static methods? 

---
In reply to: [199527049](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199527049) [](ancestors = 199527049)

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -32,90 +32,248 @@
...
+            RefreshToken = 2002,
+            IdToken = 2003,
+            Password = 2004
+        }
 
         private const bool _defaultSyncSetting = false;
         private const SecAccessible _defaultAccessiblityPolicy = SecAccessible.AfterFirstUnlockThisDeviceOnly;
 
+        private readonly string DefaultKeychainGroup = "com.microsoft.adalcache";
```
bgavrilMS, line 62: I wonder if this could have been part of a separte PR. #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -32,90 +32,248 @@
...
+        private readonly string TeamIdKey = "teamIDHint";
```
bgavrilMS, line 63: const

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -238,29 +237,46 @@ internal class CacheFallbackOperations
...
+                        && rtCacheItem.ClientId.Equals(clientId))
                     {
-                        Result = new AdalResult(null, null, DateTimeOffset.MinValue),
-                        RefreshToken = rtCacheItem.Secret,
-                        RawClientInfo = rtCacheItem.RawClientInfo
-                    };
+                        foreach (MsalAccountCacheItem accountCacheItem in accounts)
+                        {
```
henrik-me: logic seems very complicated. I would like to see a comment explaining why we have to to this. #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -32,90 +32,248 @@
...
+
+        private string keychainGroup;
+
+        private string GetBundleId()
+        {
+            return NSBundle.MainBundle.BundleIdentifier;
+        }
+   
+        public void SetSecurityGroup(string securityGroup)
```
bgavrilMS, line 72: We don't enforce code sytle yet, but we should try to order the code elements as described in the MSFT code style: http://stylecop.soyuz5.com/SA1201.html namely: 

Fields
Constructors
Delegates
Events
Enums
Properties
public Methods
private methods



 #Pending

core/src/Cache/CredentialType.cs
```diff
@@ -29,8 +29,8 @@ namespace Microsoft.Identity.Core.Cache
...
+        accesstoken,
+        refreshtoken
```
henrik-me, line 34: Any reason these are not using CamelCasing ? #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
bgavrilMS, line 72: Not sure what this does, but I'd have 2 methods - one taking a string security group and throwing an ArgumentNull Exception if it is null or empty, and one taking 0 args and doing the GetBundleId() thing. Otherwise, it does too much. #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -32,90 +32,248 @@
...
+
+            SecRecord match = SecKeyChain.QueryAsRecord(queryRecord, out SecStatusCode resultCode);
+
+            if (resultCode == SecStatusCode.ItemNotFound)
+            {
+                SecKeyChain.Add(queryRecord);
+                match = SecKeyChain.QueryAsRecord(queryRecord, out resultCode);
+            }
+
+            return match.AccessGroup.Split('.')[0];
```
bgavrilMS, line 100: does this always work, i.e. returns a string with a dot inside? #Pending

henrik-me:
Do you need the resource.designer.cs file? IMO, it should be removed.

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -32,90 +32,248 @@
...
+            return GetValues((int)CredentialAttrType.IdToken);
         }
 
         public ICollection<string> GetAllAccountsAsString()
         {
-            return GetValues(AccountServiceId);
+            return GetValues(AuthorityTypeToAttrType[AuthorityType.MSSTS.ToString()]);
         }
-
+        /*
```
bgavrilMS, line 271: dead code #Pending

henrik-me (changes requested):
:clock1:

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -169,12 +329,13 @@ private ICollection<string> GetValues(string service)
 
             return res;
         }
-
+        /*
```
bgavrilMS, line 332: dead code #Pending

bgavrilMS (changes requested):
A few notes: 

1. There seem to be very few test changes relative to the amount of code added. We should strive to have at least 80% code coverage on new code (I'm looking at enforcing this through the build). 
2. The PR build fails with compilation errors

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: public methods should have docs.
I totally disagree that we should write comments to all methods.

1. Comments typically used as a crutch for bad code, easily readable code is the best "comments".
2. It is quite hard to write good comments, it is harder than writing good code. Most of time "comments" are not just useless, very often they are misleading or not synced with code. 
 #Pending

SomkaPe: It is helper class used to share SSO between adal and Msal sdk. 
i  did not get why having state would be better #Pending

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: added #Resolved

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: Count is O(1) operation, according to msdn - https://msdn.microsoft.com/en-us/library/27b47ht3(v=vs.110).aspx #WontFix

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: Resharper disagree with you, the type is obvious here #Pending

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: this is not relevant to this pr, in this code base Culture is not used for string comparison,
Please consider creating task #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
SomkaPe: already wrote my opinion about code documenting #Pending

core/src/Cache/CredentialType.cs
SomkaPe: this values are used in Msal schema, were it is lower case  #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
SomkaPe: yes we do not, set some meeting were code style enforcing might be discuss #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
SomkaPe: unifed cache is not finalized yet, code cleanup will be done later #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
SomkaPe: why? #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
SomkaPe: please clarify you question  #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
SomkaPe: all this is called from critical section with lock on cache state #Pending

core/src/Features/Cache/TokenCacheAccessor.cs
SomkaPe: not sure what you meant here, lets restrict scope of this pr To this pr  #Pending

core/src/Helpers/ExtensionMethods.cs
SomkaPe: this is not about comparison, this is about transforming date to expected format #ByDesign

core/src/Cache/CredentialType.cs
SomkaPe: this values are part of cache format schemas #Pending

core/src/Platforms/android/TokenCacheAccessor.cs
SomkaPe: please keep scope  of pr to changes in pr #Pending

core/src/Cache/CacheFallbackOperations.cs
jennyf19: I think this is for FxCop globalization checks in VSTS.  #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
SomkaPe: this magic number used by iOS Msal so if we want to be able to share cache with them, we have to use them

core/src/Platforms/iOS/TokenCacheAccessor.cs
SomkaPe: this is part of Existing Api to Msal iOS , implementation in .net just reflects it,
discussion of final API in .NET is still in progress. #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
SomkaPe: yes access group in iOS has particular format #Pending

core/src/Cache/CacheFallbackOperations.cs
bgavrilMS: Sure, I agree that good code is self documenting and I am not saying we should document every parameter etc. The fact that writing good comments is hard is proof that there is value in it. 

While you know all the technical decisions made around using MSAL + ADAL caches, others that will contribute after you (including your future self in 6 months+) do not know this. 

For example for the method above, it is not clear from the method signature what happens if the find operation fails - does it return a null? Does it throw? Does the wrapper encapsulate the possibility of not having a result?
 #Pending

bgavrilMS: Having too many parameters is bad for maintainability and readability. Many static analyzers have rules for this, including Java's Checkstyle and C#'s FxCop. Making this an instance class would be a way of dealing with this - but there are others o course (an intermediary object, ctor, builder pattern). #Pending

core/src/Cache/CredentialType.cs
bgavrilMS: Better to have a method that takes in a enum value and gives you the string - it is more intentional. #Pending

core/src/Cache/CacheFallbackOperations.cs
bgavrilMS: Sure, a list knows how many items it has. This is more of a general rule of thumb, because usually we work with IEnumerable and then you do not know what sort of collection sits behind IEnumerable - it could be a list (in which case Count in O(1)) or it could be a collection created on the fly with yield return (in which case Count is O(n)). So personally I prefer to always use ".Any()"

https://stackoverflow.com/questions/305092/which-method-performs-better-any-vs-count-0 #Pending

core/src/Cache/CacheFallbackOperations.cs
bgavrilMS: R# does not follow Microsoft standard coding practice, which states:

"Do not use var when the type is not apparent from the right side of the assignment."

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#implicitly-typed-local-variables #Pending

core/src/Cache/CacheFallbackOperations.cs
bgavrilMS: This will break the compliance build if you try to check it in.

core/src/Features/Cache/TokenCacheAccessor.cs
bgavrilMS: This is should be documented as it is a critical part of the system: 

- performance expectations
- thread safety expectations
- platform differences
 #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
bgavrilMS: By enforcing I meant having StyleCop run as part of the build and potentially breaking the build. Right now the enforcement comes through PRs :). Do you not agree with having a consistent way of ordering elements?

 #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
bgavrilMS: I may be misunderstanding the utility of the "dev" branch. My understanding is that dev branch should be stable because it is where we run tests and it acts as a de-facto release branch. As developers, we do not control the release schedule, and we could be asked to ship at any time. So having clean code in dev is pretty important. 

Maybe you should push all your changes to a feature branch and then merge everything with dev  when everything is complete? You can still send PRs based on this feature branch. #WontFix

core/src/Platforms/iOS/TokenCacheAccessor.cs
bgavrilMS: Sure, just saying this method does too much and it would make the code clearer if there 2 methods.  #Pending

core/src/Platforms/iOS/TokenCacheAccessor.cs
henrik-me: Context: For now we use dev as an all features branch for common core. adalv3dev is the real dev branch for adal. When cache and web views are done we will complete an FTP to validate the state of the entire branch (which we have never tested fully e2e since the refactoring's into core started). For this particular branch this approach is expected until we reach the above mentioned milestones. After that point it will switch to a true dev branch for our unified libraries. Hope this helps explain the reasoning behind. #ByDesign

core/src/Cache/CacheFallbackOperations.cs
henrik-me: I do not see the comments yet.

---
In reply to: [199976862](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199976862) [](ancestors = 199976862)

msal/tests/Test.MSAL.NET.Unit/CacheTests/TokenCacheTests.cs
```diff
@@ -57,7 +58,9 @@ public class TokenCacheTests
         [TestInitialize]
         public void TestInitialize()
         {
+            new TestPlatformInformation();
```
henrik-me, line 61: >  new TestPlatformInformation(); [](start = 10, length = 32)

are these two (platforminformation and testlogger) really needed for every test? would it be enough just to initialize for the entire test suite?

msal/tests/Test.MSAL.NET.Unit/CacheTests/TokenCacheTests.cs
```diff
@@ -252,20 +244,18 @@ public void GetAccessTokenExpiryInRangeTest()
...
-            atItem.CreateDerivedProperties();
+            var atItem = new MsalAccessTokenCacheItem(
+                TestConstants.ProductionEnvironment,
+                TestConstants.ClientId,
+                "Bearer",
+                TestConstants.Scope.AsSingleString(),
+                TestConstants.Utid,
+                "",
+                new DateTimeOffset(DateTime.UtcNow + TimeSpan.FromMinutes(4)),
+                MockHelpers.CreateClientInfo());
```
henrik-me, line 255: Seems like these are almost duplicated. What's the reason not to put this in a method or set of methods, that potentially re-uses the same base and just sets a different dateimeoffset on the at?
e.g.:
GetExpiredAT()
{ return GetAT(expiredTime) }
GetAT(DateTimeOffset()...

IMO: Would reveal the intend in the code and remove duplicates

henrik-me:
>             response.TokenType = "Bearer";

most of these are similar as well... could follow a similar model, imo (I know it's not your current code but would be nice to make the two very similar concepts reveal intent from the actual code)

---
Refers to: msal/tests/Test.MSAL.NET.Unit/CacheTests/TokenCacheTests.cs:552 in dae509d. [](commit_id = dae509dd0522e69bc7328c7db18e68b874de1e4f, deletion_comment = False)

henrik-me:
Do you need the resource.designer.cs file? IMO, it should be removed.: still ?

---
In reply to: [401862261](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#issuecomment-401862261) [](ancestors = 401862261)

core/src/Cache/CacheFallbackOperations.cs
henrik-me: this one is pending a follow up discussion on code style

---
In reply to: [199526233](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199526233) [](ancestors = 199526233)

henrik-me: pending discussion on code style

---
In reply to: [199551935](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199551935) [](ancestors = 199551935,199527049)

core/src/Cache/CacheFallbackOperations.cs
henrik-me: I suppose I still don't get it... would have preferred additional methods which could have helped clarified the intent of the code

---
In reply to: [200183263](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r200183263) [](ancestors = 200183263,199976862)

core/src/Cache/CacheFallbackOperations.cs
henrik-me: IMO should go into our discussion on code style. imo when it is a list then using count should be ok.

---
In reply to: [199527404](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199527404) [](ancestors = 199527404)

core/src/Cache/CacheFallbackOperations.cs
henrik-me: pending discussion on code style

---
In reply to: [199528185](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199528185) [](ancestors = 199528185)

core/src/Cache/CacheFallbackOperations.cs
henrik-me: pls. take in the code style discussion

---
In reply to: [199528285](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199528285) [](ancestors = 199528285)

core/src/Cache/CacheFallbackOperations.cs
henrik-me: pls. take in the code style discussion

---
In reply to: [199528488](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199528488) [](ancestors = 199528488)

core/src/Features/Cache/TokenCacheAccessor.cs
henrik-me: part of the code style conversation

---
In reply to: [199529765](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199529765) [](ancestors = 199529765)

core/src/Features/Cache/TokenCacheAccessor.cs
henrik-me: IMO: Meaningful class doc makes a lot of sense

---
In reply to: [199539634](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199539634) [](ancestors = 199539634)

core/src/Features/Cache/TokenCacheAccessor.cs
henrik-me: good point visibility is currently the same. pls. ensure this is discussed in the code style discussion.

---
In reply to: [199533054](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199533054) [](ancestors = 199533054)

core/src/Features/Cache/TokenCacheAccessor.cs
henrik-me: pls. take in the code style discussion

---
In reply to: [199534046](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199534046) [](ancestors = 199534046)

core/src/Features/Cache/TokenCacheAccessor.cs
henrik-me: pls. take in the code style discussion

---
In reply to: [199541077](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199541077) [](ancestors = 199541077)

core/src/Features/Cache/TokenCacheAccessor.cs
henrik-me: pls. take in the code style discussion

---
In reply to: [199534322](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199534322) [](ancestors = 199534322)

core/src/Features/Cache/TokenCacheAccessor.cs
henrik-me: IMO: could be addressed in this PR or in a follow up PR. But absolutely worth putting this in a method or similar.

---
In reply to: [199538424](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199538424) [](ancestors = 199538424)

core/src/Features/Cache/TokenCacheAccessor.cs
henrik-me: pls. take in the code style discussion. the intent and the lock is hidden away in code far away... not obvious, would be interesting to get perspectives on what we can do to make it more obvious.

---
In reply to: [199538253](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199538253) [](ancestors = 199538253)

core/src/Features/Cache/TokenCacheAccessor.cs
henrik-me: Agree. Source history will get it back easily if needed.

---
In reply to: [199540085](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199540085) [](ancestors = 199540085)

core/src/Features/Cache/TokenCacheAccessor.cs
henrik-me: At least if not implemented shouldn't it have been virtual allowing for iOS to override? Or perhaps it should only be in the iOS API?

---
In reply to: [199540409](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199540409) [](ancestors = 199540409)

core/src/Helpers/ExtensionMethods.cs
henrik-me: I'm not sure what you expect here Bogdan. I suppose if this string was Chinese it would just not do anything. (to my knowledge it would not ever be Chinese). Do you suggest to add a comparer? Or do you suggest another way to actually lower case the string before converting it into a set?

---
In reply to: [199543859](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199543859) [](ancestors = 199543859)

core/src/Cache/ITokenCacheAccessor.cs
henrik-me: pls. take in the code style discussion

---
In reply to: [199545504](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199545504) [](ancestors = 199545504)

core/src/Cache/ITokenCacheAccessor.cs
henrik-me: imo: remove, source history can easily bring those back. But essentially a discussion for code style.

---
In reply to: [199545302](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199545302) [](ancestors = 199545302)

core/src/Cache/CredentialType.cs
henrik-me: (agree) pls. take in the code style discussion

---
In reply to: [199545185](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199545185) [](ancestors = 199545185)

core/src/Instance/AadAuthority.cs
henrik-me: Petro, I believe this is the Authority Migration code. I suppose this will come in a follow up PR before completing cache work. I suggest leaving but ensuring you comment here that this is expected to complete right after this PR completes.

---
In reply to: [199546203](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199546203) [](ancestors = 199546203)

core/src/Platforms/android/TokenCacheAccessor.cs
henrik-me: see related comments. pls. take in the code style discussion

---
In reply to: [199549788](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199549788) [](ancestors = 199549788)

core/src/Platforms/android/TokenCacheAccessor.cs
henrik-me: see related comments

---
In reply to: [199550077](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199550077) [](ancestors = 199550077)

core/src/Platforms/android/TokenCacheAccessor.cs
henrik-me: pls. take in the code style discussion. IMO if the base interface has the right amount of description we might be good. But seems relevant with a few well crafted words

---
In reply to: [199550027](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199550027) [](ancestors = 199550027)

core/src/Platforms/iOS/TokenCacheAccessor.cs
henrik-me: +1

---
In reply to: [199552098](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199552098) [](ancestors = 199552098)

core/src/Platforms/iOS/TokenCacheAccessor.cs
henrik-me: is this needed here?

---
In reply to: [199550365](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199550365) [](ancestors = 199550365)

core/src/Platforms/iOS/TokenCacheAccessor.cs
henrik-me: Not sure I understand this comment

---
In reply to: [199552042](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199552042) [](ancestors = 199552042)

core/src/Platforms/iOS/TokenCacheAccessor.cs
henrik-me: any reason the numbers from this enum is related to the numbers above, if so any reason it's not using the enum.

Aside: code style: enum should imo be in a serarate file.

---
In reply to: [199980125](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199980125) [](ancestors = 199980125)

core/src/Platforms/iOS/TokenCacheAccessor.cs
henrik-me: pls. take in the code style discussion

---
In reply to: [200084285](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r200084285) [](ancestors = 200084285)

henrik-me: pls. take in the code style discussion

---
In reply to: [199553595](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199553595) [](ancestors = 199553595)

core/src/Cache/CacheFallbackOperations.cs
henrik-me: pls. take in the code style discussion

---
In reply to: [199973661](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199973661) [](ancestors = 199973661)

henrik-me: pls. take in the code style discussion

---
In reply to: [200073976](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r200073976) [](ancestors = 200073976)

core/src/Platforms/android/TokenCacheAccessor.cs
henrik-me: pls. take in the code style discussion

---
In reply to: [199979992](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1108#discussion_r199979992) [](ancestors = 199979992)

core/src/Cache/CacheFallbackOperations.cs
bgavrilMS: Not sure about this one because it will break the compliance build. You can either disable this particular rule (it's FxCop based) and check that all these string comparisons methods w/o culture info do the right thing. Or you could be explicit and add culture information, which is what Wilson code does for example.

core/src/Cache/CredentialType.cs
bgavrilMS: Not sure about this one. Are you saying that the token cache keys / value strings have these names in them? In that case, the code is brittle - I can come in and rename these enums and I'll break the cache because how would I know these are used as strings as well? 

Do we at least have unit tests that would break?

core/src/Helpers/ExtensionMethods.cs
bgavrilMS: I'm saying that using toLower() is a Java construct to perform case insensitive operations. In .net you initialize your Set with a string comparison and that's it. This has the following advantages: 

- it's fast (every call to ToLower() creates a new string because strings are immutable, whereas with the comparer there is no string duplication)
- it's concise (just initialize your structure with the comparer and that's it)
- it's safe (.net guidelines specify that you should always specify a string comparer explicity ... in my experience it's worth doing this - it will save a few ICMs with customers in Japan :) )

henrik-me: Thanks for the clarification on the IComparer 
In this case one new string will be made from which the set is created, however not multiple new strings, right?
The code is generically not multi language enabled and I believe the scenario here may not be subject to globalization challenges. IMO: something also worthwhile taking to our general discussion on code style.

core/src/Cache/CacheFallbackOperations.cs
henrik-me: I wonder how the build is green if the rules should stop it. Then we are possibly forcing the wrong build on the PR's

bgavrilMS: @henrik-me - the build is green because we do not run all compliance tasks on PR build. I have a task to move some of these tasks there, but note that some of the build errors are not actionable by external contributors because the tools are internal.

bgavrilMS (approved):

henrik-me:
>                 acquireResponseLabel.Text = String.Format("MsalException -\nError Code: {0}\nMessage: {1}", exception.ErrorCode, exception.Message);

technicality... but shouldn't this have been s ?

---
Refers to: msal/tests/dev apps/XForms/XForms/AcquirePage.xaml.cs:163 in f5e83a1. [](commit_id = f5e83a19c519ea0bbf2d5b9521bd87c334a5b311, deletion_comment = False)

henrik-me:
>                 acquireResponseLabel.Text = String.Format("MsalException -\nError Code: {0}\nMessage: {1}", exception.ErrorCode, exception.Message);

is cultureinfo missing here?

---
Refers to: msal/tests/dev apps/XForms/XForms/AcquirePage.xaml.cs:163 in f5e83a1. [](commit_id = f5e83a19c519ea0bbf2d5b9521bd87c334a5b311, deletion_comment = False)

henrik-me (approved):
:shipit:

PR #1114, opened 2018-07-07
Include AccessDenied key for telemetry event

jennyf19:
Include more info in exception result in xamarin test app

### DIFF
--- a/msal/src/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
+++ b/msal/src/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
@@ -114,6 +114,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
                         AuthenticationRequestParameters.RequestContext)
                         .ConfigureAwait(false);
                 uiEvent.UserCancelled = _authorizationResult.Status == AuthorizationStatus.UserCancel;
+                uiEvent.AccessDenied = _authorizationResult.Status == AuthorizationStatus.ProtocolError;
             }
             finally
             {
--- a/msal/tests/Test.MSAL.NET.Unit/RequestsTests/InteractiveRequestTests.cs
+++ b/msal/tests/Test.MSAL.NET.Unit/RequestsTests/InteractiveRequestTests.cs
@@ -211,6 +211,8 @@ namespace Test.MSAL.NET.Unit.RequestsTests
                 anEvent[EventBase.EventNameKey].EndsWith("ui_event") && anEvent[UiEvent.UserCancelledKey] == "false"));
             Assert.IsNotNull(_myReceiver.EventsReceived.Find(anEvent =>  // Expect finding such an event
                 anEvent[EventBase.EventNameKey].EndsWith("api_event") && anEvent[ApiEvent.UiBehaviorKey] == "select_account"));
+            Assert.IsNotNull(_myReceiver.EventsReceived.Find(anEvent =>  // Expect finding such an event
+                anEvent[EventBase.EventNameKey].EndsWith("ui_event") && anEvent[UiEvent.AccessDeniedKey] == "false"));
         }
 
         [TestMethod]
--- a/msal/tests/dev apps/XForms/XForms/AcquirePage.xaml.cs
+++ b/msal/tests/dev apps/XForms/XForms/AcquirePage.xaml.cs
@@ -160,7 +160,7 @@ namespace XForms
             }
             catch (MsalException exception)
             {
-                acquireResponseLabel.Text = "MsalException - " + exception.Message;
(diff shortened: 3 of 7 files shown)


msal/src/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
```diff
@@ -114,6 +114,7 @@ protected override string GetUIBehaviorPromptValue()
                         AuthenticationRequestParameters.RequestContext)
                         .ConfigureAwait(false);
                 uiEvent.UserCancelled = _authorizationResult.Status == AuthorizationStatus.UserCancel;
+                uiEvent.AccessDenied = _authorizationResult.Status == AuthorizationStatus.ProtocolError;
```
jennyf19, line 117: when I was testing embedded webview, w/corp account and getting the protocolError, it was getting registered as a UserCancelled, even though it was AccessDenied in the telemetry. I added in this line to register the protocolError.

msal/tests/Test.MSAL.NET.Unit/RequestsTests/InteractiveRequestTests.cs
```diff
@@ -211,6 +211,8 @@ public void NoCacheLookup()
                 anEvent[EventBase.EventNameKey].EndsWith("ui_event") && anEvent[UiEvent.UserCancelledKey] == "false"));
             Assert.IsNotNull(_myReceiver.EventsReceived.Find(anEvent =>  // Expect finding such an event
                 anEvent[EventBase.EventNameKey].EndsWith("api_event") && anEvent[ApiEvent.UiBehaviorKey] == "select_account"));
+            Assert.IsNotNull(_myReceiver.EventsReceived.Find(anEvent =>  // Expect finding such an event
+                anEvent[EventBase.EventNameKey].EndsWith("ui_event") && anEvent[UiEvent.AccessDeniedKey] == "false"));
```
jennyf19, line 215: Added `UiEvent.AccessDeniedKey` to preexisting unit tests

msal/tests/dev apps/XForms/XForms/AcquirePage.xaml.cs
```diff
@@ -160,7 +160,7 @@ private string GetSelectedUserId()
             }
             catch (MsalException exception)
             {
-                acquireResponseLabel.Text = "MsalException - " + exception.Message;
+                acquireResponseLabel.Text = String.Format("MsalException -\nError Code: {0}\nMessage: {1}", exception.ErrorCode, exception.Message);
```
jennyf19, line 163: The Xamarin app was just showing `MsalServiceException`, so now it returns a little more info. not all the exceptions have a message.

msal/src/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
henrik-me: Why call it access denied if it's a protocol error?

---
In reply to: [200799703](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1114#discussion_r200799703) [](ancestors = 200799703)

msal/tests/Test.MSAL.NET.Unit/RequestsTests/InteractiveRequestTests.cs
henrik-me: are there any real unit tests for this class? It should be relatively simple to validate.
also seems like this integration test only validated that it's false (not set). Ideally the integration testing would also show an error situation.

---
In reply to: [200799705](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1114#discussion_r200799705) [](ancestors = 200799705)

msal/src/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
jennyf19: The error that is returned is `access_denied`, and we equate it with an `AuthorizationStatus.ProtocolError` [here](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/core/src/UI/AuthorizationResult.cs#L114), but seems like `access_denied` is more specific, which is why I included that for the telemetry, but can change it to protocol error.

msal/tests/Test.MSAL.NET.Unit/RequestsTests/InteractiveRequestTests.cs
jennyf19: I added in more test coverage for the uiEvent I added. There are Telemetry tests in core.

henrik-me (approved):
:shipit:

PR #1117, opened 2018-07-10
Fix some of the build warnings

bgavrilMS:
This should fix all but 2 warnings: 

- Xamarin wanting to connect to a MAC 
- warning NU1603: runtime.native.System.IO.Compression 4.3.0

These are both external and I am tracking them

### DIFF
--- a/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/netstandard1.1/PlatformParameters.cs
+++ b/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/netstandard1.1/PlatformParameters.cs
@@ -30,9 +30,15 @@ using System;
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 {
+    /// <summary>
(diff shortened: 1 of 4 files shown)


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/netstandard1.1/PlatformParameters.cs
```diff
@@ -30,9 +30,15 @@
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 {
+    /// <summary>
+    ///  Additional parameters used in acquiring user's authorization, specific to netstadard 1.1
```
jennyf19, line 34: `netstandard 1.1` needs an "n"  here and in 1.3 file. Thanks for adding this (I should have done this when I did the webview changes)

jennyf19 (approved):
🚀

jennyf19:
@bgavrilMS Where are you tracking the work items for these? VSTS?

bgavrilMS:
Merged manually

PR #1123, opened 2018-07-11
Device Code retry fix and unit test

bgavrilMS:
One of our tests was flaky and it looks like there is a bug in the retry code (see comments inline)

### DIFF
--- a/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenByDeviceCodeHandler.cs
+++ b/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenByDeviceCodeHandler.cs
@@ -49,9 +49,8 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Flows
         {
             TimeSpan timeRemaining = deviceCodeResult.ExpiresOn - DateTimeOffset.UtcNow;
             AdalResultWrapper resultEx = null;
-            //the interval is added so that the while loop does not end before aquiring the last response from the STS stating that the code has expired.
-            //Without it, resultEx will end up being null.
-            while (timeRemaining.TotalSeconds + deviceCodeResult.Interval > 0)
+
+            while (timeRemaining.TotalSeconds > 0)
             {
                 try
                 {
@@ -70,6 +69,13 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Flows
                 timeRemaining = deviceCodeResult.ExpiresOn - DateTimeOffset.UtcNow;
             }
 
+            if (resultEx == null)
+            {
+                throw new AdalServiceException(
+                    AdalErrorEx.DeviceCodeAuthorizationCodeExpired, 
+                    AdalErrorMessageEx.DeviceCodeAuthorizationCodeExpired);
+            }
+           
             return resultEx;
         }
 
--- a/adal/tests/Test.ADAL.NET.Integration/DeviceProfileTests.cs
+++ b/adal/tests/Test.ADAL.NET.Integration/DeviceProfileTests.cs
@@ -186,13 +186,17 @@ namespace Test.ADAL.NET.Integration
         }
 
         [TestMethod]
-        public async Task NegativeDeviceCodeTimeoutTest()
+        public async Task NegativeDeviceCodeTimeoutTest_WithOneRetry()
         {
             MockHttpMessageHandler mockMessageHandler = new MockHttpMessageHandler(TestConstants.DefaultAuthorityHomeTenant)
             {
                 Method = HttpMethod.Get,
                 Url = TestConstants.DefaultAuthorityHomeTenant + "oauth2/devicecode",
-                ResponseMessage = MockHelpers.CreateSuccessDeviceCodeResponseMessage("1")
+                ResponseMessage = MockHelpers.CreateSuccessDeviceCodeResponseMessage(
+                    // do not lower this to 1-2s as test execution may be slow and the flow 
+                    // will never call the server
+                    expirationTimeInSeconds: 3, 
+                    retryInternvalInSeconds: 2)
             };
 
             AdalHttpMessageHandlerFactory.AddMockHandler(mockMessageHandler);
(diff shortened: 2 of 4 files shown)


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenByDeviceCodeHandler.cs
```diff
@@ -49,10 +50,13 @@ public AcquireTokenByDeviceCodeHandler(RequestData requestData, DeviceCodeResult
         {
             TimeSpan timeRemaining = deviceCodeResult.ExpiresOn - DateTimeOffset.UtcNow;
             AdalResultWrapper resultEx = null;
+
             //the interval is added so that the while loop does not end before aquiring the last response from the STS stating that the code has expired.
             //Without it, resultEx will end up being null.
-            while (timeRemaining.TotalSeconds + deviceCodeResult.Interval > 0)
```
bgavrilMS: This condition can fail, probably not in real scenarios where the timeouts are large, but it did fail in tests. The underlying issue is that we should contact the server at least once. Alternatively if we notice the timeout has already happened, we could timeout directly, but this requires more code and behaviour change.

SomkaPe: I have a few concerns with this approach. First, I would consider sending a request to service with expired code as a Service contract violation. Second, why would we try to use expired code when we already know it is expired? Also I see the idea of changing business logic because some unit test is failing as very dangerous. From my perspective, you can just increase timeout in test data - it is a unit test - we can mock any server response with any parameters. I think this would be a solution.

Also, why do you suggest we contact the server at least once? I would like to know more about your reasoning here.

rayluo: I would also agree with @SomkaPe that ideally we should not need to send out that request when we already know the code has expired. And, to be fair, @bgavrilMS did mention "alternatively if we notice the timeout has already happened, we could timeout directly", so he also saw this, hence not necessarily "suggest" us to "contact server at least once".

The underlying issue, imo, is not that we *should* contact the server at least once. It is that our pre-existing code happened to [already be written to expect a result available](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1123/files#diff-c04b555b31a145194bbca57706839214R205) and then we kinda had to [follow it](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1123/files#diff-ac38c278be1bc474043033ed72031a67R54). <sub>Yeah, it is one of those [JDD](https://programmingzen.com/jenga-driven-development/) moments.</sub>

In terms of the solution/workaround this time, given that this condition rarely happens in real life, can we actually consider simply raise an exception for the timeout, and catch it in the test case, and then call it a day?

bgavrilMS: +1 for the JDD reference :)

bgavrilMS:
I rebased the fix and forced pushed and it looks like GitHub deleted the comments. 

@rayluo @SomkaPe  Please have a look now.

adal/tests/Test.ADAL.NET.Integration/DeviceProfileTests.cs
```diff
@@ -231,5 +235,31 @@ public void NegativeDeviceCodeTest()
...
+            AdalHttpMessageHandlerFactory.AddMockHandler(mockMessageHandler);
+
+            TokenCache cache = new TokenCache();
+            context = new AuthenticationContext(TestConstants.DefaultAuthorityHomeTenant, cache);
+            DeviceCodeResult dcr = await context.AcquireDeviceCodeAsync("some resource", "some authority");
+
+            Assert.IsNotNull(dcr);
+            AuthenticationResult result;
+            AdalServiceException ex = AssertException.TaskThrows<AdalServiceException>(async () => result = await context.AcquireTokenByDeviceCodeAsync(dcr));
+            Assert.IsTrue(ex.Message.Contains("Verification code expired"));
```
rayluo: I know that historically this code base uses `Assert.IsTrue(ex.Message.Contains("blah blah blah"))` a lot. But, at least for newly added test cases, do we consider to switch to a different style as `Assert.AreEqual(AdalErrorEx.DeviceCodeAuthorizationCodeExpired, ex.ErrorCode)`?

Rationale: When error code and error message are both available, an unspoken assumption is that the messages are considered as documentation therefore could potentially be changed in future (in order to fix typo, or rewrite for a better explanation, etc.), but the error code would be considered as part of the contract and remains the same (unless the designer would introduce a breaking change).

adal/tests/Test.ADAL.NET.Integration/DeviceProfileTests.cs
```diff
@@ -186,13 +186,17 @@ public void NegativeDeviceCodeTest()
...
         {
             MockHttpMessageHandler mockMessageHandler = new MockHttpMessageHandler(TestConstants.DefaultAuthorityHomeTenant)
             {
                 Method = HttpMethod.Get,
                 Url = TestConstants.DefaultAuthorityHomeTenant + "oauth2/devicecode",
-                ResponseMessage = MockHelpers.CreateSuccessDeviceCodeResponseMessage("1")
+                ResponseMessage = MockHelpers.CreateSuccessDeviceCodeResponseMessage(
+                    // do not lower this to 1-2s as test execution may be slow and the flow 
+                    // will never call the server
+                    expirationTimeInSeconds: 3,
```
rayluo: If we use a much larger value here, say, `expirationTimeInSeconds: 30`, this test case would still pass, wouldn't it? If so, I would suggest to use a larger value here. Because, if value `1` or `2` would introduce those intermittent test failure, then `3` seems still too close. The purpose of this PR is to eleminate those failures, so why not do it in a more reliable way?

rayluo (commented):
Thanks @bgavrilMS , like your new approach a lot! :-)  Just a couple minor comments below.

By the way, I also use rebase a lot. Github would not delete our previous comments, it just marks them as "outdated" and hides them, because you already changed those commented areas. I actually like this behavior and use it as my to-do list to focus on the remaining comments on my PRs.

adal/tests/Test.ADAL.NET.Integration/DeviceProfileTests.cs
bgavrilMS: Yes, great point!

rayluo (approved):
LGTM. Thanks @bgavrilMS !

bgavrilMS:
Merged manually.

PR #1124, opened 2018-07-11
Trwalke/subject issuer auth for MSAL

trwalke:
Implementation of subject name issuer auth feature in ADAL/MSAL. The main goal of this feature is for the SDKs to send an "x5c" claim to the STS which is the url encoded public certificate. 

https://identitydivision.visualstudio.com/DevEx/_workitems/edit/487409


adal/tests/Test.ADAL.NET.Common/TestConstants.cs
```diff
@@ -58,6 +58,7 @@ public class TestConstants
         public static readonly string UserRealmEndPoint = "userrealm";
         public static readonly string DiscoveryEndPoint = "discovery/instance";
         public static readonly string DefaultRefreshTokenValue = "RefreshTokenValue";
+        public static readonly string DefaultAuthorizationCode = "AwABAAAAvPM1KaPlrEqdFSBzjqfTGBCmLdgfSTLEMPGYuNHSUYBrqqf_ZT_p5uEAEJJ_nZ3UmphWygRNy2C3jJ239gV_DBnZ2syeg95Ki-374WHUP-i3yIhv5i-7KU2CEoPXwURQp6IVYMw-DjAOzn7C3JCu5wpngXmbZKtJdWmiBzHpcO2aICJPu1KvJrDLDP20chJBXzVYJtkfjviLNNW7l7Y3ydcHDsBRKZc3GuMQanmcghXPyoDg41g8XbwPudVh7uCmUponBQpIhbuffFP_tbV8SNzsPoFz9CLpBCZagJVXeqWoYMPe2dSsPiLO9Alf_YIe5zpi-zY4C3aLw5g9at35eZTfNd0gBRpR5ojkMIcZZ6IgAA";
```
henrik-me: >DefaultAuthorizationCode [](start = 38, length = 24)

shouldn't this be in keyvault? #Resolved

henrik-me:
Did you consider the refactoring's needed to move the functionality to Core? #Closed

henrik-me:
You added a resource.designer.cs file, pls. remove it. #Closed

henrik-me (changes requested):
:clock1:

adal/tests/Test.ADAL.NET.Common/TestConstants.cs
trwalke: it is a dummy code that I randomly generated and it is just for testing the api. It is never sent to any endpoints as it is intercepted by the mock handlers. #Closed

henrik-me: Thanks for the clarification.
It may though be though of by a tool as an important key... not entirely sure how to reveal that intent, however perhaps if you add that to the name of the test constant (e.g. DummyAuthorizationCode) #Closed

henrik-me:
> 

nit: you have this extra line.... #Closed

---
Refers to: adal/devApps/XFormsApp.iOS/XFormsApp.iOS.csproj:159 in a5e7fe6. [](commit_id = a5e7fe6c0a1d407d254c5c58b1a0bcf8096ae923, deletion_comment = False)

henrik-me:
lgtm - do we have any e2e tests or manual test cases for this?  #Closed

henrik-me (approved):
:shipit:

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -642,6 +642,29 @@ internal IWebUI CreateWebAuthenticationDialog(PlatformParameters parameters)
...
+        /// <param name="resource">Identifier of the target resource that is the recipient of the requested token. It can be null if provided earlier to acquire authorizationCode.</param>
+        /// <param name="sendX5c">This parameter enables application developers to achieve easy certificates roll-over
+        /// in Azure AD: setting this parameter to true will send the public certificate to Azure AD
+        /// along with the token request, so that Azure AD can use it to validate the subject name based on a trusted issuer policy.
+        /// This saves the application admin from the need to explicitly manage the certificate rollover
+        /// (either via portal or powershell/CLI operation)</param>
+        /// <returns>It contains Access Token, its expiration time, user information.</returns>
+        public async Task<AuthenticationResult> AcquireTokenByAuthorizationCodeAsync(string authorizationCode,
+            Uri redirectUri, IClientAssertionCertificate clientCertificate, string resource, bool sendX5c)
+        {
```
SomkaPe, line 683: As I understand the point of this api is to not send cert value in the client assertion, It is not cleat how it will "enables application developers to achieve easy certificates roll-over" #Resolved

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
```diff
@@ -91,10 +91,10 @@ public JsonWebToken(IClientAssertionCertificate certificate, string audience)
             };
         }
 
-        public ClientAssertion Sign(IClientAssertionCertificate credential)
+        public ClientAssertion Sign(IClientAssertionCertificate credential, bool sendX5C)
```
SomkaPe: Why this parameter called sendX5C ? 
this Method create ClientAssertion object from this (JsonWebToken) object, there is no such thing as send (this class should not know about how it will be used in outer context) #Resolved

SomkaPe: also X5C is not precise enough - is it thumbprint or cert value or both #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
```diff
@@ -115,9 +115,9 @@ private static string UrlEncodeSegment(byte[] segment)
             return Base64UrlEncoder.Encode(segment);
         }
 
-        private static string EncodeHeaderToJson(IClientAssertionCertificate credential)
+        private static string EncodeHeaderToJson(IClientAssertionCertificate credential, bool sendX5C)
```
SomkaPe: sendX5C name does not reflect usage of parameter #Resolved

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
```diff
@@ -128,10 +128,10 @@ private static long ConvertToTimeT(DateTime time)
             return (long)(diff.TotalSeconds);
         }
 
-        private string Encode(IClientAssertionCertificate credential)
+        private string Encode(IClientAssertionCertificate credential, bool sendX5C)
```
SomkaPe: sendX5C name does not reflect usage of parameter #Resolved

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
```diff
@@ -209,18 +209,19 @@ internal class JWTPayload
         [DataContract]
         internal sealed class JWTHeaderWithCertificate : JWTHeader
         {
-            private string _thumbPrint;
-
-            public JWTHeaderWithCertificate(IClientAssertionCertificate credential)
+            public JWTHeaderWithCertificate(IClientAssertionCertificate credential, bool sendX5C)
```
SomkaPe: sendX5C name does not reflect usage of parameter #Resolved

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
```diff
@@ -209,18 +209,19 @@ internal class JWTPayload
...
                 : base(credential)
             {
-                _thumbPrint = this.Credential.Thumbprint;
+                X509CertificateThumbprint = this.Credential.Thumbprint;
                 X509CertificatePublicCertValue = null;
 
+                if (!sendX5C)
+                    return;
+
```
SomkaPe: are you sure that you want to return here ?
As I understand the idea of next  check was to force usage of  library provided implementation of IClientAssertionCertificate  #Closed

SomkaPe: looks like with that change it will be possible to use developer provided impl of IClientAssertionCertificate ,  it is api change  #ByDesign

SomkaPe (dismissed):

core/tests/Test.Microsoft.Identity.Core.Unit/Mocks/MockHttpMessageHandler.cs
```diff
@@ -54,6 +54,8 @@ internal class MockHttpMessageHandler : HttpMessageHandler
 
         public Exception ExceptionToThrow { get; set; }
 
+        public Action<HttpRequestMessage> AdditionalRequestValidation { get; set; }
```
SomkaPe, line 56: why do you need this ? you can validate QueryParams and PostData without this #Resolved

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -642,6 +642,29 @@ internal IWebUI CreateWebAuthenticationDialog(PlatformParameters parameters)
                     new ClientKey(clientAssertion), resource).ConfigureAwait(false);
         }
 
+#if !(ANDROID || iOS || WINDOWS_APP)
+        /// <summary>
+        /// Acquires security token from the authority using an authorization code previously received.
```
jmprieur: I would change the first sentence to
**In a Web App, ** attemps to acquire a security token from the authority using an authorization code previously received (after a call to one of the overrides of `<see cref="M:AcquireTokenByAuthorizationCodeAsync">AcquireTokenByAuthorizationCodeAsync</see>`). For more details see https://aka.ms/adal-net-authorization-code
 #Resolved

jmprieur: @trwalke @henrik-me @SomkaPe @jennyf19 @bgavrilMS : note that I want to generalize the aka.ms linking to the conceptual documentation. for the moment they link to the wiki, but in the future, they might redirect to docs.ms #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -714,6 +737,28 @@ internal IWebUI CreateWebAuthenticationDialog(PlatformParameters parameters)
                     userAssertion).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Acquires an access token from the authority on behalf of a user. It requires using a user token previously received.
```
jmprieur, line 762: Please add:
"for more details, see https://aka.ms/adal-net-on-behalf-of" #Resolved

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -764,6 +809,27 @@ internal IWebUI CreateWebAuthenticationDialog(PlatformParameters parameters)
                 .ConfigureAwait(false);
         }
 
+#if !(ANDROID || iOS || WINDOWS_APP)
+        /// <summary>
+        /// Acquires a security token from the authority while enabling simplified Azure AD certificate roll-over.
```
jmprieur: Please change to:
Acquire a security token for the application (without a user) from the authority while enabling simplified Azure AD certificate roll over. For more details, see https://aka.ms/adal-net-client-credentials #Resolved

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
```diff
@@ -150,7 +150,42 @@ public sealed class ConfidentialClientApplication : ClientApplicationBase, IConf
         {
             return
                 await
-                    AcquireTokenForClientCommonAsync(scopes, forceRefresh, ApiEvent.ApiIds.AcquireTokenForClientWithScopeRefresh).ConfigureAwait(false);
+                    AcquireTokenForClientCommonAsync(scopes, forceRefresh, ApiEvent.ApiIds.AcquireTokenForClientWithScopeRefresh, false).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Acquires token from the service for the confidential client. This method attempts to look up valid access token in the cache.
```
jmprieur: Please change to:

Acquires token from the service for the confidential client using the client credentials flow. See https://aka.ms/msal-net-client-credentials
 #Resolved

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
```diff
@@ -150,7 +150,42 @@ public sealed class ConfidentialClientApplication : ClientApplicationBase, IConf
...
+        /// </summary>
+        /// <param name="scopes">Array of scopes requested for resource</param>
+        /// <param name="sendCert">This parameter enables application developers to achieve easy certificates roll-over
+        /// in Azure AD: setting this parameter to true will send the public certificate to Azure AD
+        /// along with the token request, so that Azure AD can use it to validate the subject name based on a trusted issuer policy.
+        /// This saves the application admin from the need to explicitly manage the certificate rollover
+        /// (either via portal or powershell/CLI operation)</param>
+        /// <returns>Authentication result containing application token for the requested scopes</returns>
+        public async Task<AuthenticationResult> AcquireTokenForClientAsync(bool sendCert, IEnumerable<string> scopes)
```
jmprieur: why is sendCert the first parameter?  #Resolved

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
```diff
@@ -150,7 +150,42 @@ public sealed class ConfidentialClientApplication : ClientApplicationBase, IConf
...
+        {
+            return
+                await
+                    AcquireTokenForClientCommonAsync(scopes, false, ApiEvent.ApiIds.AcquireTokenForClientWithScope, sendCert).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Acquires token from the service for the confidential client. This method attempts to look up valid access token in the cache.
```
jmprieur: Please change to:
Acquires token from the service for the confidential client using the client credentials flow. See https://aka.ms/msal-net-client-credentials

Also, I think the second sentence is wrong: "This method attempts to look up valid access token in the cache" 
it should probably be:
"This method attempts to look up valid access token in the cache unless `<paramref name="forceRefresh"/>` is <c>true</c>"
 #Resolved

jmprieur (changes requested):

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
jmprieur: @SomkaPe proposing to solve that through the aka.ms links. Let's discuss at standup. #Resolved

jmprieur (changes requested):
I've proposed a few changes to the XML doc strings in order to improve them, and start the work on providing a rich reference documentation (linking to the conceptual documentation).

I propose we add improving the XML comments in the definition of done,

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -642,6 +642,29 @@ internal IWebUI CreateWebAuthenticationDialog(PlatformParameters parameters)
...
+#if !(ANDROID || iOS || WINDOWS_APP)
+        /// <summary>
+        /// Acquires security token from the authority using an authorization code previously received.
+        /// This method does not lookup token cache, but stores the result in it, so it can be looked up using other methods such as <see cref="AuthenticationContext.AcquireTokenSilentAsync(string, string, UserIdentifier)"/>.
+        /// </summary>
+        /// <param name="authorizationCode">The authorization code received from service authorization endpoint.</param>
+        /// <param name="redirectUri">The redirect address used for obtaining authorization code.</param>
+        /// <param name="clientCertificate">The client certificate to use for token acquisition.</param>
+        /// <param name="resource">Identifier of the target resource that is the recipient of the requested token. It can be null if provided earlier to acquire authorizationCode.</param>
+        /// <param name="sendX5c">This parameter enables application developers to achieve easy certificates roll-over
```
jmprieur, line 675: sendCertificate instead of sendX5c (everywhere). This is a breaking change in ADAL.NET for existing methods, but since we are going to release a preview of a new major version, I guess this is ok
(breaking change as customers could have used the sendX5c : true syntax) #Resolved

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
henrik-me: Currently no functionality is exposed, however eventually the functionality will be available via the Portal. Would be good to add aka.ms links. Also can be explained here that the cert to be rolled over is supplied through different mechanisms. This cert is just used to validate that the cert on the server side and client side are good for the scenario (the server will extract the relevant values from the cert which is used to validate the server side cert). #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
henrik-me: @trwalke thoughts on this one? #Closed

core/tests/Test.Microsoft.Identity.Core.Unit/Mocks/MockHttpMessageHandler.cs
henrik-me: @trwalke thoughts? #Resolved

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -624,8 +624,9 @@ internal IWebUI CreateWebAuthenticationDialog(PlatformParameters parameters)
         }
 
         /// <summary>
-        /// Acquires security token from the authority using an authorization code previously received.
-        /// This method does not lookup token cache, but stores the result in it, so it can be looked up using other methods such as <see cref="AuthenticationContext.AcquireTokenSilentAsync(string, string, UserIdentifier)"/>.
+        /// **In a Web App, ** attemps to acquire a security token from the authority using an authorization code previously received
```
jmprieur: The ** were for the PR comment, @trwalke they won't be good in the XML comment :)
otherwise this looks good :) #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
jmprieur: (resolved) #Resolved

jmprieur (approved):
Thanks for taking my suggestions, @trwalke 
We are good for the XML comments

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
trwalke: oh I see, Ill fix #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
trwalke: Resolved #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
trwalke: (resolved) #Resolved

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
trwalke: (resolved) #Closed

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
trwalke: (resolved) #Closed

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
trwalke: @jmprieur Visual studio showed an error when I put sendCertificate behind the scopes parameter because there is already an overload that takes form of AcquireTokenForClientAsync(IEnumerable<string>, bool).

Thus there would be ambiguity between:
AcquireTokenForClientAsync(IEnumerable<string> scopes, bool **sendCertificate**)
and
AcquireTokenForClientAsync(IEnumerable<string> scopes, bool **forceRefresh**) #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
henrik-me: @trwalke

---
In reply to: [202845504](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r202845504) [](ancestors = 202845504)

henrik-me (changes requested):
:clock1:

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
henrik-me: @trwalke

---
In reply to: [202846673](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r202846673) [](ancestors = 202846673)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
henrik-me: @trwalke

---
In reply to: [202846635](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r202846635) [](ancestors = 202846635)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
henrik-me: @trwalke

---
In reply to: [202846565](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r202846565) [](ancestors = 202846565)

core/tests/Test.Microsoft.Identity.Core.Unit/Mocks/MockHttpMessageHandler.cs
henrik-me: @trwalke

---
In reply to: [202848908](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r202848908) [](ancestors = 202848908)

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
henrik-me: @jmprieur
Perhaps it should be AquireTokenWithCertificateAsync ? At least to me it will be more intuitive... than having to remember the order of params. Then we will not need to have the bool in the api.

---
In reply to: [206953484](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r206953484) [](ancestors = 206953484)

adal/tests/Test.ADAL.NET.Common/TestConstants.cs
henrik-me: Can you randomly generate this one each run as part of the test code?

---
In reply to: [201777626](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r201777626) [](ancestors = 201777626)

henrik-me:
Still a few outstanding comments in the PR. #Closed

adal/tests/Test.ADAL.NET.Common/TestConstants.cs
trwalke: Yes, we can do that #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
trwalke: Updated parameter to "sendCertificate" #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
trwalke: Updated parameter to "sendCertificate" #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
trwalke: Updated parameter to "sendCertificate" #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
trwalke: Updated parameter to "sendCertificate" #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
trwalke: yes, we return here because if there is no request to send the x5c, there is no reason to check for developer implementation or set the X509CertificatePublicCertValue. The only reason we are checking for the developer implementation is because we cannot send the x5c if the developer implemented their own IClientAssertionCertificate . #Closed

core/tests/Test.Microsoft.Identity.Core.Unit/Mocks/MockHttpMessageHandler.cs
trwalke: This is how ADAL is designed to check QueryParams and PostData currently. I wanted the two to be the same so that we can write tests using the same methods that we do in ADAL. How else can we verify QueryParams and PostData in unit tests? #Resolved

trwalke: Followed up offline, currently, MSAL can check if the value exists in the headers but it cannot check if the parameter does not exist. The AdditionalRequestValidation will enable many more validation scenarios. (resolved) #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ClientCreds/JsonWebToken.cs
henrik-me: the parameters used here are still called sendX5C adaiks

---
In reply to: [207037978](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r207037978) [](ancestors = 207037978)

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/JsonWebToken.cs
```diff
@@ -86,10 +87,10 @@ public JsonWebToken(string clientId, string audience)
             };
         }
 
-        public string Sign(ClientAssertionCertificate credential)
+        public string Sign(ClientAssertionCertificate credential, bool sendX5C)
```
henrik-me: >sendX5C [](start = 71, length = 7)

as Petro suggested. stick to one name all the way through. Ideally the one agreed to in the public API with the PM #Resolved

msal/src/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -66,8 +66,11 @@ internal class AuthenticationRequestParameters
         public UserAssertion UserAssertion { get; set; }
 
         public bool IsClientCredentialRequest { get; set; } = false;
+
         public string SliceParameters { get; set; }
 
+        public bool SendX5c { get; set; }
```
henrik-me: >SendX5c [](start = 20, length = 7)

... #Closed

msal/tests/Test.MSAL.NET.Unit/Test.MSAL.NET.Unit.csproj
```diff
@@ -84,5 +85,10 @@
     <None Update="Resources\valid_cert.cer">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Update="valid_cert.pfx">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
   </ItemGroup>
+<Target Name="ProjectReferenceSolution" BeforeTargets="ResolveAssemblyReferences" DependsOnTargets="ResolveProjectReferences">
```
henrik-me: :) #ByDesign

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
henrik-me: Seems like there are still **'s here and there.

---
In reply to: [206764640](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r206764640) [](ancestors = 206764640)

trwalke: #Resolved #Closed

henrik-me:
very minor... but the formatting you are using are adding empty lines to the end of the files... others do not have that. I recommend removing it and work with Bogdan to align on the formatting and share that with the rest of the team. (see e.g. your 3 files with just this change) #Closed

adal/devApps/AdalCoreCLRTestApp/Program.cs
```diff
@@ -75,4 +75,4 @@ private static X509Certificate2 GetCertificateByThumbprint(string thumbprint)
             }
         }
     }
-}
\ No newline at end of file
+}
```
henrik-me: revert #Closed

adal/devApps/XFormsApp/XFormsApp.csproj
```diff
@@ -30,4 +30,4 @@
       <Name>Microsoft.IdentityModel.Clients.ActiveDirectory</Name>
     </ProjectReference>
   </ItemGroup> 
-</Project>
\ No newline at end of file
+</Project>
```
henrik-me: revert #Closed

adal/devApps/XFormsApp.iOS/XFormsApp.iOS.csproj
```diff
@@ -156,4 +156,4 @@
     <Reference Include="Xamarin.iOS" />
   </ItemGroup>
   <Import Project="$(MSBuildExtensionsPath)\Xamarin\iOS\Xamarin.iOS.CSharp.targets" />
-</Project>
\ No newline at end of file
+</Project>
```
henrik-me: revert #Closed

adal/tests/Test.ADAL.NET.Unit/Test.ADAL.NET.Unit.csproj
```diff
@@ -58,6 +58,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.7.2" />
     <PackageReference Include="MSTest.TestAdapter" Version="1.3.1" />
     <PackageReference Include="MSTest.TestFramework" Version="1.3.1" />
+    <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="5.1.5" />
```
henrik-me: ><PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="5.1.5" /> [](start = 4, length = 78)

remember to register the usage. Also I believe the latest version is 5.2.4 #Closed

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
henrik-me: @trawalke pls. follow up with Jean-Marc on this.

---
In reply to: [206999250](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r206999250) [](ancestors = 206999250,206953484)

henrik-me:
only major outstanding item is the public api. imo it's confusing for developers but not sure what I propose is better...
also: the version of System.IdentityModel is not the latest and greatest #Closed

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
henrik-me: I just discussed with JM. We should rename these. Will stop by.

---
In reply to: [207337998](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r207337998) [](ancestors = 207337998,206999250,206953484)

adal/devApps/XFormsApp/XFormsApp.csproj
trwalke: #Resolved #Closed

adal/devApps/XFormsApp.iOS/XFormsApp.iOS.csproj
trwalke: #Resolved #Closed

adal/tests/Test.ADAL.NET.Unit/Test.ADAL.NET.Unit.csproj
trwalke: #Resolved #Closed

adal/devApps/AdalCoreCLRTestApp/Program.cs
trwalke: #Resolved #Closed

adal/tests/Test.ADAL.NET.UIAutomation/XamarinDroidADALTests.cs
```diff
@@ -39,6 +39,7 @@ public void InitializeBeforeTest()
         /// Runs through the standard acquire token flow
         /// </summary>
         [Test]
+        [Ignore]
```
henrik-me: >[Ignore] [](start = 8, length = 8)

? #Closed

adal/tests/Test.ADAL.NET.Unit/Test.ADAL.NET.Unit.csproj
henrik-me: I do not see this updated to the latest version

---
In reply to: [207337639](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r207337639) [](ancestors = 207337639)

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
```diff
@@ -38,7 +38,7 @@ namespace Microsoft.Identity.Client
     /// <summary>
     /// Class to be used for confidential client applications like Web Apps/API.
     /// </summary>
-    public sealed class ConfidentialClientApplication : ClientApplicationBase, IConfidentialClientApplication
+    public sealed class ConfidentialClientApplication : ClientApplicationBase, IConfidentialClientApplication, IConfidentialClientApplicationWithCertificate
```
henrik-me, line 42: >IConfidentialClientApplicationWithCertificate [](start = 111, length = 45)

It seems like the obo flow is missing for MSAL, or am I missing something? #Closed

henrik-me:
OBO flow for MSAL?

---
In reply to: [410034878](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#issuecomment-410034878) [](ancestors = 410034878)

henrik-me:
I assume that we do not have any manual test cases for these scenarios? Could we get some?

---
In reply to: [410309081](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#issuecomment-410309081) [](ancestors = 410309081,410034878)

henrik-me:
5 remaining comments

---
In reply to: [410309397](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#issuecomment-410309397) [](ancestors = 410309397,410309081,410034878)

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
jmprieur: I agree. It's in ADAL. #Closed

trwalke: #Resolved #Closed

adal/tests/Test.ADAL.NET.UIAutomation/XamarinDroidADALTests.cs
trwalke: reverted
#Resolved #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -87,29 +87,14 @@ internal class AuthenticationRequestParameters
...
                 {
-                    if (ClientCredential.Assertion == null || ClientCredential.ValidTo != 0)
+                    if (ClientCredential.Certificate != null)
                     {
-                        bool assertionNearExpiry = (ClientCredential.ValidTo <=
-                                                    Jwt.JsonWebToken.ConvertToTimeT(DateTime.UtcNow +
-                                                                                    TimeSpan.FromMinutes(
-                                                                                    Constants 
-                                                                                    .ExpirationMarginInMinutes)));
-                        if (assertionNearExpiry)
```
trwalke, line 94: @jmprieur Can you think of any reasons as to why we would want to cache a signed client assertion for 10 minutes instead of resigning it for each authentication request in MSAL? ADAL does not have this caching #Closed

jmprieur: @trwalke : no. Do we understand how the code went to MSAL.NET ? I don't think that we tested the OBO flow yet (as it's only recently been available on the v2.0 endpoint)

#Resolved #Closed

jmprieur: maybe we should ask @SomkaPe as well?

#Resolved #Closed

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
```diff
@@ -118,6 +118,38 @@ public async Task<AuthenticationResult> AcquireTokenOnBehalfOfAsync(IEnumerable<
                         .ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Acquires token using On-Behalf-Of flow.
```
jmprieur: Can we please add (See https://aka.ms/msal-net-on-behalf-of)
 #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
trwalke: @jmprieur I checked an old PR (https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/191) and it did not mention anything specific about this. I also talked with Jason who reviewed the PR and he did not remember anything about it as well. So at this point we are not sure. However,  I see no value in it as it will only be useful when there are a significant amount of authentication requests in a short period of time which I do not believe to be a very common use case for the client credential flow.  #Closed

henrik-me:
the utests are failing #Closed

adal/tests/Test.ADAL.NET.Unit/Test.ADAL.NET.Unit.csproj
henrik-me: ? still not updated to the latest version

---
In reply to: [207600270](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r207600270) [](ancestors = 207600270,207337639)

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
```diff
@@ -114,7 +114,39 @@ public async Task<AuthenticationResult> AcquireTokenOnBehalfOfAsync(IEnumerable<
             Authority authorityInstance = Core.Instance.Authority.CreateAuthority(authority, ValidateAuthority);
             return
                 await
-                    AcquireTokenOnBehalfCommonAsync(authorityInstance, scopes, userAssertion, ApiEvent.ApiIds.AcquireTokenOnBehalfOfWithScopeUserAuthority)
+                    AcquireTokenOnBehalfCommonAsync(authorityInstance, scopes, userAssertion, ApiEvent.ApiIds.AcquireTokenOnBehalfOfWithScopeUserAuthority, false)
+                        .ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Acquires token using On-Behalf-Of flow. (See https://aka.ms/msal-net-on-behalf-of)
```
henrik-me, line 123: >(See https://aka.ms/msal-net-on-behalf-of) [](start = 52, length = 42)

pls. add the same link to the other obo flow methods. btw. add relevant links everywhere here. not just on the added methods. #Closed

henrik-me:
>         /// Acquires token using On-Behalf-Of flow.

for obo flow the aka.ms link to obo would go after this. #Closed

---
Refers to: msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs:105 in 622b2e4. [](commit_id = 622b2e48ed41d297be3dc254953e41edbad348a1, deletion_comment = False)

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ClientCredential.cs
```diff
@@ -47,6 +47,8 @@ public ClientCredential(ClientAssertionCertificate certificate)
         internal ClientAssertionCertificate Certificate { get; private set; }
         internal string Assertion { get; set; }
         internal long ValidTo { get; set; }
+        internal bool CachedX5C { get; set; }
```
henrik-me: >CachedX5C [](start = 22, length = 9)

why is this caching introduced? (based on the most recent discussion I thought we wouldn't need it but I may have missed something) #Closed

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/IConfidentialClientApplicationWithCertificate.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// This saves the application admin from the need to explicitly manage the certificate rollover
+        /// (either via portal or powershell/CLI operation)
+        /// </summary>
+        /// <param name="scopes">Array of scopes requested for resource</param>
+        /// <param name="forceRefresh">If TRUE, API will ignore the access token in the cache and attempt to acquire new access token using client credentials</param>
+        /// <returns>Authentication result containing application token for the requested scopes</returns>
+        Task<AuthenticationResult> AcquireTokenForClientWithCertificateAsync(IEnumerable<string> scopes, bool forceRefresh);
+
+        /// <summary>
+        /// Acquires token using On-Behalf-Of flow.
```
henrik-me: >Acquires token using On-Behalf-Of flow. [](start = 12, length = 39)

you can add the aka.ms link here as well as below #Closed

trwalke: #Resolved #Closed

adal/tests/Test.ADAL.NET.Unit/Test.ADAL.NET.Unit.csproj
henrik-me: Please update to latest System.IdentityModel ...

---
In reply to: [208421892](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r208421892) [](ancestors = 208421892,207600270,207337639)

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/IConfidentialClientApplicationWithCertificate.cs
henrik-me: link? same below btw.

---
In reply to: [208423187](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1124#discussion_r208423187) [](ancestors = 208423187)

msal/src/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -84,22 +87,20 @@ internal class AuthenticationRequestParameters
...
-                    if (ClientCredential.Assertion == null || ClientCredential.ValidTo != 0)
+                    if (ClientCredential.Assertion == null || ClientCredential.ValidTo != 0 || ClientCredential.ContainsX5C != SendCertificate)
                     {
-                        bool assertionNearExpiry = (ClientCredential.ValidTo <=
-                                                    Jwt.JsonWebToken.ConvertToTimeT(DateTime.UtcNow +
-                                                                                    TimeSpan.FromMinutes(
-                                                                                    Constants 
-                                                                                    .ExpirationMarginInMinutes)));
-                        if (assertionNearExpiry)
+                        bool assertionExpired = !this.Equals(ClientCredential);
```
henrik-me: >!this.Equals(ClientCredential); [](start = 48, length = 31)

You will need a utest to validate that even two different objects with the same assertions will result in being equal. The test may already be there, thus if you can point me to it that would be great. #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -117,6 +118,30 @@ internal class AuthenticationRequestParameters
...
+        }
+
+        public override int GetHashCode()
+        {
+            int hash = 13;
+            hash *= hash + 31 + Authority.SelfSignedJwtAudience.GetHashCode();
+            hash *= hash + 31 + SendCertificate.GetHashCode();
+            return hash;
+        }
+#endif
```
henrik-me: this is not right. It's not the AuthenticationRequestParameter we should be looking at. It's the assertions we need to validate for equality. #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -84,22 +87,20 @@ internal class AuthenticationRequestParameters
...
                 else
                 {
-                    if (ClientCredential.Assertion == null || ClientCredential.ValidTo != 0)
+                    if (ClientCredential.Assertion == null || ClientCredential.ValidTo != 0 || ClientCredential.ContainsX5C != SendCertificate)
                     {
-                        bool assertionNearExpiry = (ClientCredential.ValidTo <=
-                                                    Jwt.JsonWebToken.ConvertToTimeT(DateTime.UtcNow +
-                                                                                    TimeSpan.FromMinutes(
-                                                                                    Constants 
-                                                                                    .ExpirationMarginInMinutes)));
```
henrik-me, line 93: Seems from the code that we should also adhere to the ValidTo. This may be an external contract. The code should though be moved to the ClientCredentials class, to ensure separation of concern is maintained. 

You could add a bool to that class which is not serialized, but which returns whether the Assertion has expired. e.g. HasExpired #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -117,6 +118,30 @@ internal class AuthenticationRequestParameters
             return parameters;
         }
 
+#if DESKTOP || NETSTANDARD1_3
```
henrik-me: Why is this needed? #Closed

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ClientCredential.cs
```diff
@@ -45,8 +45,11 @@ public ClientCredential(ClientAssertionCertificate certificate)
         }
 
         internal ClientAssertionCertificate Certificate { get; private set; }
-        internal string Assertion { get; set; }
+        internal string Assertion { get;
+            set; }
```
henrik-me: nit: remove new line #Closed

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ClientCredential.cs
```diff
@@ -45,8 +45,11 @@ public ClientCredential(ClientAssertionCertificate certificate)
...
         internal long ValidTo { get; set; }
```
henrik-me: >ValidTo [](start = 22, length = 7)

Given this is completely internal now... however we might want to keep the expired functionality which was there before. #Closed

henrik-me:
>     public sealed class ClientCredential

>     public sealed class ClientCredential

move all the HasExpired/IsExpired logic to this class. as discussed you can do that my overloading equals. It should be done here and not in the AuthenticationRequestParameter.

and then add a utest that validates the behavior

#Resolved
---
Refers to: msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ClientCredential.cs:36 in 2b3ece1. [](commit_id = 2b3ece1034852a4490bc87dbe20248befdf34311, deletion_comment = False) #Closed

---
Refers to: msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ClientCredential.cs:36 in 2b3ece1. [](commit_id = 2b3ece1034852a4490bc87dbe20248befdf34311, deletion_comment = False)

msal/src/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
trwalke: #closed #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
trwalke: #Resolved #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
trwalke: #Resolved #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
trwalke: The client credential object is not available for the android/iOS platform so the compiler throws an error if it is not there.
#Resolved #Closed

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ClientCredential.cs
trwalke: #Resolved #Closed

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ClientCredential.cs
trwalke: #Resolved #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestValidationHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+            //The clientCredential object contains the previously used values in the cached client assertion string
+            var clientCredential = clientAssertionParameters.ClientCredential;
+            var expired = (clientCredential.ValidTo <=
+                                        Jwt.JsonWebToken.ConvertToTimeT(DateTime.UtcNow +
+                                                                        TimeSpan.FromMinutes(
+                                                                        Constants
+                                                                        .ExpirationMarginInMinutes)));
+
+            return !expired && clientCredential.Audience == clientAssertionParameters.Authority.SelfSignedJwtAudience
+                && clientCredential.ContainsX5C == clientAssertionParameters.SendCertificate;
```
henrik-me: I recommend putting this on a separate line with a meaningful variable name which can help reveal the intent. #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestValidationHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+using Microsoft.Identity.Core;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Internal.Requests
+{
+    internal static class RequestValidationHelper
```
henrik-me, line 37: >RequestValidationHelper [](start = 26, length = 23)

pls. ensure to add unit tests for this class. #Closed

adal/tests/Test.ADAL.NET.Unit/Test.ADAL.NET.Unit.csproj
trwalke: #Resolved #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestValidationHelper.cs
trwalke: #Resolved #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestValidationHelper.cs
trwalke: #Resolved #Closed

henrik-me (approved):
:shipit:

PR #1126, opened 2018-07-11
clear cache in both formats

SomkaPe:


adal/tests/Test.ADAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -88,16 +86,11 @@ public void Initialize()
...
-          //  legacyCachePersistance.LoadCache().Returns();
-
-
-
             Assert.IsTrue(adalTokenCache.TokenCacheAccessor.GetAllAccessTokensAsString().Count == 0);
             Assert.IsTrue(adalTokenCache.TokenCacheAccessor.GetAllRefreshTokensAsString().Count > 0);
 
             // clear Adal Cache
-            adalTokenCache.Clear();
+            adalTokenCache.ClearAdalCache();
```
jennyf19: we might need to add more test coverage for the different clearing scenarios.

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -145,6 +135,40 @@ internal class CacheFallbackOperations
...
+                {
+                    dictionary.Remove(key);
+                }
+
+                legacyCachePersistance.WriteCache(AdalCacheOperations.Serialize(dictionary));
+            }
+            catch (Exception ex)
+            {
+                CoreLoggerBase.Default.Warning("RemoveAdalUser falid due to Exception - " + ex);
+            }
```
jennyf19, line 168: spelling error: "failed"

Pii vs/ noPii logging needed here?

adal/devApps/XFormsApp/SecondPage.cs
```diff
@@ -63,9 +63,9 @@ public SecondPage()
                 Text = "Conditional Access"
             };
 
-            var clearButton = new Button
```
jennyf19: In the MSAL Xamarin Dev app, we have all three options (`ClearAdalCache`, `ClearMsalCache`, `ClearAllCache`)...but in adal, just `ClearAllCache` ...making a note of this here as something to add later for testing.

henrik-me:
>         internal TokenCacheAccessor TokenCacheAccessor = new TokenCacheAccessor();

this confused me while reviewing this PR, and I had to go and figure out if it was a static instance or what it was... can this be renamed? (start with lower case will make it more readable to me)

---
Refers to: adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs:78 in 84cb358. [](commit_id = 84cb358c4837cf28b2115ac89ff570c1345b03d3, deletion_comment = False)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -261,18 +261,29 @@ public virtual void Clear()
...
+                this.tokenCacheDictionary.Count));
+            this.tokenCacheDictionary.Clear();
+            CoreLoggerBase.Default.Info("Successfully Cleared Cache");
+            this.HasStateChanged = true;
+            this.OnAfterAccess(args);
+        }
+
+        internal void ClearMsalCache()
+        {
+            TokenCacheAccessor.Clear();
```
henrik-me: >TokenCacheAccessor [](start = 12, length = 18)

see above

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -145,6 +135,40 @@ internal class CacheFallbackOperations
...
+                foreach (AdalTokenCacheKey key in keysToRemove)
+                {
+                    dictionary.Remove(key);
+                }
+
+                legacyCachePersistance.WriteCache(AdalCacheOperations.Serialize(dictionary));
+            }
+            catch (Exception ex)
+            {
+                CoreLoggerBase.Default.Warning("RemoveAdalUser falid due to Exception - " + ex);
```
henrik-me, line 167: >ex [](start = 92, length = 2)

perhaps it should be more specific? Do we need the stack trace or is the error message enough?
Aside: @jennyf19 one thing to keep in mind for the error/exception handling work

henrik-me:
>         internal readonly TelemetryTokenCacheAccessor TokenCacheAccessor = new TelemetryTokenCacheAccessor();

pls. see wish from adal cache impl. regarding naming

---
Refers to: msal/src/Microsoft.Identity.Client/TokenCache.cs:50 in 84cb358. [](commit_id = 84cb358c4837cf28b2115ac89ff570c1345b03d3, deletion_comment = False)

henrik-me:
>             Assert.IsNotNull(result1);

You are validating that a result is retrieved, would it make sense to also validate the actual token in the cache? or at least the before and after count in the msal cache?

---
Refers to: msal/tests/Test.MSAL.NET.Unit/UnifiedCacheTests.cs:141 in 84cb358. [](commit_id = 84cb358c4837cf28b2115ac89ff570c1345b03d3, deletion_comment = False)

core/src/Cache/CacheFallbackOperations.cs
henrik-me: good point on the pii logging given the ex is used (which will contain the stack trace which could potentially contain pii afaik).

---
In reply to: [201867787](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1126#discussion_r201867787) [](ancestors = 201867787)

jennyf19:
>         internal void ClearCache()

two things:
1 - in adal, this is public - should it be public here as well?
2 - rename to Clear() to match adal
3 - can this be moved to core? seems to be doing the same thing in both libraries, so should be in core

---
Refers to: msal/src/Microsoft.Identity.Client/TokenCache.cs:958 in 84cb358. [](commit_id = 84cb358c4837cf28b2115ac89ff570c1345b03d3, deletion_comment = False)

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -945,26 +959,39 @@ internal void ClearCache()
...
             {
-                try
+                ClearMsalCache();
+                ClearAdalCache();
+            }
+        }
+
+        internal void ClearAdalCache()
+        {
+            IDictionary<AdalTokenCacheKey, AdalResultWrapper> dictionary = AdalCacheOperations.Deserialize(legacyCachePersistance.LoadCache());
```
jennyf19, line 969: In adal version there are log messages here...can you add them here for msal (or move to core instead).

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -945,26 +959,39 @@ internal void ClearCache()
         {
```
jennyf19, line 959: three things:
1 - in adal, this method `ClearCache()` is public
2 - rename to `Clear()` to match adal
3 - can this be moved to core? seems to be doing the same thing in both libraries

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: looks like all logging in core should be reviewed - all usage of CoreLoggerBase, it is used incorrectly - it is called only as pii or not pii - should be both, exception message is logged with non pii log methods

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
SomkaPe: ok

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: logging should be reviewed completely in core

core/src/Cache/CacheFallbackOperations.cs
jennyf19: @SomkaPe sounds good - I'll work on this later and then we can do a pass over core, adal, and msal
cc: @henrik-me

msal/src/Microsoft.Identity.Client/TokenCache.cs
SomkaPe: 1. it is not public in msal. 3. no TokeCache is implementations are different in adal an msal

jennyf19 (approved):
:cat:

PR #1127, opened 2018-07-12
Remove all Resource.Designer.cs files once and for all

rayluo:
Those Resources.Designer.cs files in our code base keep changing and mess up with our commit history. They don't seem to be necessary, according to [this post on Xamarin forum](https://forums.xamarin.com/discussion/65403/can-the-resource-designer-cs-files-be-ignored-in-source-control).

And, oh boy, it feels so good to delete 20032 lines of code in our code base and (presumably) still get paid. :-)


henrik-me (approved):
:shipit:

bgavrilMS (approved):
LGTM

PR #1132, opened 2018-07-13
address build warnings after async analyzer change and fix for Tenant in cache

jennyf19:


core/src/Cache/MsalAccountCacheItem.cs
```diff
@@ -58,7 +58,7 @@ internal MsalAccountCacheItem(Authority authority, MsalTokenResponse response) :
             Environment = environment;
             PreferredUsername = preferredUsername;
             Name = name;
-            TenantId = TenantId;
+            TenantId = tenantId;
```
henrik-me, line 61: >tenantI [](start = 23, length = 7)

:)

msal/samples/desktop/SampleApp/MainForm.cs
```diff
@@ -58,19 +53,21 @@ public MainForm()
             }
         }
 
-        private async void TabControl1_SelectedIndexChanged(object sender, EventArgs e)
+        // Return void because this is an event handler
+        private async void TabControl1_SelectedIndexChangedAsync(object sender, EventArgs e)
```
henrik-me: >private async void TabControl1_SelectedIndexChangedAsync(object sender, EventArgs e) [](start = 8, length = 84)

these are event ui handlers and should probably just have been added to the ignore list. It's like the Main method in a console app. (doc for async will state something similar using better words)

henrik-me: (same issue in a few places, but most of the places are spot on, thanks a ton for doing this!)

---
In reply to: [202498640](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1132#discussion_r202498640) [](ancestors = 202498640)

msal/src/Microsoft.Identity.Client/UIParent.cs
```diff
@@ -50,6 +50,9 @@ public UIParent()
         }
 
 #if iOS
+        /// <summary>
+        /// Constructor for iOS for implementing embedded webview
+        /// </summary>
```
henrik-me, line 55: not sure if you should add a remark (a real thing as part of the xml doc)/ saying this method is likely to be removed before final release

msal/samples/desktop/SampleApp/MainForm.cs
henrik-me: btw.... now you have done it I recommend leaving it as is.

---
In reply to: [202498676](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1132#discussion_r202498676) [](ancestors = 202498676,202498640)

henrik-me (approved):
:shipit:

msal/src/Microsoft.Identity.Client/UIParent.cs
jennyf19: done.

msal/samples/desktop/SampleApp/MainForm.cs
jennyf19: i fixed all the event handlers, but you'll have to show me how to get it to ignore those methods so we can remove `async` in them too. they don't show up in MSBuild, just in VS as warnings.

PR #1136, opened 2018-07-18
Globalization fix with interpolated string

bgavrilMS:


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/WindowsFormsWebAuthenticationDialogBase.cs
```diff
@@ -278,7 +278,11 @@ private string GetUrlFromDocument(Uri url, HtmlDocument document)
                     string value = elem.GetAttribute("value");
                     if (!string.IsNullOrEmpty(value))
                     {
-                        parameters.Add($"{name}={elem.GetAttribute("value")}");
```
bgavrilMS, line 281: @jennyf19 Since FxCop does not look at source code but at the assembly, it thinks that at line 281 there is a string.format (well, in fact there is one). - string interpolation is just syntactic sugar ... $"{name}=.. is string.Format("{0}", name);

jmprieur (approved):
LGTM

Issue #1135, opened 2018-07-18
Making an AT expire would cause an exception


rayluo:
This might not happen in real world, but it can be reproduced by using the [DesktopTestApp and click the Expire-Access-Token button](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/27bf9fa0a12a1d7ab7e8305d32688014ab757721/msal/tests/dev%20apps/DesktopTestApp/MsalUserAccessTokenControl.cs#L39), which will then trigger a NullReferenceException at [this line](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/27bf9fa0a12a1d7ab7e8305d32688014ab757721/msal/src/Microsoft.Identity.Client/TokenCache.cs#L1009).

jmprieur:
No customer impact. This is a more a refactoring

PR #1139, opened 2018-07-18
Fix a NullReferenceException when Test App makes an AT expire

rayluo:
This issue will not happen in real world, because the method is [only used in dev test app](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/compare/test-app-expire-button?expand=1#diff-13683a785bd5cf0439d6d937ca21369aR999). Nonetheless, it needs to be fixed to make [the test app usage](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/325d4c45903b001b26267c8c24d8bfc5aaf1ea99/msal/tests/dev%20apps/DesktopTestApp/MsalUserAccessTokenControl.cs#L39) work.

This PR fixes issue #1135


jmprieur:
you might want to update the title (fix a null reference occurring when AT expire?)

SomkaPe:
:shipit:

PR #1140, opened 2018-07-19
Broker updates for Adal Android

jennyf19:
While testing w/broker, I noticed that if the redirectUri is not correct, we don't throw an exception. The intent is returned as null and no error message is returned. @SomkaPe and I discussed this and decided it was best to throw early since the authentication cannot continue w/an incorrect redirect uri. I added an Android specific error message, and in the logs, we share what the correct uri should be for that application. This is a better user experience because now, the application receives an `AdalException` with a clear error message, compared with nothing before.

Platform specific code here, so not sure how to add unit test coverage, but I did test w/Android device OnePlus.


adal/devApps/XFormsApp/SecondPage.cs
```diff
@@ -37,6 +37,13 @@ public class SecondPage : ContentPage
     {
         private readonly StringBuilder _logs = new StringBuilder();
 
+        static string ClientId = "de49ddaf-c7f8-4a06-8463-3c6ae124fe52";
+
+        //static string AndroidBrokerRedirectURI = "msauth://com.microsoft.xformsdroid.adal/mJaAVvdXtcXy369xPWv2C7mV674=";
```
jennyf19: this is the redirectUri for broker for this application.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerHelper.cs
```diff
@@ -210,6 +210,11 @@ public void AcquireToken(IDictionary<string, string> brokerPayload)
                         RequestContext.Logger.ErrorPii(e);
                     }
                 }
+                else
```
jennyf19: we were not doing anything with a null intent here. Maybe this exception is too specific? Could just be "null intent" as an error message, but one needs to check ADAL logs to know the redirect uri was incorrect. thoughts?

adal/devApps/XFormsApp/SecondPage.cs
henrik-me: any reason this is not either left there or simply deleted? Ideally it should be kept and a button in the ui should be able to switch between broker and non-broker. thoughts?

---
In reply to: [203584400](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1140#discussion_r203584400) [](ancestors = 203584400)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerHelper.cs
henrik-me: Precise exceptions are good for supportability!

Aside: Seems to follow our existing ADAL pattern. One thing though to consider changing for MSAL exceptions. Looking forward to the discussions on exceptions for MSAL later.

---
In reply to: [203584606](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1140#discussion_r203584606) [](ancestors = 203584606)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerHelper.cs
```diff
@@ -210,6 +210,11 @@ public void AcquireToken(IDictionary<string, string> brokerPayload)
...
+                {
+                    throw new AdalException(AdalErrorAndroidEx.IncorrectBrokerRedirectUri, "The redirectUri is incorrect. The redirectUri should " +
+                        "be in this format: msauth://packagename/Base64UrlencodedSignature -> Check the ADAL logs for more details. ");
```
henrik-me: >Check the ADAL logs for more details. [](start = 94, length = 37)

what will actually be found in the logs?

henrik-me: >  [](start = 131, length = 1)

empty...

henrik-me:
approving PR expecting pointers raised to be addressed

henrik-me (approved):
:shipit:

henrik-me:
is something similar needed for the MSAL test app?

bgavrilMS (commented):
LGTM when dealt with Henrik's comments

adal/devApps/XFormsApp/SecondPage.cs
jennyf19: That's a good idea to include a button. I will do that.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerHelper.cs
jennyf19: agree.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerHelper.cs
jennyf19: I made a mistake...I added in logging here, which I didn't push up, but the [computed redirectUri is determined here](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs#L445). We could add logging in here, or i'll just change the above AdalException message. I thought we could actually log what the redirectUri should be, or we can just give the format. What do you think?

jennyf19: not sure what you mean

jennyf19:
we don't have broker support for MSAL (yet), but when we do, we should do something similar.

---
In reply to: [406158718](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1140#issuecomment-406158718) [](ancestors = 406158718)

adal/devApps/XFormsApp/SecondPage.cs
```diff
@@ -202,7 +208,54 @@ void LogCallback(LogLevel level, string message, bool containsPii)
                     this.result.Text += "Logs : " + DrainLogs();
                 });
             }
+        }
 
+        private async void AcquireTokenWithBrokerButton_Clicked(object sender, EventArgs e)
```
bgavrilMS: I believe Travis enabled some static analysis related to async. One of the rules is that all Async methods should be suffixed with Async.

adal/devApps/XFormsApp/SecondPage.cs
```diff
@@ -202,7 +208,54 @@ void LogCallback(LogLevel level, string message, bool containsPii)
...
+        private string DeterminePlatformForRedirectUri()
+        {
+            switch (Device.RuntimePlatform)
+            {
+                case Device.iOS:
+                    RedirectURI = IOSBrokerRedirectURI;
+                    break;
+                case Device.Android:
+                    RedirectURI = AndroidBrokerRedirectURI;
+                    break;
```
bgavrilMS, line 291: Nit: I like to add a default: throw new NotImplementedException just in case ... fail fast if another runtime platform is used.

bgavrilMS (approved):
LGTM, minor comments inline

adal/devApps/XFormsApp/SecondPage.cs
henrik-me: Perspective:
The rule is not meaningful for top level event handlers and can be ignored imo. This is the entry point for the async call. However either ignore and do not apply any async stuff to it or leave it with everything that the async rules says.

bgavrilMS: Yes, agreed, especially as VS generates those even handlers from the designer. Irrespective of the solution chosen we should try not to introduce new warnings into the build. 

The number of warnings has dropped a lot in the past few weeks and once it reaches 0 we can enable "treat warnings as errors" and start looking at more static analysis.

adal/devApps/XFormsApp/SecondPage.cs
jennyf19: agree...sounds good. thanks

adal/devApps/XFormsApp/SecondPage.cs
jennyf19: I agree. I will leave this as-is. Thanks.

adal/devApps/XFormsApp/SecondPage.cs
jennyf19: done.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerHelper.cs
jennyf19: i fixed this.

jennyf19: still not sure what you mean here.

adal/devApps/XFormsApp/SecondPage.cs
jennyf19: done.

jennyf19:
@henrik-me @bgavrilMS 
While testing, silent flow w/Broker was not working, so I made additional changes. I have only tested w/Android w/Company Portal on a device. Both interactive and silent flows work.

TODO: iOS broker flows

adal/devApps/XFormsApp.Droid/SecondPageRenderer.cs
```diff
@@ -59,8 +49,14 @@ protected override void OnElementChanged(ElementChangedEventArgs<Page> e)
...
-            
-            page.Parameters = new PlatformParameters(activity);
+
+            new PlatformParameters(activity);
+
+            #region Broker flow
+            // Uncomment below for broker flow
+            //var parameters = new PlatformParameters(activity, true);
+            //page.Parameters = parameters;
+            #endregion
```
jennyf19: Not sure how to get around having to uncomment this to test broker flow. This is platform specific, so iOS has it's own implementation.

adal/devApps/XFormsApp/SecondPage.cs
```diff
@@ -37,10 +37,11 @@ public class SecondPage : ContentPage
...
 
-        static string ClientId = "ebe49c8f-61de-4357-9194-7a786f6402b4";
-        static string AndroidBrokerRedirectURI = "msauth://com.microsoft.xformsdroid.adal/mJaAVvdXtcXy369xPWv2C7mV674=";
-        static string IOSBrokerRedirectURI = "adaliosapp://com.yourcompany.xformsapp";
-        string RedirectURI = "urn:ietf:wg:oauth:2.0:oob";
+        public const string ClientId = "<ClientId>";
+        public const string ClientIdBroker = "<ClientIdBroker>";
+        public const string AndroidBrokerRedirectURI = "msauth://com.microsoft.xformsdroid.adal/mJaAVvdXtcXy369xPWv2C7mV674=";
+        public const string IOSBrokerRedirectURI = "adaliosapp://com.yourcompany.xformsapp";
+        static string RedirectURI = "urn:ietf:wg:oauth:2.0:oob";
```
jennyf19: Using variables so the code is cleaner and easier to change if we need to. For using broker, the broker specific redirect uri needs to be registered in the app. I had to make my own clientId for testing this, but once we have one that we can use, we'll just register this redirectUri there.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -525,12 +525,18 @@ internal AdalResultWrapper LoadFromCacheCommon(CacheQueryData cacheQueryData, Re
 
                 IdToken idToken = IdToken.Parse(result.Result.IdToken);
 
+                string objectId = null;
+                if (idToken != null)
+                {
+                    objectId = idToken.ObjectId;
+                }
```
jennyf19: I was getting a null reference exception because the `IdToken` was null, which meant no `objectId` to pass through to `CacheFallBackOperations.WriteMsalRefreshToken(...)`. So i made a few changes to checkthis ahead of time and then just pass in null for `objectId`. Maybe @SomkaPe can confirm if it's okay to have a null `objectId` in the `MsalAccountCacheItem`.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -582,20 +588,20 @@ internal AdalResultWrapper LoadFromCacheCommon(CacheQueryData cacheQueryData, Re
...
-                            items.Where(p => p.Value.IsMultipleResourceRefreshToken).ToList();
-
-                        if (mrrtItems.Any())
                         {
-                            returnValue = mrrtItems.First();
-                            msg =
-                                "A Multi Resource Refresh Token for a different resource was found which can be used";
-                            requestContext.Logger.Info(msg);
-                            requestContext.Logger.InfoPii(msg);
+                            // There are no resource specific tokens.  Choose any of the MRRT tokens if there are any.
```
jennyf19, line 584: I think `control + kd` changed the formatting on this.

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -38,7 +38,7 @@ internal class CacheFallbackOperations
     {
         public static void WriteMsalRefreshToken(ITokenCacheAccessor tokenCacheAccessor,
             AdalResultWrapper resultWrapper, string authority, string clientId, string displayableId,
-            string identityProvider, string givenName, string familyName, string objectId)
+             string givenName, string familyName, string objectId)
```
jennyf19, line 41: `string identityProvider` wasn't being used anywhere in this method, so I just removed it. Doesn't make sense to pass in a parameter that is not used and `WriteMsalRefreshToken()` is only called here. We also don't have unit tests for this, but I need to learn more about the new cache format to write them.

adal/devApps/XFormsApp/SecondPage.cs
```diff
@@ -29,23 +29,30 @@
...
 using Xamarin.Forms;
-using Xamarin.Forms.PlatformConfiguration;
 
 namespace XFormsApp
 {
     public class SecondPage : ContentPage
     {
         private readonly StringBuilder _logs = new StringBuilder();
 
+        public const string ClientId = "d3590ed6-52b3-4102-aeff-aad2292ab01c";
```
henrik-me, line 39: >d3590ed6-52b3-4102-aeff-aad2292ab01c [](start = 40, length = 36)

did you intent to keep this id here?

adal/devApps/XFormsApp.Droid/SecondPageRenderer.cs
henrik-me: Suggestion: add a button to the ui. for each platform do one or the other depending on the value of the current ui.

---
In reply to: [204958541](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1140#discussion_r204958541) [](ancestors = 204958541)

adal/devApps/XFormsApp/SecondPage.cs
jennyf19: yes.

henrik-me: :shipit:

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
SomkaPe: Was it client credential grant case?

jennyf19: no...IDToken is not present in the current broker model

henrik-me: is a dummy id needed in this case? @SomkaPe what will this do for the cache?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerHelper.cs
SomkaPe: Yes this exception is to specific here, it might be any exception - not necessary related to incorrect URL,

---
In reply to: [203584606](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1140#discussion_r203584606) [](ancestors = 203584606)

SomkaPe: Exception should be thrown in the place were it is generated

henrik-me: does broker throw exceptions or allow us in any way to get the details of the error condition?

---
In reply to: [205907511](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1140#discussion_r205907511) [](ancestors = 205907511,203584606)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
jennyf19: @henrik-me we got it figured out. it is okay to pass null here.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerHelper.cs
jennyf19: @SomkaPe Thanks for the feedback. I was concerned about that. will make the changes you suggested. 🐱

jennyf19: done.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
jennyf19: @SomkaPe - took your suggestion of doing this `idtoken?.ObjectId` instead of all the other logic. thanks.
And I re-tested interactive and silent flows w/broker w/and w/out a correct redirectUri and things work as expected.
cc: @henrik-me

adal/devApps/XFormsApp.Droid/SecondPageRenderer.cs
jennyf19: went w/an idea from @SomkaPe

adal/devApps/XFormsApp/SecondPage.cs
```diff
@@ -195,21 +251,62 @@ void LogCallback(LogLevel level, string message, bool containsPii)
...
             string claims = "{\"access_token\":{\"polids\":{\"essential\":true,\"values\":[\"5ce770ea-8690-4747-aa73-c5b3cd509cd4\"]}}}";
-            
+
             try
             {
-                AuthenticationResult result = await ctx.AcquireTokenAsync("https://graph.windows.net", "<CLIENT_ID>",
-                        new Uri("adaliosapp://com.yourcompany.xformsapp"),
-                        Parameters, new UserIdentifier("<USER>", UserIdentifierType.OptionalDisplayableId), null, claims).ConfigureAwait(false);
+                AuthenticationResult result = await ctx.AcquireTokenAsync("https://graph.windows.net", ClientId,
+                        new Uri(IOSBrokerRedirectURI),
```
henrik-me, line 308: is this intentional or did you mean to use the Device redirect uri?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
```diff
@@ -364,17 +364,13 @@ public Intent GetIntentForBrokerActivity(AuthenticationRequest request, Activity
...
-                RequestContext.Logger.Error(e);
-                RequestContext.Logger.ErrorPii(e);
-            }
             catch (Exception e)
             {
                 // Authenticator gets problem from webrequest or file read/write
                 var ex = new AdalException("Authenticator cancels the request", e);
                 RequestContext.Logger.Error(ex);
                 RequestContext.Logger.ErrorPii(ex);
+                throw ex;
```
henrik-me: >throw ex; [](start = 16, length = 9)

was this a previous miss or is this an intentional behavior change? Is this handled anywhere in our code or do we pass this exception to the user?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
```diff
@@ -364,17 +364,13 @@ public Intent GetIntentForBrokerActivity(AuthenticationRequest request, Activity
...
             }
-            catch (OperationCanceledException e)
-            {
-                RequestContext.Logger.Error(e);
-                RequestContext.Logger.ErrorPii(e);
-            }
             catch (Exception e)
             {
                 // Authenticator gets problem from webrequest or file read/write
                 var ex = new AdalException("Authenticator cancels the request", e);
```
henrik-me: >"Authenticator cancels the request" [](start = 43, length = 35)

imo, the exeptions are missing an id, or somehow it should be easy for the developer to do the right thing in case an exception occurs. Is this message available in a const anywhere?
should it be cancelled instead?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
jennyf19: @SomkaPe and I looked this over. It is an intentional behavior change. The exception is passed to the user.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
```diff
@@ -364,17 +364,13 @@ public Intent GetIntentForBrokerActivity(AuthenticationRequest request, Activity
...
                 }
             }
-            catch (OperationCanceledException e)
-            {
-                RequestContext.Logger.Error(e);
-                RequestContext.Logger.ErrorPii(e);
-            }
             catch (Exception e)
             {
                 // Authenticator gets problem from webrequest or file read/write
```
henrik-me: >// Authenticator gets problem from webrequest or file read/write [](start = 16, length = 64)

is this comment relevant? #Resolved

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
henrik-me: shouldn't this be in the above now removed catch intead. and we should leave the generic exception alone?

---
In reply to: [205926366](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1140#discussion_r205926366) [](ancestors = 205926366)

jennyf19: The `OperationCancelledException` was just suppressed and written to the logs. I think we should catch the exceptions and throw an AdalException.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
jennyf19: it's coming from the broker intent, so it would make sense to make it a const message, but leave as is. our cancelled one is for user cancelling auth, which is not the case here.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
jennyf19: maybe...maybe not...doesn't seem relevant now. ⛄️

adal/devApps/XFormsApp/SecondPage.cs
jennyf19: that's what was there before, so i just replaced it w/what it had before.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
jennyf19: removed.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
jennyf19: done.

henrik-me: Cancels ... is that the right wording?

---
In reply to: [205926754](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1140#discussion_r205926754) [](ancestors = 205926754)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
henrik-me: yes, but you are now catching all exceptions and not just the specific one regarding operationcanceledexception... imo this code should be moved up to what was there before. and a general throw should be put in place for the general exception. (or it should not be handled at all)

---
In reply to: [205926683](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1140#discussion_r205926683) [](ancestors = 205926683)

jennyf19: addressed all PR comments and tested

PR #1143, opened 2018-07-23
Exception refactoring - core and msal

bgavrilMS:
1. Added a new core exception hierarchy
2. Adapters between public surface of MSAL and Core

Pending: Adapters between ADAL and Core. This is actually not strictly needed because ADAL doesn't seem to call into core and deal with exception, but it will, so I will send a separate PR for this. 

Note: Please have a look at individual commits, a lot of the code is boilerplate


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AdalServiceException.cs
```diff
@@ -118,12 +118,12 @@ internal AdalServiceException(string errorCode, string message, string[] service
         /// <summary>
         /// Gets the specific error codes that may be returned by the service.
         /// </summary>
-        public string[] ServiceErrorCodes { get; set; }
+        public string[] ServiceErrorCodes { get; private set; }
```
bgavrilMS, line 121: Unrelated - the setters can be made private.

core/src/Exceptions/CoreErrorCodes.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Microsoft.Identity.Core.Exceptions
+{
+    /// <summary>
+    /// Error codes attached to each exception.
+    /// These need to be duplicated and publicly exposed in the MSAL and ADAL because users refer to them
+    /// </summary>
+    internal class CoreErrorCodes //TODO: add a test that reflects over this class and the MSAL / ADAL ones
+    {
+        public const string JsonParseError = "json_parse_failed";
```
bgavrilMS, line 37: I am going to add a unit test that checks that these error codes are available as constants in MSAL.

core/src/Platforms/Android/SystemWebview/AndroidConstants.cs
```diff
@@ -27,7 +27,8 @@
 
 namespace Microsoft.Identity.Core.UI.SystemWebview
 {
-    internal static class MsalErrorAndroidEx
+    // TODO: rename this as "core" ?
```
bgavrilMS: Actually these are not used. I will delete them

PR #1145, opened 2018-07-24
Exception refactoring - allow for MSAL and ADAL exceptions in core

bgavrilMS:
The current design uses an abstract factory in core. Concrete factories are injected and throw Msal and Adal errors. 

Note: Adal does not yet use code from core that requires exceptions to be thrown. I will add a simple implementation only for ADAL.


core/src/Exceptions/CoreErrorMessages.cs
```diff
@@ -0,0 +1,60 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+
+namespace Microsoft.Identity.Core
+{
+    internal class CoreErrorMessages
```
henrik-me, line 31: >CoreErrorMessages [](start = 19, length = 17)

:)

core/src/Exceptions/CoreErrorCodes.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Microsoft.Identity.Core
+{
+    /// <summary>
+    /// Error codes attached to each exception.
+    /// These need to be duplicated and publicly exposed in the MSAL and ADAL because users refer to them
+    /// </summary>
+    internal class CoreErrorCodes //TODO: add a test that reflects over this class and the MSAL / ADAL ones
+    {
+        public const string JsonParseError = "json_parse_failed";
```
henrik-me: >string [](start = 21, length = 6)

@Jean-Marc:: is there a way we can switch to ints for these instead of strings (I realize it's breaking for ADAL thus we will have to convert to string for ADAL)... I believe the developer model is much better when integers, and all that string comparison is really waste of cpu cycles. thoughts?

core/src/IdToken.cs
```diff
@@ -98,8 +101,10 @@ public static IdToken Parse(string idToken)
             }
             catch (Exception exc)
             {
-                throw new MsalClientException(MsalClientException.JsonParseError,
-                    "Failed to parse the returned id token.", exc);
+                throw CoreExceptionFactory.Instance.GetClientException(
+                    CoreErrorCodes.JsonParseError,
+                    "Failed to parse the returned id token.",
```
henrik-me: >"Failed to parse the returned id token." [](start = 20, length = 40)

should these errors also be added to the CoreErrorMessages?

core/src/Http/HttpRequest.cs
```diff
@@ -125,11 +123,17 @@ private static HttpRequestMessage CreateRequestMessage(Uri endpoint, Dictionary<
                 requestContext.Logger.InfoPii(message);
                 if (toThrow != null)
                 {
-                    throw new MsalServiceException(MsalServiceException.RequestTimeout, "Request to the endpoint timed out.", toThrow);
+                    throw CoreExceptionFactory.Instance.GetServiceException(
+                        CoreErrorCodes.RequestTimeout,
+                        "Request to the endpoint timed out.",
```
henrik-me: >"Request to the endpoint timed out." [](start = 24, length = 36)

see related comment ... there are a number of those

core/src/Instance/AdfsAuthority.cs
```diff
@@ -52,7 +53,9 @@ protected override bool ExistsInValidatedAuthorityCache(string userPrincipalName
         {
             if (string.IsNullOrEmpty(userPrincipalName))
             {
-                throw new MsalException("UPN is required for ADFS authority validation.");
+                throw CoreExceptionFactory.Instance.GetClientException(
+                    "upn_required",
+                    "UPN is required for ADFS authority validation.");
```
henrik-me: ... will not add more messages on this. :)

core/src/Exceptions/CoreExceptionFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+
+using System;
+
+namespace Microsoft.Identity.Core
+{
+    /// <summary>
+    /// Abstract factory for spewing exceptions for Adal and Msal. Use the <see cref="Instance"/>
+    /// singleton to access an actual implementation which will have been injected.
+    /// </summary>
+    internal abstract class CoreExceptionFactory
```
henrik-me: >CoreExceptionFactory [](start = 28, length = 20)

I'm inclined to call this a singleton... even though it's not fully a singleton, but it's also not really a factory (imo)

henrik-me:
generally, looks awesome

msal/src/Microsoft.Identity.Client/Exceptions/MsalExceptionFactory.cs
```diff
@@ -0,0 +1,158 @@
...
+                if (ex is MsalException)
+                {
+                    MsalException msalException = ex as MsalException;
+                    sb.AppendLine(String.Format(
+                        CultureInfo.CurrentCulture,
+                        ", ErrorCode: {0}",
+                        msalException.ErrorCode));
+                }
+
+                if (ex is MsalServiceException)
```
henrik-me, line 150: if is...Exception could possibly also be moved to the factory class (allowing it to return a bool for these 2 cases). then we could perhaps move scrubbing to core? not a big deal and perhaps not worth the effort though.

henrik-me:
>     internal static class MsalErrorMessage

a number of strings are removed from here... how will clients validate those going forward when the values are not available here?

---
Refers to: msal/src/Microsoft.Identity.Client/Exceptions/MsalErrorMessage.cs:13 in f8899c4. [](commit_id = f8899c46b4b7469ffccd3417b6a349e76555540f, deletion_comment = False)

msal/src/Microsoft.Identity.Client/Exceptions/ExceptionExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+
+using System;
+using System.Globalization;
+using System.Text;
+
+namespace Microsoft.Identity.Client
+{
+    internal static class ExceptionExtensions
+    {
+        internal static string GetPiiScrubbedDetails(this Exception ex)
```
henrik-me: >GetPiiScrubbedDetails [](start = 31, length = 21)

seems like this  code also exists in the factory? perhaps calling this place from the factory?

henrik-me:
I'm surprised to see that nothing is done for ADAL. We probably don't need to do anything, but would we want to use the same pattern?

msal/src/Microsoft.Identity.Client/Internal/PlatformPlugin.cs
```diff
@@ -45,21 +45,21 @@ static PlatformPlugin()
 
         public static void InitializeByAssemblyDynamicLinking()
         {
+            CoreExceptionFactory.Instance = new MsalExceptionFactory();
```
henrik-me: >CoreExceptionFactory.Instance = new MsalExceptionFactory(); [](start = 12, length = 59)

minor: as discussed perhaps moving to an actual assembly loaded event as that reveals the intent much better than this... it's a bit of magic that this happens before anything else is called

henrik-me (approved):
:shipit:

henrik-me (changes requested):
:clock1:

core/src/Instance/AdfsAuthority.cs
jennyf19: can `upn_required` be a const somewhere? same w/the error message. maybe that's what @henrik-me  is referring to?

jennyf19 (commented):
I think this looks great. I'm curious as to what is happening on the adal side of things (how is ADAL handling the core exceptions?). I'll pull your branch later and try a few things out and get back to you. :)

bgavrilMS:
I haven’t added the AdalExceptionFactory yet because it is not used. I’ll probably add a simple one as I address the other PR comments.

core/src/Exceptions/CoreExceptionFactory.cs
bgavrilMS: I'll call it CoreExceptionService

msal/src/Microsoft.Identity.Client/Exceptions/ExceptionExtensions.cs
bgavrilMS: Yes, good point.

msal/src/Microsoft.Identity.Client/Internal/PlatformPlugin.cs
bgavrilMS: Will address this and an AdalFactory in a separate PR.

core/src/Exceptions/CoreExceptionFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+
+using System;
+
+namespace Microsoft.Identity.Core
+{
+    /// <summary>
+    /// Abstract factory for spewing exceptions for Adal and Msal. Use the <see cref="Instance"/>
+    /// singleton to access an actual implementation which will have been injected.
+    /// </summary>
+    internal abstract class CoreExceptionService
```
rayluo, line 36: Since there will be one set of the exceptions named "ServiceException" later, shall we name this overall factory just `CoreExceptionFactory`? As a matter of fact, you did name those concrete implementations as something like `MsalExceptionFactory`.

core/src/Instance/AdfsAuthority.cs
```diff
@@ -134,7 +147,7 @@ protected override void AddToValidatedAuthorities(string userPrincipalName)
             catch (Exception exc)
             {
                 const string msg = "On-Premise ADFS enrollment server endpoint lookup failed. Error - ";
-                string noPiiMsg = exc.GetPiiScrubbedDetails();
+                string noPiiMsg = CoreExceptionService.Instance.GetPiiScrubbedDetails(exc);
```
rayluo, line 150: The previous implementation was conceptually imprecise because an `Exception` instance does not necessarily have a method `GetPiiScrubbedDetails()`. This new implementation happens to eliminate that problem. 😄 👍

msal/src/Microsoft.Identity.Client/Exceptions/MsalExceptionFactory.cs
```diff
@@ -0,0 +1,187 @@
...
+using System;
+using System.Globalization;
+using System.Text;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Implementation of the <see cref="CoreExceptionService"/> that throws <see cref="MsalException"/>
+    /// </summary>
+    internal class MsalExceptionFactory : CoreExceptionService
```
henrik-me, line 39: >MsalExceptionFactory [](start = 19, length = 20)

I believe Ray is saying something similar... should this have been named Service?

msal/src/Microsoft.Identity.Client/Exceptions/MsalServiceException.cs
```diff
@@ -45,6 +45,21 @@ public class MsalServiceException : MsalException
...
+
+        /// <summary>
+        /// No passive auth endpoint
+        /// </summary>
+        public const string MissingPassiveAuthEndpoint = "missing_passive_auth_endpoint";
+
+        /// <summary>
+        /// Invalid authority
+        /// </summary>
+        public const string InvalidAuthroity = "invalid_authority";
```
henrik-me, line 61: >InvalidAuthroity [](start = 28, length = 16)

spelling: InvalidAuthority

henrik-me:
approving. hoping to see the AdalExceptionService, as very soon there will be code sharing to both adal and msal which needs to share exceptions.... would be great to ensure the framework is available.

henrik-me:
kindly validate the public API after this change and ensure everything is captured. approving based on comments being addressed and additional validation completes without issues.

henrik-me (approved):
:shipit:

bgavrilMS:
Some minor changes like naming will be addressed in the next 2 PRs

msal/src/Microsoft.Identity.Client/Exceptions/MsalExceptionFactory.cs
rayluo: I was actually saying the other way around. I would suggest them to be named Factory, to avoid a similar name already used by its [method GetServiceException(...)](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1145/files/d88cb3a466401f89ea4a837d4eb1a551bde06791#diff-00356570595e619b39da7095f2dabb15R45).
Given that this PR is now merged without changing this name, I think we can continue this conversation in those follow-up PRs.

PR #1147, opened 2018-07-25
Create explicit module initialization entry points

bgavrilMS:


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ModuleInitializer.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal
+{
+    /// <summary>
+    /// Initializes the MSAL module. This can be considered an entry point into MSAL
+    /// for initialization purposes. 
+    /// </summary>
+    /// <remarks>
+    /// The CLR defines a module initilizer, however this is not implemented in C# and to
```
jmprieur: defines **the notion of** module initializer? 
also you have a typo in *initializer*

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ModuleInitializer.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal
+{
+    /// <summary>
+    /// Initializes the MSAL module. This can be considered an entry point into MSAL
+    /// for initialization purposes. 
+    /// </summary>
+    /// <remarks>
+    /// The CLR defines a module initilizer, however this is not implemented in C# and to 
+    /// use this it would require IL weaving, which does not seem to work on all target frameworks.
+    /// Instead, call <see cref="EnsureModuleInitialized"/> from static ctors of public entry points.
```
jmprieur: public entry points : anything that can be instantiated in the library? AuthenticationContext, PublicClientApplication, ConfidentialClientApplication (I have not seen), UIParent, PlatformParameter ? 

Do you think you have them all both in ADAL and MSAL (I have not seen the static constructor for `ConfidentialClientApplication` for instance, and maybe we'd want to EnsureModuleInitialized in the BaseClientApplication ?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ModuleInitializer.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// </remarks>
+    internal class ModuleInitializer
+    {
+        private static bool isInitialized = false;
+
+        public static void EnsureModuleInitialized()
+        {
+            if (!isInitialized)
+            {
+                System.Runtime.CompilerServices.RuntimeHelpers.RunClassConstructor(
```
jmprieur: Doesn't this require more security attributes than we had so far? (breaking change?)

msal/src/Microsoft.Identity.Client/PublicClientApplication.cs
```diff
@@ -42,6 +42,11 @@ namespace Microsoft.Identity.Client
     /// </summary>
     public sealed partial class PublicClientApplication : ClientApplicationBase, IPublicClientApplication
     {
+        static PublicClientApplication()
+        {
+            ModuleInitializer.EnsureModuleInitialized();
```
jmprieur, line 47: Why not in ClientApplicationBase rather?

jmprieur (changes requested):
This seems neat (and some code seems less fragile)
I'm wondering if you have applied it to everything (I have not seen ConfidentialClientApplications for instance).

How can we make sure that if we introduce a new class requiring logger or other, we'd need to add a static constructor and that static constructor calls EnsureModuleInitialized?

msal/src/Microsoft.Identity.Client/PublicClientApplication.cs
bgavrilMS: Static constructors do not fire unless the actual class is used. It does not fire if you use a derived class.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ModuleInitializer.cs
bgavrilMS: This is pre-existing code and didn't really want to touch it. But you have a good point, I'll move the logic of that static initializer to the module initializer.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ModuleInitializer.cs
bgavrilMS: Excellent points.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ModuleInitializer.cs
bgavrilMS: The AdalLogger implementation needs work, it's too complex for a logger with a callback. I will inistantiate a logger here.

msal/src/Microsoft.Identity.Client/PublicClientApplication.cs
jmprieur: oh, I did not know! thanks for the explanation.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/PlatformParameters.cs
```diff
@@ -37,6 +38,11 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     [CLSCompliant(false)]
     public class PlatformParameters : IPlatformParameters
     {
+        static PlatformParameters()
+        {
+            ModuleInitializer.EnsureModuleInitialized();
```
henrik-me, line 43: >ModuleInitializer.EnsureModuleInitialized(); [](start = 12, length = 44)

I find it weird that it goes here... I realize this is what ADAL does today but still, to me this is very odd and not intuitive

henrik-me:
I'm a bit concerned that we have to throw the call to the initializer in so many places. It should be possible to have one entry point for each library... like void main() ... perhaps a public static which is forced to load ....

msal/tests/Test.MSAL.NET.Unit/ModuleInitializerTest.cs
```diff
@@ -0,0 +1,36 @@
...
+    public class ModuleInitializerTest
+    {
+        [TestMethod]
+        public void InitializesExceptionsAndLogs()
+        {
+            // Act
+            ModuleInitializer.EnsureModuleInitialized();
+
+            // Assert
+            MsalExceptionFactory factory = CoreExceptionService.Instance as MsalExceptionFactory;
```
henrik-me, line 23: we are back to the question on naming from another PR... perhaps both should be called service?

bgavrilMS:
I wasn't able to find something like this. If we had control over the application, we could hook to the assembly loaded event, but as a library we cannot.

The alternative is to use a "module constructor", which is something that the CLR defines (cctor IL instruction). However, C# does not define this. I found a project called Fody/ModuleInit that injects IL code into the assembly after build, but I could not get it to work with ADAL, only with MSAL.

msal/src/Microsoft.Identity.Client/Internal/ModuleInitializer.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// for initialization purposes. 
+    /// </summary>
+    /// <remarks>
+    /// The CLR defines a module initializer, however this is not implemented in C# and to 
+    /// use this it would require IL weaving, which does not seem to work on all target frameworks.
+    /// Instead, call <see cref="EnsureModuleInitialized"/> from static ctors of public entry points.
+    /// </remarks>
+    internal class ModuleInitializer
+    {
+        private static bool isInitialized = false;
```
henrik-me, line 44: I believe its not required, but not sure. Should this be declared as volatile?

bgavrilMS:
Yes, I will have renamed it in the subsequent PR where I create the ADALException Service

henrik-me (changes requested):
:clock1:

msal/src/Microsoft.Identity.Client/Internal/ModuleInitializer.cs
bgavrilMS: I've added a lock to be on the safe side (this is in the Adal Exception Factory PR, but I can add it here if needed)

msal/tests/Test.MSAL.NET.Unit/ModuleInitializerTest.cs
bgavrilMS: This will have been fixed in the next PR related to ADAL Exception Service

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/PlatformParameters.cs
bgavrilMS: I somewhat agree. PlatformParameters is a data class and should neither log nor throw exceptions. But it is possible that a developer adds logging or exception code here. In any case, these static constructors fire once, irrespective of threads etc. so there shouldn't be a performance issue.

PR #1148, opened 2018-07-25
Adal exception factory hooked into module init

bgavrilMS:


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Exceptions/AdalExceptionService.cs
```diff
@@ -0,0 +1,139 @@
...
+using System;
+using System.Globalization;
+using System.Text;
+
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory
+{
+    /// <summary>
+    /// Implementation of the <see cref="CoreExceptionService"/> that throws <see cref="AdalException"/>
+    /// </summary>
+    /// <remarks>Does not currently throw <see cref="AdalSilentTokenAcquisitionException"/> and
```
jmprieur, line 38: What does this remark mean? Can you please elaborate (I'm confused if this is a good thing, a bad thing, more work is required ?)

jmprieur (commented):
added a question

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Exceptions/AdalExceptionService.cs
henrik-me: from related PR, this is the placeholder for ADAL exceptions ensuring we follow the same path for both libraries. I assume it means those exceptions are still thrown in place and not using this structure.

---
In reply to: [205396669](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1148#discussion_r205396669) [](ancestors = 205396669)

bgavrilMS: Yes, exactly, I am not sure we'll ever throw those exceptions from core. Both adal and msal still throw exceptions using `throw new AdalServiceException` and do not use the factories.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Exceptions/AdalExceptionService.cs
```diff
@@ -0,0 +1,139 @@
...
+using System.Text;
+
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory
+{
+    /// <summary>
+    /// Implementation of the <see cref="CoreExceptionService"/> that throws <see cref="AdalException"/>
+    /// </summary>
+    /// <remarks>Does not currently throw <see cref="AdalSilentTokenAcquisitionException"/> and 
+    /// <see cref="AdalUserMismatchException"/></remarks>
+    internal class AdalExceptionService : CoreExceptionService
```
henrik-me, line 40: >AdalExceptionService [](start = 19, length = 20)

I think I understand the name Factory used previously as this causes a bit of a name confusion with AdalServiceException...

msal/src/Microsoft.Identity.Client/Exceptions/MsalExceptionService.cs
```diff
@@ -36,7 +36,7 @@ namespace Microsoft.Identity.Client
     /// <summary>
     /// Implementation of the <see cref="CoreExceptionService"/> that throws <see cref="MsalException"/>
     /// </summary>
-    internal class MsalExceptionFactory : CoreExceptionService
+    internal class MsalExceptionService : CoreExceptionService
```
henrik-me, line 39: >MsalExceptionService [](start = 19, length = 20)

this naming is confusing... sorry, perhaps going back to factory is better...

henrik-me (approved):
:shipit:

msal/src/Microsoft.Identity.Client/Exceptions/MsalExceptionService.cs
bgavrilMS: I'll revert to the old name (*ExceptionFactory) and get this checked in

PR #1149, opened 2018-07-26
Remove unsused adal specific public enum from core.

bgavrilMS:


henrik-me (approved):
:shipit:

rayluo (approved):
Since they are not currently used, and all the tests still pass after the removal, this is a valid fix.

And this would also mean that ADAL keeps its own PromptBehavior, and MSAL keeps its own [UIBehavior](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Acquiring-tokens-interactively#controlling-the-interactivity-with-the-user-with-the-behavior-parameter-uibehavior), we are not attempting to combine them in core.

PR #1151, opened 2018-07-27
authority aliases support in msal

SomkaPe:
Adding authority aliases support.
In order to eliminate “islands of SSO” effects when tokens  in cache can be stored under different authority aliases like login.windows.net and login.microsoftonline.com,  we do issue instance discovery call to get authority metadata which contains aliases and pref cache and pref network calls aliases.
Pref network value should be used for network calls,
Pref cache value should be used while caching tokens ( for write operations pref cache value used, for read operation all aliases should be checked)


core/src/Instance/AadInstanceDiscovery.cs
```diff
@@ -0,0 +1,111 @@
...
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Core.Instance
+{
+    internal static class AadInstanceDiscovery
+    {
+        // Keys are host strings
```
henrik-me: >// Keys are host strings [](start = 8, length = 24)

remove comment, it's already in the field signature. Consider however to comment on the class and make it a singleton instead as that reveals the intent (imo) of this class better.  #Resolved

core/src/Instance/Authority.cs
```diff
@@ -118,27 +118,41 @@ public static void ValidateAsUri(string authority)
             }
         }
 
-        private static Authority CreateInstance(string authority, bool validateAuthority)
+        internal static AuthorityType GetAuthorityType(string authority)
         {
-            authority = CanonicalizeUri(authority);
-            ValidateAsUri(authority);
             string[] pathSegments = new Uri(authority).AbsolutePath.Substring(1).Split('/');
```
henrik-me: >new Uri(authority).AbsolutePath.Substring(1).Split('/'); [](start = 36, length = 56)

can you use this: [https://msdn.microsoft.com/en-us/library/system.uri.segments(v=vs.110).aspx](https://msdn.microsoft.com/en-us/library/system.uri.segments(v=vs.110).aspx) #Resolved

henrik-me: [https://docs.microsoft.com/en-us/dotnet/api/system.uri.segments?view=netframework-4.7.1#System_Uri_Segments](https://docs.microsoft.com/en-us/dotnet/api/system.uri.segments?view=netframework-4.7.1#System_Uri_Segments)

---
In reply to: [205900836](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1151#discussion_r205900836) [](ancestors = 205900836)

henrik-me:
>     }

should be in a separate file #Resolved

---
Refers to: core/src/Instance/Authority.cs:43 in 9e71d2e. [](commit_id = 9e71d2e7732be85bf66e57f0384602df2695c252, deletion_comment = False)

henrik-me:
>             string[] pathSegments = authorityUri.Uri.AbsolutePath.Substring(1).Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);

[https://docs.microsoft.com/en-us/dotnet/api/system.uri.segments?view=netframework-4.7.1#System_Uri_Segments](https://docs.microsoft.com/en-us/dotnet/api/system.uri.segments?view=netframework-4.7.1#System_Uri_Segments) #Resolved

---
Refers to: core/src/Instance/Authority.cs:65 in 9e71d2e. [](commit_id = 9e71d2e7732be85bf66e57f0384602df2695c252, deletion_comment = False)

core/src/Instance/Authority.cs
```diff
@@ -118,27 +118,41 @@ public static void ValidateAsUri(string authority)
...
+            {
+                case AuthorityType.Adfs:
+                    throw CoreExceptionFactory.Instance.GetClientException(CoreErrorCodes.InvalidAuthorityType,
+                       "ADFS is not a supported authority");
+
+                case AuthorityType.B2C:
+                    return new B2CAuthority(authority, validateAuthority);
+
+                default:
+                    return new AadAuthority(authority, validateAuthority);
```
henrik-me: suggestion: switch the AuthorityType.Aad and throwing an exception in default. has downsides, thus looking forward to your thoughts on this

core/src/Instance/InstanceDiscoveryResponse.cs
```diff
@@ -35,5 +35,8 @@ internal sealed class InstanceDiscoveryResponse : OAuth2ResponseBase
     {
         [DataMember(Name = "tenant_discovery_endpoint", IsRequired = false)]
         public string TenantDiscoveryEndpoint { get; set; }
+
+        [DataMember(Name = "metadata")]
```
henrik-me, line 39: seems like this should be required. however the same above. Do you know why the tenent_discovery_endpoint is not required?

msal/samples/desktop/SampleApp/MainForm.cs
```diff
@@ -43,7 +43,7 @@ public MainForm()
             tabControl1.Appearance = TabAppearance.FlatButtons;
             tabControl1.ItemSize = new Size(0, 1);
             tabControl1.SizeMode = TabSizeMode.Fixed;
-            user = _msalHelper.Application.Users.First();
+            user = _msalHelper.Application.GetUsers().Result.First();
```
henrik-me: >First [](start = 61, length = 5)

FirstOrDefault ? and make the sample app handle null elegantly?

henrik-me:
would be great to see a description of your change in the PR... right now it's blank

henrik-me:
let me know your thoughts on my comments. marking for waiting until I see your responses. #Resolved

henrik-me (changes requested):
:clock1:

core/src/Instance/InstanceDiscoveryResponse.cs
SomkaPe: in case of some error these fields will not be present , instead fields like 'error' 'error_description' will be returned

core/src/Instance/AadInstanceDiscovery.cs
SomkaPe: switched to singelton #Resolved

core/src/Instance/Authority.cs
SomkaPe: switched to Segments api #Resolved

core/src/Instance/Authority.cs
SomkaPe: changed

msal/samples/desktop/SampleApp/MainForm.cs
SomkaPe: changed

core/src/Instance/AadAuthority.cs
```diff
@@ -35,63 +35,51 @@ namespace Microsoft.Identity.Core.Instance
...
+            "login.microsoftonline.de", // Microsoft Azure Blackforest
+            "login-us.microsoftonline.com", // Microsoft Azure US Government - Legacy
             "login.microsoftonline.us", // Microsoft Azure US Government
             "login.microsoftonline.com", // Microsoft Azure Worldwide
-            "login.microsoftonline.de" // Microsoft Azure Blackforest
+            "login.cloudgovapi.us" // Microsoft Azure US Government
         };
 
-        public AadAuthority(string authority, bool validateAuthority) : base(authority, validateAuthority)
+        public const string DefaultTrustedAuthority = "login.microsoftonline.com";
```
bgavrilMS, line 49: Inconsistent naming - TrustedHostList  / DefaultTrustedAuthority  - also inconsistent visibility - please make both public as the class is internal.

core/src/Instance/AadAuthority.cs
```diff
@@ -35,63 +35,51 @@ namespace Microsoft.Identity.Core.Instance
 {
     internal class AadAuthority : Authority
     {
-        private const string AadInstanceDiscoveryEndpoint = "https://login.microsoftonline.com/common/discovery/instance";
-        private static readonly HashSet<string> TrustedHostList = new HashSet<string>()
+        internal static readonly HashSet<string> TrustedHostList = new HashSet<string>()
```
bgavrilMS, line 38: I assume this should be read-only, consider using a property { get; private set;} and maybe use an ImmutableHashSet

core/src/Instance/AadAuthority.cs
```diff
@@ -114,5 +102,10 @@ internal static bool IsInTrustedHostList(string host)
                 !string.IsNullOrEmpty(
                     TrustedHostList.FirstOrDefault(a => string.Compare(host, a, StringComparison.OrdinalIgnoreCase) == 0));
         }
+
+        internal override async Task Init(RequestContext requestContext)
```
bgavrilMS, line 106: Please place this near the ctor.

core/src/Instance/AadAuthority.cs
```diff
@@ -114,5 +102,10 @@ internal static bool IsInTrustedHostList(string host)
...
+        {
+            await UpdateCanonicalAuthorityAsync(requestContext).ConfigureAwait(false);
```
bgavrilMS, line 108: How will users of this class know they have to call Init?

core/src/Instance/AadInstanceDiscovery.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Core.Instance
+{
+    internal class AadInstanceDiscovery
+    {
+        AadInstanceDiscovery(){}
```
bgavrilMS, line 17: why?

core/src/Instance/AadInstanceDiscovery.cs
```diff
@@ -0,0 +1,114 @@
...
+                        .ConfigureAwait(false);
+
+            return discoveryResponse;
+        }
+
+        private static void Validate(InstanceDiscoveryResponse instanceDiscoveryResponse)
+        {
+            if (instanceDiscoveryResponse.TenantDiscoveryEndpoint == null)
+            {
+                throw CoreExceptionFactory.Instance.GetClientException(instanceDiscoveryResponse.Error,
```
bgavrilMS, line 91: I think for maintainability we should make sure all error codes are public in adal / msal. @henrik-me to confirm this is a goal

core/src/Instance/AadInstanceDiscovery.cs
```diff
@@ -0,0 +1,114 @@
...
+                     instanceDiscoveryResponse.ErrorDescription);
+            }
+        }
+
+        private void CacheInstanceDiscoveryMetadata(string host, InstanceDiscoveryResponse instanceDiscoveryResponse)
+        {
+            foreach (var entry in instanceDiscoveryResponse?.Metadata ?? Enumerable.Empty<InstanceDiscoveryMetadataEntry>())
```
bgavrilMS, line 98: nice #Resolved

core/src/Instance/Authority.cs
```diff
@@ -118,27 +111,50 @@ public static void ValidateAsUri(string authority)
...
+        internal static string GetFirstPathSegment(string authority)
+        {
+            return new Uri(authority).Segments[1].TrimEnd('/');
+        } 
+
+        internal static AuthorityType GetAuthorityType(string authority)
+        {
+            var firstPathSegment = GetFirstPathSegment(authority);
+
+            if (string.Compare(firstPathSegment, "adfs", StringComparison.OrdinalIgnoreCase) == 0)
```
bgavrilMS, line 123: string.Equals

core/src/Instance/Authority.cs
```diff
@@ -118,27 +111,50 @@ public static void ValidateAsUri(string authority)
             }
         }
 
+        internal static string GetFirstPathSegment(string authority)
+        {
+            return new Uri(authority).Segments[1].TrimEnd('/');
+        } 
+
+        internal static AuthorityType GetAuthorityType(string authority)
```
bgavrilMS, line 119: This class seems to do too much, consider having a separate AuthorityFactory

core/src/Instance/B2CAuthority.cs
```diff
@@ -35,7 +35,7 @@ internal class B2CAuthority : AadAuthority
     {
         public const string Prefix = "tfp"; // The http path of B2C authority looks like "/tfp/<your_tenant_name>/..."
 
-        public B2CAuthority(string authority, bool validateAuthority) : base(authority, validateAuthority)
+        internal B2CAuthority(string authority, bool validateAuthority) : base(authority, validateAuthority)
```
bgavrilMS, line 38: a public method in an internal class has the same visibility as an internal method in an internal class, so there is no reason to change this.

bgavrilMS (commented):
LGTM apart minor changes

core/src/Instance/AadInstanceDiscovery.cs
henrik-me: Agree with the goal. Not sure how we are going to expose these for Core... at this point in time. Only way is to have a public interface in Core, which I'm not sure we agreed upon.

core/src/Instance/B2CAuthority.cs
henrik-me: Did we agree to change to same visibility or only set visibility on the class? not sure where we are with that.

core/src/Instance/AadInstanceDiscovery.cs
```diff
@@ -0,0 +1,114 @@
...
+
+namespace Microsoft.Identity.Core.Instance
+{
+    internal class AadInstanceDiscovery
+    {
+        AadInstanceDiscovery(){}
+
+        public static AadInstanceDiscovery Instance { get; } = new AadInstanceDiscovery();
+
+        internal readonly ConcurrentDictionary<string, InstanceDiscoveryMetadataEntry> InstanceCache =
```
henrik-me, line 21: minor: could this be named Cache?

henrik-me:
I like the refactorings and actions on the comments. A few newer comments are available. I ack this change assuming you will be responding to all the remaining comments and open work items for any additional work for Auth aliases. #Pending

henrik-me (approved):
:shipit:

henrik-me:
Please ensure to document the changes in the interface for MSAL in the wiki ensuring developers have an easy to read doc to get them from the current MSAL to the new one. There seems to be a number of relevant cache information to notify about including how to handle the api changes most effectively.

core/src/Instance/AadInstanceDiscovery.cs
SomkaPe: private constructor, singleton pattern

core/src/Instance/AadAuthority.cs
SomkaPe: good point it is host not authority, changed to host

core/src/Instance/AadAuthority.cs
SomkaPe: yes, it is read only, did not managed to add ImmutableHashSet - it is not visible due to some issues with multi targeting projects

core/src/Instance/AadAuthority.cs
SomkaPe: moved

core/src/Instance/Authority.cs
SomkaPe: switched

core/src/Instance/B2CAuthority.cs
SomkaPe: public does not reflect reality even though compiler is fine with that

core/src/Instance/AadInstanceDiscovery.cs
SomkaPe: switched

core/src/Instance/AadAuthority.cs
SomkaPe: removed Init

core/src/Instance/AadInstanceDiscovery.cs
SomkaPe: we need to expose exact OAuth2Response eror codes returned from STS to simplify maintenance

PR #1152, opened 2018-07-29
Add int values to PromptBehavior enums

jennyf19:


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Features/NonWinCommon/PromptBehavior.cs
```diff
@@ -37,27 +37,27 @@ public enum PromptBehavior
         /// Acquire token will prompt the user for credentials only when necessary.  If a token
         /// that meets the requirements is already cached then the user will not be prompted.
         /// </summary>
-        Auto,
+        Auto = 0,
```
bgavrilMS, line 40: This all seems to be in line with the Enum design guidelines https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum 

I assume the reason for doing this is to add that the default value for PromptBehaviour is Auto? If so, please add an enum comment.

bgavrilMS (approved):
LGTM, see minor comment inline

henrik-me:
I assume this will not be a breaking change as likely the output would be the same. Did you validate?

henrik-me (approved):
:shipit:

jennyf19:
closing as the change will be in this [PR](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1154)

PR #1153, opened 2018-07-30
Remove winrt support and add UAP support

bgavrilMS:
The only logic change is related to the User / UserInformation API, which affects how ADAL deals with ClientCrendentials(), i.e. it tries to figure out the principal name of the current user (e.g. bogavril@microsoft.com) 

Tested with: 
- MSA user (this fails as there is no principal name)
- AD-joined machine  
- AAD-joined machine
- app without capabilities (this fails and the error message suggests which capabilities to add)

For reviewers: there are 2 commits, first one with the logic changes, second is less important and deals with dev app changes)

Note about builds: 

- the builds will fail because they are still trying to build the old VS2015 artefacts
- please don't merge this PR, I will do it and then immediately fix the builds (I have draft versions of the builds for this branch)


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/uap/PlatformInformation.cs
```diff
@@ -53,21 +56,69 @@ public override string GetEnvironmentVariable(string variable)
...
+            var userDetails = await Task.WhenAll(getUserDetailTasks).ConfigureAwait(false);
+
+            // try to get a user that has both domain name and upn
+            var userDetailWithDomainAndPn = userDetails.FirstOrDefault(
+                d => !String.IsNullOrWhiteSpace(d.Domain) &&
+                !String.IsNullOrWhiteSpace(d.PrincipalName));
+
+            if (userDetailWithDomainAndPn != null)
+            {
+                return userDetailWithDomainAndPn.PrincipalName;
```
bgavrilMS, line 94: Exceptions are not thrown if the app manifest does not have the right capabilities, the properties returned are empty

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/uap/PlatformInformation.cs
```diff
@@ -53,21 +56,69 @@ public override string GetEnvironmentVariable(string variable)
...
+                return userDetailWithPn.PrincipalName;
+            }
+
+            // user has domain name, but no upn -> missing Enterprise Auth capability
+            if (userDetails.Any(d => !String.IsNullOrWhiteSpace(d.Domain)))
             {
-                throw new AdalException(AdalErrorEx.UnauthorizedUserInformationAccess, AdalErrorMessageEx.UnauthorizedUserInformationAccess, ex);
+                throw new AdalException(
+                 AdalErrorEx.CannotAccessUserInformationOrUserNotDomainJoined,
+                 CoreErrorMessages.UapCannotFindUpn);
```
bgavrilMS, line 111: "Cannot find the principal name for the user, but found a domain the name. Possible cause: the application does not request the Enterprise Authentication capability.";

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/uap/PlatformInformation.cs
```diff
@@ -53,21 +56,69 @@ public override string GetEnvironmentVariable(string variable)
...
             }
+
+            // no domain, no upn -> missing User Info capability
+            throw new AdalException(
+             AdalErrorEx.CannotAccessUserInformationOrUserNotDomainJoined,
+             CoreErrorMessages.UapCannotFindDomainUser);
```
bgavrilMS, line 117: "Cannot find the principal name for the user. Possible causes: the application does not request the User Account Information and the Enterprise Authentication capabilities or the user is not AD or AAD joined.";

msal/src/Microsoft.Identity.Client/Platforms/uap/PlatformInformation.cs
```diff
@@ -72,32 +74,8 @@ public override string GetDeviceModel()
...
-            {
-                requestContext.Logger.Warning(ae.Message);
-                const string msg = "Cannot try Windows Integrated Auth due to lack of Enterprise capability.";
-                requestContext.Logger.Info(msg);
-                requestContext.Logger.InfoPii(msg);
-                // This mostly means Enterprise capability is missing, so WIA cannot be used and
-                // we return true to add form auth parameter in the caller.
-                return true;
-            }
+            IReadOnlyList<Windows.System.User> users = await Windows.System.User.FindAllAsync();
```
henrik-me: >IReadOnlyList<Windows.System.User> users = await Windows.System.User.FindAllAsync(); [](start = 12, length = 84)

minor: change we are no longer telling in the log that there is no WIA users available (local)

msal/src/Microsoft.Identity.Client/Platforms/uap/PlatformInformation.cs
```diff
@@ -72,32 +74,8 @@ public override string GetDeviceModel()
...
+            return users.Any(u => u.Type == UserType.LocalGuest || u.Type == UserType.LocalGuest);
```
henrik-me: >u.Type == UserType.LocalGuest || u.Type == UserType.LocalGuest [](start = 34, length = 62)

what's the difference? should one of these have been User.LocalUser ?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/uap/PlatformInformation.cs
```diff
@@ -90,33 +141,8 @@ public override string GetDeviceModel()
...
-                var msg = "Cannot try Windows Integrated Authentication due to lack of Enterprise capability.";
-                requestContext.Logger.Info(msg);
-                requestContext.Logger.InfoPii(msg);
-
-                // This mostly means Enterprise capability is missing, so WIA cannot be used and
-                // we return true to add form auth parameter in the caller.
-                return true;
-            }            
+            IReadOnlyList<User> users = await User.FindAllAsync();
+            return users.Any(u => u.Type == UserType.LocalGuest || u.Type == UserType.LocalGuest);
```
henrik-me: >u.Type == UserType.LocalGuest || u.Type == UserType.LocalGuest) [](start = 34, length = 63)

see comment for MSAL as well

msal/src/Microsoft.Identity.Client/Platforms/uap/PlatformInformation.cs
```diff
@@ -112,6 +90,7 @@ public override string GetRedirectUriAsString(Uri redirectUri, RequestContext re
                 : redirectUri.OriginalString;
         }
 
+
         private static class NativeMethods
```
henrik-me: > private static class NativeMethods [](start = 7, length = 35)

Can this be removed/changed to something .net'ish?
- it seems like a no go to use NativeMethods in uap

if not.. this should belong to another file.

henrik-me (changes requested):
:clock1:

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/uap/PlatformInformation.cs
```diff
@@ -37,14 +37,17 @@
 using Windows.System.UserProfile;
 using Microsoft.Identity.Core;
 using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.OAuth2;
+using Windows.System;
+using System.Collections.Generic;
+using Windows.Foundation.Collections;
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
 {
     internal class PlatformInformation : PlatformInformationBase
```
henrik-me: >PlatformInformation [](start = 19, length = 19)

would it make sense to have a base for Uap? Not sure why some of this is different for adal vs. msal

msal/src/Microsoft.Identity.Client/Platforms/uap/PlatformInformation.cs
bgavrilMS: Yes, good catch.

msal/src/Microsoft.Identity.Client/Platforms/uap/PlatformInformation.cs
bgavrilMS: I did a quick test and this seems to work inside a WIn app. I was thinking of refactoring PlatformInformation but it would introduce quite a lot of code churn so close to the release. I think it would be safer to leave as is for now. 

I'll refactor NativeMethods by moving them to core.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/uap/PlatformInformation.cs
bgavrilMS: Yes and no. I agree that platform information could use a refactoring as it is too complex and has no tests (mainly because of multi-targetting). However, this is possibly not the best time to touch it before release. A base UAP class is non trivial because there already exists an inheritance hierarchy.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/uap/PlatformInformation.cs
henrik-me: Will you update this one as well?

---
In reply to: [206241535](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1153#discussion_r206241535) [](ancestors = 206241535)

msal/src/Microsoft.Identity.Client/Platforms/uap/PlatformInformation.cs
henrik-me: ok

---
In reply to: [206492047](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1153#discussion_r206492047) [](ancestors = 206492047)

msal/src/Microsoft.Identity.Client/Platforms/uap/PlatformInformation.cs
henrik-me: pls. see the other place where there is a similar change required

---
In reply to: [206491651](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1153#discussion_r206491651) [](ancestors = 206491651)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/uap/PlatformInformation.cs
henrik-me: ok

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/uap/PlatformInformation.cs
henrik-me: LocalUser ???

---
In reply to: [206592998](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1153#discussion_r206592998) [](ancestors = 206592998,206241535)

henrik-me:
one place more I believe is missing LocalUser intead of LocalGuest

henrik-me (approved):
seems like all important stuff was addressed :)

bgavrilMS:
Closing and pushing manually

PR #1155, opened 2018-07-31
Add private assests to `AsyncUsageAnalyzers`

jennyf19:
The analyzers leak by default into all dependent projects.


henrik-me (approved):
lgtm :shipit:

PR #1157, opened 2018-07-31
Jennyf/fix namespace

jennyf19:
I used ApiReviewer to compare currently released ADAL 3.19.8 and current dev. The namespace in the Telemetry and IdToken classes were not in Core. Since I was there, I checked that all the Core files have the license header information and removed unnecessary usings.
I ran the unit tests, rebuilt each individual project, and ran msbuild...will see how VSTS reacts.


core/src/OAuth2/OAuth2Client.cs
```diff
@@ -94,7 +90,7 @@ public void AddBodyParameter(string key, string value)
             HttpResponse response = null;
             Uri endpointUri = CreateFullEndpointUri(endPoint);
             var httpEvent = new HttpEvent() { HttpPath = endpointUri, QueryParams = endpointUri.Query };
-            Client.Telemetry.GetInstance().StartEvent(requestContext.TelemetryRequestId, httpEvent);
+            Telemetry.Telemetry.GetInstance().StartEvent(requestContext.TelemetryRequestId, httpEvent);
```
henrik-me: >Telemetry.Telemetry [](start = 12, length = 19)

while correct, it does look odd... consider re-using the alias from above to point to Telementry.Telemetry ... or perhaps the Telemetry class should be named differently #Closed

henrik-me:
>     public class Telemetry

Should this one be public from Core? . it will impact ADAL as well... 

---
Refers to: core/src/Telemetry.cs:38 in a7d013a. [](commit_id = a7d013a2962302494969a79851ed4cc5c1207abc, deletion_comment = False)

henrik-me:
>     public class Telemetry

This is basically a singleton... should we start naming those accordingly? Or as with the exception work.. Factory or Service? . It will also avoid us having code which talks about Telemetry.Telemetry #Closed

---
Refers to: core/src/Telemetry.cs:38 in a7d013a. [](commit_id = a7d013a2962302494969a79851ed4cc5c1207abc, deletion_comment = False)

jennyf19:
>     public class Telemetry

That would solve the weird Telemetry.Telemetry issue

---
In reply to: [409415457](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1157#issuecomment-409415457) [](ancestors = 409415457)

---
Refers to: core/src/Telemetry.cs:38 in a7d013a. [](commit_id = a7d013a2962302494969a79851ed4cc5c1207abc, deletion_comment = False)

henrik-me:
>     public class Telemetry

Perhaps the easiest is to keep the namespace as it was, then at least it will not be visible in the class explorer. Ideally though it should not be exposed in Core

---
In reply to: [409415240](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1157#issuecomment-409415240) [](ancestors = 409415240)

---
Refers to: core/src/Telemetry.cs:38 in a7d013a. [](commit_id = a7d013a2962302494969a79851ed4cc5c1207abc, deletion_comment = False)

henrik-me:
one minor comment regarding the visibility and naming of the class Telemetry. Marking as waiting.. however you can also revert that change and create a separate work item / PR to address that particular issue #Resolved

henrik-me (changes requested):
:clock1:

jennyf19:
>     public class Telemetry

but then it's available in ADAL, as an MSAL class. 

---
In reply to: [409415744](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1157#issuecomment-409415744) [](ancestors = 409415744,409415240)

---
Refers to: core/src/Telemetry.cs:38 in a7d013a. [](commit_id = a7d013a2962302494969a79851ed4cc5c1207abc, deletion_comment = False)

jennyf19:
sounds good. i'll think about it.

---
In reply to: [409416052](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1157#issuecomment-409416052) [](ancestors = 409416052)

jennyf19:
no problem making it internal and changing the name to TelemetryService (unless VSTS blows up...locally things were fine)

---
In reply to: [409416269](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1157#issuecomment-409416269) [](ancestors = 409416269,409416052)

henrik-me:
>     public class Telemetry

yes, it should not be available to ADAL, that's the real fix. However I wonder how people are supposed to hook up to the telemetry events. If you make this one internal users can't. Also unit tests will not catch this as we have internals visible in the utests. Do we have any test apps which uses Telemetry? We might have to use a similar approach as what Bogdan used for the Exceptions, ensuring we can inject the telemetry class from the Module Initializer he created.

---
In reply to: [409416173](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1157#issuecomment-409416173) [](ancestors = 409416173,409415744,409415240)

---
Refers to: core/src/Telemetry.cs:38 in a7d013a. [](commit_id = a7d013a2962302494969a79851ed4cc5c1207abc, deletion_comment = False)

henrik-me:
I just reviewed the code in the MSAL repo. This change will be a breaking change. We have to move this class back to msal and inject it into core. as mentioned I suggest following a path similar to the one Bogdan did (though there will be some differences).

jennyf19:
>     public class Telemetry

i reverted back the telemetry changes and will create a work item in VSTS for this.

---
In reply to: [409450300](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1157#issuecomment-409450300) [](ancestors = 409450300,409416173,409415744,409415240)

---
Refers to: core/src/Telemetry.cs:38 in a7d013a. [](commit_id = a7d013a2962302494969a79851ed4cc5c1207abc, deletion_comment = False)

henrik-me (approved):
:shipit:

PR #1160, opened 2018-08-01
Fix nuget and build warnings

bgavrilMS:
This should fix all the warnings except the one below, which is not repairable without a xamarin update 

Warning	VSX1000	No enough information has been provided to MSBuild in order to establish a connection to a Remote Server. The project 'Microsoft.Identity.Client' will continue building offline	Microsoft.Identity.Client	C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\MSBuild\Xamarin\iOS\Xamarin.Messaging.targets	41


jmprieur (approved):
LGTM

PR #1161, opened 2018-08-01
Enable rule UseAsyncSuffix and set it to error level + code fixes.

bgavrilMS:
Note that all the changes were done through refactoring, so there should be very low risk. As far as I can tell, the only public changes are in MSAL - GetUser, GetUsers and Remove


jennyf19 (approved):
🤖

henrik-me:
>     class AcquireTokenByAuthorizationCodeHandler : AcquireTokenHandlerBase

would have been nice if this one (and a few other places) was marked as internal

---
Refers to: adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenByAuthorizationCodeHandler.cs:35 in 1965741. [](commit_id = 196574149e1f24f8c46bdad0e10e99c9202bfa06, deletion_comment = False)

henrik-me:
Generally ok. Sample apps needs to be updated :)
As far as I could spot there are only the 2 public api changes to MSAL, right?

henrik-me (approved):
:shipit:

bgavrilMS:
@henrik-me  - there are 3 public api changes in MSAL - GetUserAsync, GetUsersAsync and RemoveAsync

PR #1164, opened 2018-08-02
Docs fix for https://github.com/AzureAD/microsoft-authentication-libr…

bgavrilMS:
…ary-for-dotnet/issues/542

### DIFF
--- a/msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
+++ b/msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
@@ -38,7 +38,8 @@ namespace Microsoft.Identity.Client
     /// <summary>
     /// Class to be used for confidential client applications like Web Apps/API.
     /// </summary>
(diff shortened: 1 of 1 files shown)


jmprieur (approved):
LGTM

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
```diff
@@ -38,7 +38,8 @@ namespace Microsoft.Identity.Client
     /// <summary>
     /// Class to be used for confidential client applications like Web Apps/API.
     /// </summary>
-    public sealed class ConfidentialClientApplication : ClientApplicationBase, IConfidentialClientApplication
+    public sealed class
```
henrik-me: >  [](start = 23, length = 1)

This is not needed. I think the width we have on our screens these days are ok with this. to me this is reducing readability

bgavrilMS: Oops, I was playing with .editorconfig to enforce styles and VS changed the formatting. I'll revert.

PR #1168, opened 2018-08-03
High CPU under load fix

mikelapierre:
The default headers should only be changed once so I moved the change to the factory. Otherwise, the following problem occurs under load: https://blogs.msdn.microsoft.com/tess/2009/12/21/high-cpu-in-net-app-using-a-static-generic-dictionary/

### DIFF
--- a/core/src/Http/HttpClientFactory.cs
+++ b/core/src/Http/HttpClientFactory.cs
@@ -26,6 +26,7 @@
 //------------------------------------------------------------------------------
 
 using System.Net.Http;
+using System.Net.Http.Headers;
 
 namespace Microsoft.Identity.Core.Http
 {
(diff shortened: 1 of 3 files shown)


bgavrilMS:
LGTM, but could you please add a unit test. There is an existing class HttpClientFactoryTests - just use the factory to get an HttpClient and assert the headers.

bgavrilMS (changes requested):
See inline, please add a unit test

bgavrilMS (approved):
LGTM

henrik-me (approved):
:shipit:

bgavrilMS:
@mikelapierre - I believe this myget package has your fix https://www.myget.org/feed/aad-clients-nightly/package/nuget/Microsoft.Identity.Client/2.0.0-internal21503dev

PR #1170, opened 2018-08-06
Does not expose the interactive overrides in .NET Core

jmprieur:
Does not expose the interactive overrides in .NET Core of AcquireTokenAsync in .NET Core.

See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/557

### DIFF
--- a/msal/src/Microsoft.Identity.Client/IPublicClientApplication.cs
+++ b/msal/src/Microsoft.Identity.Client/IPublicClientApplication.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Identity.Client
 
         // expose the interactive API without UIParent only for platforms that 
         // do not need it to operate like desktop, UWP, iOS.
-#if !ANDROID && !FACADE
+#if !ANDROID && !NETSTANDARD1_1 && !NETSTANDARD1_3
         /// <summary>
         /// Interactive request to acquire token. 
         /// </summary>
@@ -139,7 +139,7 @@ namespace Microsoft.Identity.Client
 
 #endif
 
-#if !FACADE
+#if !NETSTANDARD1_1 && !NETSTANDARD1_3
         // these API methods are exposed on other platforms.
         /// <summary>
         /// Interactive request to acquire token. 
--- a/msal/src/Microsoft.Identity.Client/IPublicClientApplication.cs
+++ b/msal/src/Microsoft.Identity.Client/IPublicClientApplication.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Identity.Client
 
         // expose the interactive API without UIParent only for platforms that 
         // do not need it to operate like desktop, UWP, iOS.
-#if !ANDROID
+#if !ANDROID && !FACADE
         /// <summary>
         /// Interactive request to acquire token. 
         /// </summary>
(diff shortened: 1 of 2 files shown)


jmprieur:
@bgavrilMS @jennyf19 @henrik-me 
I thought we could do that part of the move to remove WinRT?

this will avoid complains from users that we expose something which is not implemented?

henrik-me:
I agree it's better to hide than to throw a NotSupportedException from somewhere. Aside: Will need to make the same changes to PublicClientApplication.

henrik-me (changes requested):
:clock1:

jmprieur:
dizzy me :) I started, was randomized, and when I got back thought I was done.
This PR breaks the msal\tests\dev apps\XForms\XForms\AcquirePage.xaml.cs as this is .NET standard 1.1

@jennyf19 @bgavrilMS does it need to be .NET standard 1.1 ? or can we do a bit better?
should we have a platform for .NET Core? (and which ?)

jennyf19:
@jmprieur - I don't think the app needs to be 1.1 ...on another note... also need to remove the WebUIFactory from net1.3

---
In reply to: [410724671](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1170#issuecomment-410724671) [](ancestors = 410724671)

bgavrilMS:
@jmprieur - we discussed removing support for netstandard11 in MSAL but we do not have a good picture of other netstandard1.x libraries that take a depdency on it. Maybe it is a good time to get rid of it during the MSAL 2.0.0 preview and bring it back if needed?

jmprieur:
Yes, I was wondering if we should not expose the .NET Core platform (or platforms?). This would be unambiguous, and would enable us to not expose the interactive acquire token API in .NET Core (as this cannot be implemented).
Let's have a brainstorm on that.

jennyf19:
I like this idea.

---
In reply to: [410734281](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1170#issuecomment-410734281) [](ancestors = 410734281)

henrik-me:
Are you proposing even removing netstandard13 ?

msal/src/Microsoft.Identity.Client/IPublicClientApplication.cs
```diff
@@ -46,7 +46,7 @@ public interface IPublicClientApplication : IClientApplicationBase
 
         // expose the interactive API without UIParent only for platforms that 
         // do not need it to operate like desktop, UWP, iOS.
-#if !ANDROID
+#if !ANDROID && !FACADE
```
bgavrilMS: Looks like FACADE is only defined for netstandard1.1; The system automatically defines the compiler variables NETSANDARD1_1 and NETSTANDARD1_3 which are used here and there throughout the code. I think FACADE should be defined for both. Let me know if you want me to pick this up.

jennyf19: I agree FACADE should be defined for both. This would simplify things a lot.

bgavrilMS:
I'm going to close this PR because the current approach will not work for code sharing through netstandard libraries.

PR #1172, opened 2018-08-07
Make WebBrowserNavigateErrorEventHandler internal in both ADAL and MSAL

bgavrilMS:
I tested by running net45 samples with interactive auth


jmprieur:
this is a public API breaking change strickly speaking, but I don't see why they were public in the first place
We should try on all the platform that use it.

bgavrilMS:
@jmprieur - it seems that this handler was public only on net45, for which I have tested; 
I'll make another test pass to verify the public API for other platforms.

jennyf19 (approved):
:shipit: Looks good.

PR #1173, opened 2018-08-07
Moving telemetry to MSAL

trwalke:
Removing Telemetry from core and moving it to MSAL. Core will now use CoreTelemetry which will be injected with either the ADAL or MSAL telemetry implementation on startup. This way, MSAL will have a public telemetry class (MSALTelemetry) and core will have access to only internal telemetry classes (CoreTelemetry, iTelemetry)

VSTS 510046


core/src/OAuth2/OAuth2Client.cs
```diff
@@ -92,8 +88,8 @@ internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, R
 
             HttpResponse response = null;
             Uri endpointUri = CreateFullEndpointUri(endPoint);
-            var httpEvent = new HttpEvent() { HttpPath = endpointUri, QueryParams = endpointUri.Query };
-            Client.Telemetry.GetInstance().StartEvent(requestContext.TelemetryRequestId, httpEvent);
+            //var httpEvent = new HttpEvent() { HttpPath = endpointUri, QueryParams = endpointUri.Query };
+            //Client.Telemetry.GetInstance().StartEvent(requestContext.TelemetryRequestId, httpEvent);
```
henrik-me: ? #Closed

msal/src/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -59,6 +59,14 @@
...
     <Compile Remove="Platforms\**\*.*" />
     <Compile Remove="Features\**\*.*" />
+    <None Remove="Telemetry.cs" />
+    <None Remove="Telemetry\ApiEvent.cs" />
+    <None Remove="Telemetry\CacheEvent.cs" />
+    <None Remove="Telemetry\DefaultEvent.cs" />
+    <None Remove="Telemetry\Event.cs" />
+    <None Remove="Telemetry\HttpEvent.cs" />
+    <None Remove="Telemetry\TelemetryTokenCacheAccessor.cs" />
+    <None Remove="Telemetry\UiEvent.cs" />
```
henrik-me: ? .. Telemetry/*.cs #Closed

msal/src/Microsoft.Identity.Client/Telemetry.cs
```diff
@@ -25,11 +25,11 @@
 //
 //------------------------------------------------------------------------------
 
+using Microsoft.Identity.Client;
```
henrik-me: is this needed? #Closed

henrik-me (changes requested):
:clock1:

core/src/CoreTelemetry.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Core
+{
+    internal class CoreTelemetry
+    {
+        public static iTelemetry Instance;
```
bgavrilMS: please use ITelemetry #Closed

core/src/Microsoft.Identity.Core.csproj
```diff
@@ -127,4 +127,6 @@
       <Version>27.0.2</Version>
     </PackageReference>
   </ItemGroup>
+	  <Target Name="ProjectReferenceSolution" BeforeTargets="ResolveAssemblyReferences" DependsOnTargets="ResolveProjectReferences">
```
bgavrilMS, line 127: What is the purpose of this? #Closed

core/tests/Test.Microsoft.Identity.Core.Unit/Test.Microsoft.Identity.Core.Unit.csproj
```diff
@@ -83,4 +83,6 @@
   <ItemGroup>
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
+  <Target Name="ProjectReferenceSolution" BeforeTargets="ResolveAssemblyReferences" DependsOnTargets="ResolveProjectReferences">
```
bgavrilMS, line 87: same as above #Closed

msal/src/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -59,6 +59,14 @@
     <None Include="**\*.cs;**\*.xml;**\*.axml" Exclude="obj\**\*.*;bin\**\*.*" />
     <Compile Remove="Platforms\**\*.*" />
     <Compile Remove="Features\**\*.*" />
+    <None Remove="Telemetry.cs" />
```
bgavrilMS: You probably do not need these at all. #Closed

msal/src/Microsoft.Identity.Client/MSALTelemetry.cs
```diff
@@ -204,6 +204,21 @@ private List<EventBase> CollateOrphanedEvents(String requestId)
             return orphanedEvents;
         }
 
+        iTelemetry iTelemetry.GetInstance()
+        {
+            return GetInstance();
+        }
+
+        void iTelemetry.StartEvent(string requestId, EventBase eventToStart)
```
bgavrilMS: There is no need to explicitly implement interfaces unless your object implements multiple interfaces and 2 of those interfaces define a method that has the same name. 

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation #Closed

core/src/CoreTelemetry.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Core
+{
+    internal class CoreTelemetry
```
bgavrilMS: I had a look at this and have a few suggestions: 

1. the constructor of CoreTelemetry should take in an ITelemetry param. This way your class is immutable and easier to use. You should also expose properties not members, i.e. public static ITelemetry Instance { get; private set;}
2. I don't like that you have to do CoreTelemetry.Instance.GetInstance() ... Ideally we'd just have CoreTelemetry.Instance
3. The interface ITelemetry should not need to have a GetInstance()
4. It would be cleaner if the classes MSALTelemetry and ADALTelemetry would not be singletons but normal classes with state etc. and you'd use CoreTelemetry everywhere. Not sure how much work this would be though, but if it's not a lot, it's probably worth it. #Closed

msal/src/Microsoft.Identity.Client/Internal/ModuleInitializer.cs
```diff
@@ -56,6 +56,7 @@ public static void EnsureModuleInitialized()
                 if (!isInitialized)
                 {
                     CoreExceptionFactory.Instance = new MsalExceptionFactory();
+                    CoreTelemetry.Instance = new MSALTelemetry();
```
bgavrilMS: Hmm, this is possibly a bug - you are creating a new instance of MSALTelemetry and also a new instance in the MSALTelemetry class. So in fact you have 2 singletons - one in MSAL and one in core for the same object. 

I see 2 ways out of this: 

1. As stated above, refactor MSAL/ADAL Telemetry objects to not be singletons
2. or initialize CoreTelemetry.Instance with MsalTelemetry.Singleton #Closed

core/src/CoreTelemetry.cs
bgavrilMS: Hmm, how about CoreTelemetryService .... this name makes it a bit clearer that it is a singleton... ? #Closed

msal/src/Microsoft.Identity.Client/MSALTelemetry.cs
trwalke: This is needed because the itelemetry interface is internal only as we are trying to remove public classes from core. by implementing the interface explicitly,  it allows itelemetry to be internal and used by core while MSALTelemetry can remain public and not be exposed to core. #Closed

core/src/Microsoft.Identity.Core.csproj
trwalke: This solves one of the visual studio bugs that results in broken project references in multi-targeted solutions. I only put it into test projects as they are the only ones effected. #Closed

core/src/CoreTelemetry.cs
trwalke: #Resolved #Closed

msal/src/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
trwalke: Auto generated by visual studio. I can remove #Closed

core/src/CoreTelemetry.cs
trwalke: I Agree #Closed

trwalke: 1. CoreTelemetry.instance is static and no object reference is available for coreTelemetry. I propose something like public static void InitializeCoreTelemetryService(ITelemetry instance)?
2. #Resolved
3. #Resolved
4. This looks like it can get tricky, i think I will keep it as is now for the sake of time and create a task tracking this. #Closed

msal/src/Microsoft.Identity.Client/Internal/ModuleInitializer.cs
trwalke: #resolved with option 2 #Closed

msal/src/Microsoft.Identity.Client/MSALTelemetry.cs
bgavrilMS: Ah, sneaky... :)
 #Closed

msal/src/Microsoft.Identity.Client/Telemetry.cs
trwalke: No
#Resolved #Closed

msal/src/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
trwalke: Removed
#Resolved #Closed

bgavrilMS (approved):

core/src/Microsoft.Identity.Core.csproj
trwalke: #Resolved #Closed

core/tests/Test.Microsoft.Identity.Core.Unit/Test.Microsoft.Identity.Core.Unit.csproj
trwalke: #Resolved #Closed

core/src/CoreTelemetryService.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Core
+{
+    internal class CoreTelemetryService
```
henrik-me, line 37: mark class as static #Closed

core/src/Telemetry/iTelemetry.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Core
```
henrik-me: Filename seems still to be with small i #Closed

henrik-me:
> using System;

missing file headers #Closed

---
Refers to: core/src/Telemetry/iTelemetry.cs:1 in 1c375ba. [](commit_id = 1c375ba7d7d72aed3b90bdf58a85ea521ad38304, deletion_comment = False)

core/src/Telemetry/CoreTestTelemetryService.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Core
+{
+    internal class CoreTestTelemetryService : ITelemetry
```
henrik-me: >TestT [](start = 23, length = 5)

Why have a test telemetry service here? #Closed

henrik-me:
All new files are missing file headers #Closed

henrik-me:
Why are you moving all the telemetry files out of the telemetry namespace? #Closed

core/src/CoreTelemetryService.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public static void InitializeCoreTelemetryService(ITelemetry instance)
+        {
+            Instance = instance;
```
henrik-me: >instance [](start = 23, length = 8)

pls. ensure this is initialized in the ModuleInitalizer of ADAL as well, as otherwise parts of the code will likely crash because the instance is not set. 
Note: only an issue if the Telemetry class is called in Core... which I assume it is given we have this service.

Aside: in the other classes the Instance is a method and called GetInstance, perhaps we should do the same here for consistency? #Closed

msal/src/Microsoft.Identity.Client/MSALTelemetry.cs
```diff
@@ -55,15 +55,15 @@ public void RegisterReceiver(Receiver r)
             _receiver = r;
         }
 
-        private static readonly Telemetry Singleton = new Telemetry();
+        private static readonly MSALTelemetry Singleton = new MSALTelemetry();
```
henrik-me: >MSALTelemetry [](start = 32, length = 13)

ITelemetry ?  #Closed

msal/src/Microsoft.Identity.Client/MSALTelemetry.cs
```diff
@@ -25,18 +25,18 @@
...
 using System.Linq;
-using Microsoft.Identity.Core.Telemetry;
 
 namespace Microsoft.Identity.Client
 {
     /// <summary>
     /// 
     /// </summary>
-    public class Telemetry
+    internal class MSALTelemetry : ITelemetry
```
henrik-me: >MSALTelemetry [](start = 19, length = 13)

Not sure this should be renamed. that will be another breaking change. The namespace takes care of separation #Closed

henrik-me: >internal [](start = 4, length = 8)

I suppose it has to be public here. How are you exposing this? #Closed

henrik-me:
>         public bool TelemetryOnFailureOnly { get; set; }

Any reason you are not moving these to Core? #Closed

---
Refers to: core/src/Telemetry.cs:74 in 1c375ba. [](commit_id = 1c375ba7d7d72aed3b90bdf58a85ea521ad38304, deletion_comment = False)

msal/src/Microsoft.Identity.Client/MSALTelemetry.cs
henrik-me: >Singleton [](start = 46, length = 9)

personal preference, call this instance #Closed

msal/src/Microsoft.Identity.Client/MSALTelemetry.cs
henrik-me: Why can't the Implicit implementation give the same (not exposed outside the library)? An interface can be internally implemented

---
In reply to: [208693813](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1173#discussion_r208693813) [](ancestors = 208693813)

msal/src/Microsoft.Identity.Client/Internal/ModuleInitializer.cs
```diff
@@ -56,6 +56,7 @@ public static void EnsureModuleInitialized()
                 if (!isInitialized)
                 {
                     CoreExceptionFactory.Instance = new MsalExceptionFactory();
+                    CoreTelemetryService.InitializeCoreTelemetryService(MSALTelemetry.GetInstance());
```
henrik-me: >InitializeCoreTelemetryService(MSALTelemetry.GetInstance()); [](start = 41, length = 60)

Can we do like for the exception factory?

henrik-me:
imo, not ready yet #Closed

trwalke:
"public bool TelemetryOnFailureOnly { get; set; }
Any reason you are not moving these to Core?"
Refers to: msal/src/Microsoft.Identity.Client/MSALTelemetry.cs:74


@henrik-me Do you mean defining these in ITelemetry so that core has access to them? at the moment, only StartEvent() and EndEvent() are used by core.  #Closed

trwalke:
> Why are you moving all the telemetry files out of the telemetry namespace?
@henrik-me Refactoring error, resolving #Closed

core/src/CoreTelemetryService.cs
trwalke: #Resolved #Closed

core/src/Telemetry/iTelemetry.cs
trwalke: #Resolved #Closed

core/src/Telemetry/CoreTestTelemetryService.cs
trwalke: No longer needed 
#Resolved #Closed

core/src/CoreTelemetryService.cs
trwalke: #Resolved #Closed

msal/src/Microsoft.Identity.Client/MSALTelemetry.cs
trwalke: #Resolved #Closed

msal/src/Microsoft.Identity.Client/MSALTelemetry.cs
trwalke: #Resolved #Closed

trwalke: Yes, exposing here 
#Resolved #Closed

msal/src/Microsoft.Identity.Client/MSALTelemetry.cs
trwalke: I agree
#Resolved #Closed

trwalke:
Added missing file headers #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Telemetry.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.Identity.Core.Telemetry;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory
+{
+    internal class Telemetry : ITelemetry
```
henrik-me: >internal class Telemetry : ITelemetry [](start = 4, length = 37)

pls. add comments to the class. and espcially a remarks section explaining why the implementation is blank #Closed

core/src/Telemetry/iTelemetry.cs
henrik-me: Seems still not updated. But this might be a tools issue more than anything.

---
In reply to: [208828746](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1173#discussion_r208828746) [](ancestors = 208828746)

core/src/CoreTelemetryService.cs
henrik-me: not what I meant. Everywhere else we have a mathod called GetInstance. Personally I prefer the property called Instance. Most of our similar code follows the pattern of a method named GetInstance() ... thus property = Intance, or method = GetInstance()

---
In reply to: [208828711](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1173#discussion_r208828711) [](ancestors = 208828711)

msal/src/Microsoft.Identity.Client/Telemetry.cs
```diff
@@ -65,7 +65,7 @@ public void RegisterReceiver(Receiver r)
         /// <returns></returns>
         public static Telemetry GetInstance()
```
henrik-me, line 66: >GetInstance() [](start = 32, length = 13)

implemented as a method... then it's called GetInstance. This is the most used pattern in the codebase. pls. stick to this one #Closed

msal/src/Microsoft.Identity.Client/Telemetry.cs
```diff
@@ -204,6 +204,16 @@ private List<EventBase> CollateOrphanedEvents(String requestId)
...
 
+        void ITelemetry.StartEvent(string requestId, EventBase eventToStart)
+        {
+            StartEvent(requestId, eventToStart);
+        }
+
+        void ITelemetry.StopEvent(string requestId, EventBase eventToStop)
+        {
+            StopEvent(requestId, eventToStop);
+        }
```
henrik-me, line 215: still I do not get why this is needed. it's already implemented implicitly #Closed

henrik-me:
>             Assert.AreEqual(logger, CoreLoggerBase.Default, "Initialization should have happened only once");

seems like you are not asserting that Telemetry is initialized

#Resolved
---
Refers to: msal/tests/Test.MSAL.NET.Unit/ModuleInitializerTest.cs:60 in 549d264. [](commit_id = 549d264798e512526bb169cf5b120aa21cf865dd, deletion_comment = False) #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Telemetry.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        private static readonly ITelemetry Instance = new Telemetry();
+
+        public static Telemetry GetInstance()
+        {
+            return Instance as Telemetry;
+        }
```
henrik-me, line 50: :) #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Telemetry.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        void ITelemetry.StartEvent(string requestId, EventBase eventToStart)
```
henrik-me: >ITelemetry [](start = 13, length = 10)

why explicitly? #Closed

trwalke: @henrik-me 
With the intention of making Telemetry.cs internal, I did not want to publicly expose the new ITelemetry interface to ADAL. Thus, the interface is internal. However, public Telemetry.cs implements the internal ITelemetry interface; the only way that I have found to do that is for the Telemetry to implement the interface explicitly. 
This way, 
1. Customers using MSAL can use the public Telemetry.cs, 
2. Core can still use the iTelemetry interface and have the implementation injected
3. Customers using ADAL will not have access to Telemetry.CS or ITelemetry.cs as neither of the two classes are used in any meaningful way in ADAL.

We can get around the explicit implementation by making ITelemetry public but we will then expose it to ADAL. 
Thoughts? #Closed

msal/src/Microsoft.Identity.Client/Telemetry.cs
trwalke: See other comment
 #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Telemetry.cs
trwalke: #Resolved #Closed

msal/src/Microsoft.Identity.Client/Telemetry.cs
trwalke: #Resolved #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Telemetry.cs
henrik-me: the class is internal, it will not be public. This should not have anything to do with the interface being implicitly implemented. I don't get why it has to be public to be explicitly implemented

---
In reply to: [209065629](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1173#discussion_r209065629) [](ancestors = 209065629)

trwalke: @henrik-me Ah, I understand, I confused this with the other public Telemetry class in MSAL, you are right, it does not need to be implemented explicitly. I will remove #Closed

core/src/CoreTelemetryService.cs
```diff
@@ -0,0 +1,51 @@
...
+using Microsoft.Identity.Core.Telemetry;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Core
+{
+    internal static class CoreTelemetryService
```
henrik-me: >static [](start = 13, length = 6)

... I think you are running into the issues because the class is static... probably that should be removed. #Closed

trwalke: #Resolved #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Telemetry.cs
trwalke: #Resolved #Closed

henrik-me (approved):
:shipit:

PR #1177, opened 2018-08-07
Jennyf/fixes

jennyf19:
Ran into some issues testing w/msal desktop app...addressing those issues here.

### DIFF
--- a/msal/src/Microsoft.Identity.Client/Account.cs
+++ b/msal/src/Microsoft.Identity.Client/Account.cs
@@ -34,12 +34,8 @@ namespace Microsoft.Identity.Client
     /// Contains information of a single account. A user can be present in multiple directorie and thus have multiple accounts.
     /// This information is used for token cache lookup and enforcing the user session on STS authorize endpont.
     /// </summary>
-    internal sealed class Account: IAccount
+    internal sealed class Account : IAccount
     {
-        public Account()
-        {
-        }
-
         public Account(AccountId homeAccountId, string username, string environment)
         {
             if (homeAccountId == null)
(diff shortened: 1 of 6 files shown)


msal/src/Microsoft.Identity.Client/Account.cs
```diff
@@ -61,11 +60,20 @@ public Account(AccountId homeAccountId, string username, string environment)
 
         public override string ToString()
         {
-            return String.Format(
+            if (HomeAccountId != null)
+            {
+                return String.Format(
                 CultureInfo.CurrentCulture,
                 "Account username: {0} environment {1} home account id: {2}",
                 Username, Environment, HomeAccountId.ToString());
```
jmprieur: you don't need the ToString() on HomeAccountId

jmprieur (approved):
LGTM

msal/src/Microsoft.Identity.Client/Account.cs
```diff
@@ -61,11 +60,20 @@ public Account(AccountId homeAccountId, string username, string environment)
 
         public override string ToString()
         {
-            return String.Format(
```
bgavrilMS: It looks like the parameterless ctor is only used by tests. Could we get rid of it instead? This way HomeAccountId would be guaranteed not null.

msal/src/Microsoft.Identity.Client/Account.cs
jennyf19: removed. thanks.

msal/src/Microsoft.Identity.Client/Account.cs
jennyf19: great idea. thanks.

jennyf19: @bgavrilMS I updated all the unit tests.

bgavrilMS (approved):
very cool, LGTM

PR #1180, opened 2018-08-09
ADAL Migration fix enaling ADAL users with no client info in MSAL

henrik-me:
Creating a PR based on Petro's branch. This will enable the null account for ADAL users without client info.


core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -113,10 +113,11 @@ internal class CacheFallbackOperations
 #endif
         }
 
-        public static Dictionary<String, AdalUserInfo> GetAllAdalUsersForMsal(ILegacyCachePersistance legacyCachePersistance,
-            ISet<string> environments, string clientId)
+        public static Tuple<Dictionary<String, AdalUserInfo>, List<AdalUserInfo>> GetAllAdalUsersForMsal
```
bgavrilMS, line 116: I suggest you use the C# 7 style tuples or at least add a comment on this method to explain what it returns. 

https://blogs.msdn.microsoft.com/mazhou/2017/05/26/c-7-series-part-1-value-tuples/

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -824,30 +825,43 @@ internal async Task<IEnumerable<IAccount>> GetAccountsAsync(string authority, bo
...
-                             AccountId.FromClientInfo(clientInfo), 
-                            pair.Value.DisplayableId, 
-                            new Uri(authority).Host);
+                        clientInfoToAccountMap[accountIdentifier] = new Account(
+                             AccountId.FromClientInfo(clientInfo), pair.Value.DisplayableId, environment);
                     }
                 }
 
-                return allUsers.Values;
+                var accounts = clientInfoToAccountMap.Values;
```
bgavrilMS, line 852: NIT: This method is quite long, consider breaking it up a bit

bgavrilMS (approved):

core/src/Cache/CacheFallbackOperations.cs
henrik-me: I believe Petro couln't get that to work in his previous PR for some reason. It might be because this is not yet available for all platforms?

bgavrilMS: Ah ok, makes sense. net45 probably does not support this (you might need to take a dependency on a NuGet package for this... not worth the trouble).

henrik-me: VS: Predefined type "System.ValueTuple'2" is not defined or imported

PR #1182, opened 2018-08-09
Update netstandard1.3 exception and exception message for WebUIFactory class

jennyf19:
Did a quick test w/desktop


msal/src/Microsoft.Identity.Client/Internal/PlatformPlugin.cs
```diff
@@ -50,8 +50,8 @@ private static void InitializeWebFactoryAndPlatform()
```
bgavrilMS: Since WebUIFactory is internal, is this really needed?

msal/src/Microsoft.Identity.Client/Platforms/netstandard1.3/WebUIFactory.cs
```diff
@@ -1,41 +0,0 @@
...
-using Microsoft.Identity.Core;
-using Microsoft.Identity.Core.UI;
-
-namespace Microsoft.Identity.Client.Internal.UI
-{
-    internal class WebUIFactory : IWebUIFactory
-    {
-        public IWebUI CreateAuthenticationDialog(CoreUIParent parent, RequestContext requestContext)
-        {
-            throw new NotImplementedException();
```
bgavrilMS: The user experience for before you removed this was that if I call this library from a platform we do not fully support, for example WinPhone7 / NetCore or from a XForms app where I forget to add a reference to MSAL from XForms.UWP, I'd get a this exception. 

What is the user experience now?

I think it would be better to leave this as it is, and instead to throw a PlatformNotSupported exception stating: 

"Possible Cause: If you are using an XForms app, or generally a netstandard assembly, make sure you add a refernce to Microsoft.Identity.Client.dll from each platform assembly (e.g. UWP, Android, iOS), not just from the common netstandard assembly" #Closed

jmprieur: yes, I agree. this would be much more actionable. #Closed

jennyf19: sounds good. done. thank you.

bgavrilMS (approved):

msal/src/Microsoft.Identity.Client/Internal/PlatformPlugin.cs
jennyf19: i reverted back and used your idea for an error message for a platformNotSupported exception.

msal/src/Microsoft.Identity.Client/Platforms/netstandard1.3/WebUIFactory.cs
henrik-me: Makes sense. At least if we add that additional info. #Closed

PR #1183, opened 2018-08-09
fix autosize so the entire app appears and bottom elements are not cu…

jennyf19:
…t off


henrik-me (approved):
:shipit:

henrik-me:
Adding @rayluo as he has a change in the same file. I suppose reverting the changes to program.cs for this PR will mitigate that.

bgavrilMS (approved):

jennyf19:
completed in another PR due to issues w/VS

PR #1185, opened 2018-08-10
Support Username Password grant in AdalDesktopTestApp

rayluo:
It was used to complete a newly added test case in our test plan. Also the test steps in VSTS is also updated accordingly.

This PR is not a roadblock for the upcoming release.


henrik-me:
@jennyf19 has a change in the same file. It might be easiest to revert Jenny's change to program.cs
#1183

henrik-me (approved):
:shipit:

PR #1187, opened 2018-08-10
Update nuspec files to include the latest metadata

bgavrilMS:


adal/build/Microsoft.IdentityModel.Clients.ActiveDirectory.nuspec
```diff
@@ -18,17 +18,53 @@ ADAL.NET takes advantage of Windows Server Active Directory and Azure Active Dir
...
-			<dependency id="NETStandard.Library" version="1.6" />
-			<dependency id="System.Runtime.Serialization.Json" version="4.3.0" />
-			<dependency id="System.Runtime.Serialization.Primitives" version="4.3.0" />
-		</group>
-		<group targetFramework="netcore45"/>
-		<group targetFramework="net45"/>
-		<group targetFramework="portable-net45+win8"/>
-		<group targetFramework="MonoAndroid10"/>
-		<group targetFramework=" Xamarin.iOS10"/>
+ <dependencies>
```
bgavrilMS, line 21: These are used for display purposes in NuGet.  I generated them by running msbuild /t:pack (which unfortunately we still cannot use to pack core.dll).

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Microsoft.Identity.Core.csproj
```diff
@@ -1,147 +0,0 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
```
bgavrilMS, line 1: Misplaced file, a copy core.csproj should not be in ADAL dir.

jennyf19 (approved):
🚀

PR #1191, opened 2018-08-13
Update nuspec tags

bgavrilMS:
NuGet tags are single words, with no commas. MSAL does it right if you'd like to compare:


jennyf19 (approved):
👾

PR #1192, opened 2018-08-13
Remove the ADAL platform dll

bgavrilMS:
This is not used in ADAL v4 and removing it is trivial (some minor changes will be required for the build)


jennyf19 (approved):
awesome. 🐈

bgavrilMS:
Build will fail because I renamed the sln file ....

henrik-me:
I'm ok with this change as right after we have shipped our current bits. Perhaps the the CoreAndUtests.sln file should also be renamed. NoSamples or LibsAndUtests?

henrik-me (approved):
:shipit:

jmprieur:
why did you remove the branch @bgavrilMS ? it would be useful later?

PR #1195, opened 2018-08-14
Tweaking the MSAL 2.0.0-preview library to provide actionable migration help

jmprieur:
This PR proposes to tweak the MSAL 2.0.0-preview library to **provide actionable migration help** for developers who want to migrate from MSAL 1.x to MSAL 2.0.0-preview. 

## Why? the migration experience if we don't do anything
Updating the MSAL NuGet package to MSAL.NET 2.0.0-preview, developers will get only a few build errors that IUser cannot be found. If they have read the blog post, they'll probably realize that they need to use IAccount instead, but otherwise, they might not understand what happens ; they might think that I’m missing an assembly. Note the tooltips are not actionable (are you missing an assembly). developers can see that other members have red squiggles, but there are no errors for that. this is confusing
image

Then, after some research, they might find that they need to use IAccount and rebuild, and this time have 12 errors. this becomes a trial/error process with no idea of the time it will take to upgrade.

## The experience proposed by this PR
This PR proposes to help the user migrate by providing **more actionable error messages, and links to the blog post**, the developer experience becoming:
-	All the location to change would be presented immediately. This might be a bit overwhelming, but at least this enables the developer to predict the amount of work to do.
-	The obsolete types / members are explicated with actionable error messages: advice on what to use instead
-	Finally the tooltips on the error messages in Visual Studio are even more actionable, explaining what happens (that’s in the XML comments), and leading the developer to the blog post, which therefore cannot be missed.

image

image

## How this is done?
This is done by:
- providing the missing types, or members, with an [Obsolete] attribute and some meaningful and actionable message.
- providing an aka.ms link to the blog post.

The only change to the product code is to set 4 interfaces and classes as partial, and provide partial implementations with obsolete attributes in the MigrationAid.cs file (which of course could be split into several files, but why bother).

### DIFF
--- a/msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
+++ b/msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Identity.Client
     /// <summary>
     /// Contains the results of one token acquisition operation.
     /// </summary>
-    public class AuthenticationResult
+    public partial class AuthenticationResult
     {
         private const string Oauth2AuthorizationHeader = "Bearer ";
         private readonly MsalAccessTokenCacheItem _msalAccessTokenCacheItem;
--- a/msal/src/Microsoft.Identity.Client/ClientApplicationBase.cs
+++ b/msal/src/Microsoft.Identity.Client/ClientApplicationBase.cs
@@ -42,7 +42,7 @@ namespace Microsoft.Identity.Client
     /// <Summary>
     /// Abstract class containing common API methods and properties. Both PublicClientApplication and ConfidentialClientApplication extend this class.
     /// </Summary>
-    public abstract class ClientApplicationBase
+    public abstract partial class ClientApplicationBase
     {
         private TokenCache userTokenCache;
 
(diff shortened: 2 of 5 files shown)


bgavrilMS (approved):
LGTM

trwalke:
Very good idea @jmprieur. 
LGTM

jennyf19 (approved):
This is excellent. thank you. 😸

henrik-me:
Looks awesome. Do we have any of the confidential flows where we want to add something similar?

henrik-me (approved):
:shipit:

jmprieur:
This should cover confidential flows as well as the partial types are IClientApplication**Base**  and ClientApplication**Base** ?

2 questions are:
- do we want to do the same for Logger => MsalLoggerSettings ? 
- do we want to have the aka.ms link in the obsolete attribute message?

jennyf19:
@jmprieur - to answer your questions, from my perspective: 1) yes 2) yes

rayluo (approved):
Cool!

jmprieur:
I've added another commit to add the aka.ms link into the obsolete message.
I won't do the aid for the Logger since we decided to re-rename it.

PR #1196, opened 2018-08-14
Use the same public key to sign all assemblies

bgavrilMS:


build/AssemblySigningVerification.targets
```diff
@@ -39,6 +39,9 @@
   </Target>
 
   <Target Name="LocateSnExe">
+
+     <Message Text="!!!! AssembliesToCheck @(AssembliesToCheck)" Importance="high"/>
```
bgavrilMS: I'll remove this.

core/src/Properties/InternalsVisibleTo.cs
```diff
@@ -28,14 +28,14 @@
...
-[assembly: InternalsVisibleTo("AutomationApp, PublicKey=00240000048000009400000006020000002400005253413100040000010001002D96616729B54F6D013D71559A017F50AA4861487226C523959D1579B93F3FDF71C08B980FD3130062B03D3DE115C4B84E7AC46AEF5E192A40E7457D5F3A08F66CEAB71143807F2C3CB0DA5E23B38F0559769978406F6E5D30CEADD7985FC73A5A609A8B74A1DF0A29399074A003A226C943D480FEC96DBEC7106A87896539AD")]
-[assembly: InternalsVisibleTo("DesktopTestApp, PublicKey=00240000048000009400000006020000002400005253413100040000010001002D96616729B54F6D013D71559A017F50AA4861487226C523959D1579B93F3FDF71C08B980FD3130062B03D3DE115C4B84E7AC46AEF5E192A40E7457D5F3A08F66CEAB71143807F2C3CB0DA5E23B38F0559769978406F6E5D30CEADD7985FC73A5A609A8B74A1DF0A29399074A003A226C943D480FEC96DBEC7106A87896539AD")]
+[assembly: InternalsVisibleTo("Microsoft.Identity.Client, PublicKey=0024000004800000940000000602000000240000525341310004000001000100B5FC90E7027F67871E773A8FDE8938C81DD402BA65B9201D60593E96C492651E889CC13F1415EBB53FAC1131AE0BD333C5EE6021672D9718EA31A8AEBD0DA0072F25D87DBA6FC90FFD598ED4DA35E44C398C454307E8E33B8426143DAEC9F596836F97C8F74750E5975C64E2189F45DEF46B2A2B1247ADC3652BF5C308055DA9")]
```
bgavrilMS, line 34: This is the public key corresponding to the snk file

jmprieur: Is that a breaking change?

bgavrilMS: I used ADAL's key to sign MSAL, so it isn't a breaking change for ADAL users. When you take a dependency on MSAL, it is a breaking change because its public key token has changed.

bgavrilMS: Should be transparent to the user for if they just use NuGet...

msal/src/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -181,6 +173,16 @@
     <Copy SourceFiles="@(FilesToPack)" DestinationFiles="$(NugetPackageRootFolder)\lib\%(RecursiveDir)%(FileName)%(Extension)" />
     <Copy SourceFiles="$(NuspecFilePath)" DestinationFolder="$(NugetPackageRootFolder)" />
   </Target>
+
+  <Target Name="VerifySignatures">
+    <ItemGroup>
+      <AssembliesToCheck Include="bin\$(Configuration)\**\Microsoft.Identity.Client.dll" />
+      <AssembliesToCheck Include="bin\$(Configuration)\**\Microsoft.Identity.Core.dll" />
+    </ItemGroup>
+     <Message Text="AssembliesToCheck @(AssembliesToCheck)" Importance="high" />
```
henrik-me: ><Message Text="AssembliesToCheck @(AssembliesToCheck)" Importance="high" />  [](start = 5, length = 76)

remove?

henrik-me (approved):
:shipit:

PR #1199, opened 2018-08-14
Supressing exceptions in cache fall back logic

SomkaPe:
[Resilience: Ensure AquireToken doesn't fail if cache persistence fails](https://identitydivision.visualstudio.com/DevEx/_workitems/edit/395064)


henrik-me:
pls. create a wiki page and an aka.ms link to that wiki page where we can start explaining why these errors could occur #Closed

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -63,54 +63,70 @@ internal class CacheFallbackOperations
...
+                tokenCacheAccessor.SaveRefreshToken(rtItem);
 
-            tokenCacheAccessor.SaveAccount(accountCacheItem);
+                MsalAccountCacheItem accountCacheItem = new MsalAccountCacheItem
+                    (new Uri(authority).Host, objectId, resultWrapper.RawClientInfo, null, displayableId, resultWrapper.Result.TenantId);
+                tokenCacheAccessor.SaveAccount(accountCacheItem);
+            }
+            catch (Exception ex)
+            {
+                msg = "An error occurred while writing MSAL refresh token.";
```
henrik-me: >"An error occurred while writing MSAL refresh token."; [](start = 22, length = 54)

see comments below #Closed

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -63,54 +63,70 @@ internal class CacheFallbackOperations
...
-            dictionary[key] = wrapper;
-            legacyCachePersistance.WriteCache(AdalCacheOperations.Serialize(dictionary));
-#endif
+                IDictionary<AdalTokenCacheKey, AdalResultWrapper> dictionary = AdalCacheOperations.Deserialize(legacyCachePersistance.LoadCache());
+                dictionary[key] = wrapper;
+                legacyCachePersistance.WriteCache(AdalCacheOperations.Serialize(dictionary));
+            }
+            catch (Exception ex)
+            {
+                string msg = "An error occurred while writing ADAL refresh token.";
```
henrik-me: >h token." [](start = 73, length = 9)

String suggestion:
ADAL-00000001 : Continuing with-out writing the token to the cache, as an error happened while writing the ADAL refresh token. Read more: aka.ms/errorswritingcache-adal.

Consider putting in a unique error id in the message that customers can use to search on the page.

Consider following this format in the other error messages.

This will mean putting these error messages in a separate file clearly numbering the errors.

This will make it easier for developers to search for an answer to their error condition using Google/Bing etc. #Closed

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -276,45 +292,57 @@ internal class CacheFallbackOperations
...
+                                    };
+                                }
                             }
                         }
                     }
                 }
             }
+            catch (Exception ex)
+            {
+                string msg = "An error occurred while searching for MSAL cache entries for ADAL.";
```
henrik-me: >"An error occurred while searching for MSAL cache entries for ADAL."; [](start = 29, length = 69)

pls. see comments above. #Closed

henrik-me:
few minor comments. #supportability #Closed

henrik-me (changes requested):
:clock1:

bgavrilMS (commented):
LGTM with Henrik's comments

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: The reason to do Suppressing exceptions in cache fall back logic  is to reduce impact of possible issues which might happen during cross library cache read write operations, so it is not part of library API , the intention is to provide more details for developers in logs to perform troubleshooting. #Closed

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: not sure what do you mean under ADAL-00000001, also "Continuing with-out writing the token to the cache" is not very accurate because this error is error of sharing RT in another format , so at this  point token already should be stored in "main" cache format.
I have updated messages. #Closed

henrik-me: 1: The ID
What I mean by the ID in front of the message it to provide a unique id for every error message we show. Thus, pls. create a data class with error id's and one with error messages. For now these can all be internal. Use those for all the new messages you are adding. Ideally you would do the same to all the messages in this file. 

2: Ensure the message indicates that it's not blocking
The message should be as precise as possible, what I was indicating is that you should add to the message that even though this is an error situation, then the situation was not throwing any exceptions and thus execution flow is not interrupted by this situation.

3: Please add and aka.ms link to the errors and ensure the link is added to the message in the log or messages sent to customers. e.g. aka.ms/msal_cache_persistance_errors.

Please stop by for additional clarifications.

---
In reply to: [210447020](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1199#discussion_r210447020) [](ancestors = 210447020)

jmprieur:
suggestion, the aka.ms link should rather be **https://aka.ms/msal-net-cache-persistance-errors** for consistency with the other aka.ms links for MSA.NET. #Closed

henrik-me:
sure. 

---
In reply to: [413440103](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1199#issuecomment-413440103) [](ancestors = 413440103)

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: 1. Do you mean for every possible log message ?
    These messages are not part of api they are log messages.
2. In this cases exception are not thrown, not sure why we want to say that something was not done, especially taking to account that warning message level is used 
3.  aka.ms/msal_cache_persistance_errors - not sure what to add for these cases taking to account  that we catch general exception - anything can be there  #Closed

henrik-me: Context: All of these log messages are happening due to some exception!

Consider that someone sends us the log we will search for the root cause in the code. ... Now consider that the user is able to lookup this information in the wiki... the developer will no longer have to open an issue or a support case to find all the information needed. The page linked to should just explain a bit about the cases where things can go wrong. If each log message/exception has an id then it will be easy for a developer to find more information about the root cause. 

Most importantly though here is to ensure each of these log messages also clearly indicates what kind of issues this means for the developer. e.g. not being able to persist a token means they will not be able to do silent authentication, thus they will always have to show ui to the user.

IMO: We have to start somewhere to deliver better error messages. And these I consider error conditions even though those are log messages as these messages are only logged if some exception happens.

---
In reply to: [210739714](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1199#discussion_r210739714) [](ancestors = 210739714)

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -63,54 +63,72 @@ internal class CacheFallbackOperations
...
 
-            tokenCacheAccessor.SaveAccount(accountCacheItem);
+                MsalAccountCacheItem accountCacheItem = new MsalAccountCacheItem
+                    (new Uri(authority).Host, objectId, resultWrapper.RawClientInfo, null, displayableId, resultWrapper.Result.TenantId);
+                tokenCacheAccessor.SaveAccount(accountCacheItem);
+            }
+            catch (Exception ex)
+            {
+                msg = "An error occurred while writing ADAL refresh token to the cache in MSAL format. " +
+                      "For details please see https://aka.ms/net-cache-persistence-errors. ";
```
henrik-me, line 79: >net-cache-persistence-errors [](start = 61, length = 28)

Jean-Marc suggested: 
msal-net-cache-persistance-errors 
for naming consistency.

henrik-me: based on conversation today that the message is used across libraries. approving

henrik-me (approved):

PR #1207, opened 2018-08-17
Prevents the fake properties from appearing in the debugger.

jmprieur:
We've added fake properties to help developers to migrate from MSAL 1.x to MSAL 2.x
however they appear under the debugger, and since they throw, they are highly visible with a red error icon
image

This changes prevents them from appearing in the debugger (where they should never have been visible in the first place), in the same way we hid them from the IntelliSense.


bgavrilMS (approved):
Good catch, LGTM

jennyf19 (approved):
awesome :shipit:

PR #1210, opened 2018-08-20
Minor changes to public API documentation and include warnings

bgavrilMS:
This resolves the 2 new build warnings we've been getting.

### DIFF
--- a/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
+++ b/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
@@ -154,26 +154,26 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
         }
 
 #if iOS
+
+        private string keychainSecurityGroup;
+
         /// <summary>
         /// Xamarin iOS specific property enables the application to share the token cache with other applications sharing the same keychain security group.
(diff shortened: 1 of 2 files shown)


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -154,26 +154,26 @@ public Guid CorrelationId
         }
 
 #if iOS
+
+        private string keychainSecurityGroup;
+
```
bgavrilMS, line 159: It is not clear from the diff, but the comments were placed above the property

jennyf19 (approved):
🐑

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/StorageDelegates.cs
```diff
@@ -26,15 +26,7 @@
...
 
 using Microsoft.Identity.Core.Cache;
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 
-#if ANDROID || iOS || WINDOWS_APP
-using Microsoft.Identity.Core.Cache;
```
bgavrilMS, line 36: double include

PR #1216, opened 2018-08-21
Allow accounts with null ID to enable ADALv3 migration

bgavrilMS:


bgavrilMS:
Note that I am still looking at writing a unit test, but this should unblock E2E testing

jennyf19 (approved):
:shipit: Nice catch

msal/src/Microsoft.Identity.Client/IAccount.cs
```diff
@@ -53,7 +53,7 @@ public interface IAccount
         /// <summary>
         /// AccountId of the home account for the user. This uniquely identifies the user across AAD tenants. 
         /// </summary>
-        /// <remarks>Can be null</remarks>
+        /// <remarks>Can be null, for example with users migrating from ADAL v3</remarks>
```
jmprieur: Can be null, for example if this account was migrated to MSAL.NET from ADAL.NET v3's token cache

jmprieur (approved):
LGTM. 
proposed a slight change in the XML comment

henrik-me (approved):
:shipit:

PR #1220, opened 2018-08-22
Include code files directly instead of of core.dll

bgavrilMS:


bgavrilMS:
A few notes: 

1. I've removed ADAL.Platform dll because it would have been more work to include core in it; we were going to remove it anyway;
2. The builds will fail because I renamed the sln files
3. I will add a readme.md in Core to explain how to work with shared files. 
4. I strongly advise to clone this branch and try it out locally. Try out experiences such as: 

- adding a new code file to core
- adding a platform specific code file to core
- running unit tests

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Microsoft.IdentityModel.Clients.ActiveDirectory.csproj
```diff
@@ -148,4 +162,21 @@
...
+
+  <Target Name="LayoutNuGet">
+    <PropertyGroup>
+      <NugetPackageRootFolder Condition="'$(NugetPackageRootFolder)' == ''">$([System.IO.Path]::Combine($(MSBuildThisFileDirectory), "..",".nugetlayout"))</NugetPackageRootFolder>
+      <NuspecFilePath>$([System.IO.Path]::Combine($(MSBuildThisFileDirectory), "..", "..", "build", "Microsoft.Identity.Client.nuspec"))</NuspecFilePath>
+    </PropertyGroup>
+    <ItemGroup>
+      <FilesToPack Include="bin\$(Configuration)\**\Microsoft.IdentityModel.Clients.ActiveDirectory.dll" />
+      <FilesToPack Include="bin\$(Configuration)\**\Microsoft.IdentityModel.Clients.ActiveDirectory.xml" />
+      <FilesToPack Include="bin\$(Configuration)\**\Microsoft.IdentityModel.Clients.ActiveDirectory.pdb" />
```
henrik-me, line 175: ><FilesToPack Include="bin\$(Configuration)\**\Microsoft.IdentityModel.Clients.ActiveDirectory.pdb" /> [](start = 6, length = 101)

This is only added in debug, or?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationContinuationHelper.cs
```diff
@@ -41,7 +41,6 @@ public static class AuthenticationContinuationHelper
         /// Returns if the response is from the broker app
         /// </summary>
         /// <param name="sourceApplication">application bundle id</param>
-        /// <returns></returns>
```
henrik-me, line 44: >/// <returns></returns> [](start = 8, length = 23)

should probably have a meaningful text instead. 
E.g.
True if the response is from broker, False otherwise.

henrik-me: Assuming this is included in the dll which remains

---
In reply to: [212023486](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1220#discussion_r212023486) [](ancestors = 212023486)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Microsoft.IdentityModel.Clients.ActiveDirectory.csproj
bgavrilMS: No, you can have PDB files in Release as well, although I am not sure about the debugging experience. In any case, the packaging task will create 2 packges - one with the PDBs and one without. NuGet should add support for symbol packages at some point.

core/src/Properties/InternalsVisibleTo.cs
```diff
@@ -27,20 +27,5 @@
 
 using System.Runtime.CompilerServices;
 
-//core test project
+// The core test project is the only one allowed to reference Core.dll for convenience. Do not add any other entries here.
```
henrik-me, line 30: >Do not add any other entries here. [](start = 88, length = 34)

:)

core/tests/Test.Microsoft.Identity.Core.Unit/Test.Microsoft.Identity.Core.Unit.csproj
```diff
@@ -79,7 +79,6 @@
     <PackageReference Include="MSTest.TestFramework" Version="1.3.1" />
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="..\..\..\msal\src\Microsoft.Identity.Client\Microsoft.Identity.Client.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.Identity.Core.csproj" />
```
henrik-me, line 82: ><ProjectReference Include="..\..\src\Microsoft.Identity.Core.csproj" /> [](start = 4, length = 71)

Not sure this line makes sense to me. Does that still exist as a separate project? I thought it was now reduced to a part of each of the libraries?

msal/src/Microsoft.Identity.Client/Account.cs
```diff
@@ -40,6 +40,8 @@ internal sealed class Account : IAccount
         /// Constructor
         /// </summary>
         /// <param name="homeAccountId">an object and tenant scoped id; can be null, for example when migrating the ADAL v3 cache</param>
+        /// <param name="username">UPN style , can be null</param>
+        /// <param name="environment">identity provider for this account, e.g. <c>login.microsoftonline.com</c></param>
```
henrik-me: >i [](start = 38, length = 1)

nit: capital I

msal/src/Microsoft.Identity.Client/Account.cs
```diff
@@ -40,6 +40,8 @@ internal sealed class Account : IAccount
         /// Constructor
         /// </summary>
         /// <param name="homeAccountId">an object and tenant scoped id; can be null, for example when migrating the ADAL v3 cache</param>
```
henrik-me: >a [](start = 40, length = 1)

nit: capital A

henrik-me:
Not for now, but should it be considered moving to the "original" nuget packager now that we are back to having one dll per project?

henrik-me (approved):
:shipit:

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Microsoft.IdentityModel.Clients.ActiveDirectory.csproj
henrik-me: Yeah, however do we generate it today during build? I don't think we do? We should have symbols for release bits though. Those should be on the public symbols server though (which I believe they are).

core/src/SharingCode.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Authoring code and tests
+
+If you add new code to core, you'll need to copy any <PackageReference/> to ADAL and MSAL (you'll get errors if you don't anyway).
+
+If you add or remove  platform specific code to core, outside the Platform directories, you'll need to update the ADAL and MSAL csproj for the conditional compilation.
+
+Tests run as normal, and core.dll has associated unit tests.
+
+### Exposing types from Core
```
henrik-me, line 48: IMO: We should not do this. We should prevent it.

core/src/SharingCode.md
```diff
@@ -0,0 +1,79 @@
...
+### How code sharing works
+
+We keep core.dll as a container for the code files. Both ADAL and MSAL include the code files directly, and do NOT reference core.dll:
+
+```
+<!-- LinkBase shows up as a directory in SolutionExplorer for all the included files -->
+<Compile Include="..\..\..\core\src\**\*.cs"  LinkBase="Core" />
+```
+
+The unified solution still contains a Core.dll with associated unit tests. However, you should not take a dependency on Core.dll from anywhere else (you'll get duplicated item exceptions anyway).
```
henrik-me, line 38: >However, you should not take a dependency [](start = 75, length = 41)

Can this be prevented somehow?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Microsoft.IdentityModel.Clients.ActiveDirectory.csproj
bgavrilMS: We do generate PDB files through build and we publish them to internal Microsoft server. Still trying to figure out how to publish them to a public server, perhaps @brentschmaltz knows more?

core/src/SharingCode.md
bgavrilMS: We discussed about moving the core files to ADAL or MSAL, but as a separate PR. Alternatively, we can add a dependency analyzer, I believe @jmprieur might have PM-ed one of those in his career :godmode:

PR #1228, opened 2018-08-27
few small spelling fixes

jennyf19:
I noticed there is a Resource file and resource.cs ...do we need that?


jmprieur (approved):
LGTM. Thanks!

PR #1229, opened 2018-08-27
Jmprieur/fix msal deserialize regression

jmprieur:
proposing a fix for MSAL.NET GitHub issue: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/603,
this is affecting a number of Web App samples (which arguably were not testing that the bytes returned from a session cache were null)


jennyf19 (approved):
looks good 
:shipit:

henrik-me:
as mentioned in the issue. this is not the right fix.

PR #1232, opened 2018-08-27
In .NET FW (4.6.2) or .NET Core (2, 2.1), the PlatformInformationBase…

jmprieur:
In .NET FW (4.6.2) or .NET Core (2, 2.1), the `PlatformInformationBase`  static constructor is not called in a program like the following:

````CSharp
       static void Main(string[] args)
        {
            TokenCache tokenCache = new TokenCache();
            tokenCache.Deserialize(null);
       }
````
hence an ArgumentNullException when any code wants to access `CorePlatformInformationBase.Instance`.

This branch is an attempt to initialize it (and this solves the issue. Not sure this is the right fix)

### DIFF
--- a/msal/src/Microsoft.Identity.Client/Internal/PlatformInformationBase.cs
+++ b/msal/src/Microsoft.Identity.Client/Internal/PlatformInformationBase.cs
@@ -34,10 +34,11 @@ namespace Microsoft.Identity.Client.Internal
 {
     internal abstract class PlatformInformationBase : CorePlatformInformationBase
     {
-           
+        public static bool Initialized {get;} = false;
         static PlatformInformationBase()
         {
             Instance = new PlatformInformation();
+            Initialized = true;
         }
 
         public override string GetAssemblyFileVersionAttribute()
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: msal/src/Microsoft.Identity.Client/Internal/PlatformInformationBase.cs [hunk 1]
```diff
@@ -34,10 +34,11 @@ namespace Microsoft.Identity.Client.Internal
 {
     internal abstract class PlatformInformationBase : CorePlatformInformationBase
     {
-           
+        public static bool Initialized {get;} = false;
         static PlatformInformationBase()
         {
```
henrik-me, line 39: Following our current pattern most likely we should ensure ModuleInitizer is called from the PlatformInformationBase static contructor instead of having this two way initializer

jmprieur: here the goal of this proposal (which I'm not too happy with) is to force the static constructor of `PlatformInformationBase` to be called (as we know that the static constructor of `ModuleInitializer` is called) so that we are sure that the following instruction get called:

`` Instance = new PlatformInformation(); ``

as currently it can be not called because stuff interact with the base class (CorePlatformInformationBase) and not the derived class? 

I don't understand how ensuring `ModuleInitializer` be called from `PlatformInformationBase` would change the problem?

jmprieur: @henrik-me : maybe call PlatformInformationBase static constructor in PlaformPlugIn ?

jmprieur:
Closing as #1233 is much better

PR #1233, opened 2018-08-27
Jennyf/logger issue in b2 cpm sample for msal with object null reference help

henrik-me:
Ensuring the PlatformInformation is initialized. Still a few more things to validate however sharing allowing additional feedback.


henrik-me:
aside: the reference to the sample app should be removed.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ModuleInitializer.cs
```diff
@@ -59,6 +59,9 @@ public static void EnsureModuleInitialized()
                 {
                     CoreLoggerBase.Default = new AdalLogger(Guid.Empty);
                     CoreTelemetryService.InitializeCoreTelemetryService(Telemetry.GetInstance() as ITelemetry);
+                    // Several statics in the library depends on platform information being timely initialized. 
+                    // The static initializer on PlatformInformationBase will ensure this gets done.
+                    new PlatformInformation();
```
jmprieur, line 64: This looks like the right fix to me. I had not realized ADAL.NET was also affected.

jmprieur (approved):
LGTM. Elegant.
I closed #1232

core/src/MsalIdHelper.cs
```diff
@@ -68,6 +68,10 @@ internal static class MsalIdHelper
     {
         public static IDictionary<string, string> GetMsalIdParameters()
         {
+            if (CorePlatformInformationBase.Instance == null)
+            {
+                throw CoreExceptionFactory.Instance.GetClientException("someErrorCode", "Platform Not Supported");
```
bgavrilMS: nit: would be helpful to put the platform here as well, platform information has it.

msal/tests/Test.NetCore21.Unit/Test.MSAL.NET.dotnetcore.Unit.csproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
bgavrilMS: very cool!

bgavrilMS (approved):

core/src/MsalIdHelper.cs
henrik-me: yeah, this whole method needs updating including the class calling it. So many re-initializations, wastes CPU at very little Memory cost. Considering whether I want to add that here or not. This part is in Core though, which doesn't have access to the full PlatformInformation class.

msal/tests/Test.NetCore21.Unit/Test.MSAL.NET.dotnetcore.Unit.csproj
jennyf19: I agree...this is great.

---
In reply to: [213117773](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1233#discussion_r213117773) [](ancestors = 213117773)

jennyf19 (approved):
I like the changes! 🚀

henrik-me:
Merging to dev. Will have to work more on getting the .netcore tests to run on VSTS.

PR #1238, opened 2018-08-29
adding required claims to error message

jennyf19:
@henrik-me asked that we include the missing claims in the log message

### DIFF
--- a/msal/src/Microsoft.Identity.Client/TokenCache.cs
+++ b/msal/src/Microsoft.Identity.Client/TokenCache.cs
@@ -27,6 +27,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.Internal;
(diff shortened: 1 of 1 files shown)


msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -276,7 +276,7 @@ private bool ContainsClaimsRequiredToStoreInCache(IdToken idToken, RequestContex
                 }
                 else
                 {
-                    msg = "Skipping storing tokens in the cache - IdToken does not contain required claims";
+                    msg = "Skipping storing tokens in the cache - IdToken does not contain required claims of TenantId or PreferredUsername.";
```
henrik-me: >of TenantId or PreferredUsername [](start = 107, length = 32)

we should use the actual names used in the claims. ideally we have const somewhere with these in that we can use. (the const is where the cache is read/persisted)

jennyf19: done. thx. ⛄️ #!

jmprieur (approved):
LGTM

henrik-me:
We should also add a link to the wiki page talking about work arounds and issues.

looks great

bgavrilMS (approved):

PR #1242, opened 2018-08-31
Fix XML comments warnings:

jmprieur:
See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/607

**Note that**
I did not fix the XML comment warning in `UserAssertion.cs`, as I think that `UserAssertion` should only be in the platforms that define `ConfidentialClientApplication` OBO


bgavrilMS (approved):
LGTM

PR #1243, opened 2018-08-31
Unit test refactoring for core tests and initialization fix

bgavrilMS:
Note: this is in preparation of having platform specific unit tests. This PR can be reviewed and merged on its own. 

1. Reference core tests in ADAL / MSAL tests. This has a few advantages: 

a. it matches the structure of the product code
b. it avoids estranged objects, e.g. TokenCacheAccessor exists in MSAL but also in core (as a convenience). If a test method in CoreTests like Foo(TokenCacheAccessor) exists, it will use the type from core - meaning that the method cannot be used in MSALTests

2. Fix initialization issues with tests - because tests have access to the internal structure of ADAL / MSAL, they should re-initialize the module (factories, statics etc.) - otherwise things like CorePlatformInformation is initialized either with net45PlatformInformation or with testPlatformInformation, depending on what test runs first.


LibsNoSamples.sln
```diff
@@ -21,6 +21,12 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test.ADAL.NET.Integration",
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "tests", "tests", "{5D231979-0A13-456D-B9F9-CED41F935FD9}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "adal", "adal", "{8E9E2563-C457-40EF-9D12-CAE51B5DFEBD}"
```
bgavrilMS, line 24: Added a few directories here.

core/src/MsalIdHelper.cs
```diff
@@ -66,38 +66,32 @@ internal static class MsalIdParameter
     /// </summary>
     internal static class MsalIdHelper
     {
-        private static CorePlatformInformationBase corePlatformInformationBaseInstance;
```
bgavrilMS, line 69: Minor rewrite - just CorePlatformInformationBase.Instance - it's easier to read.

core/tests/Test.Microsoft.Identity.Core.Unit/CommonHelpers/Resources/OpenidConfiguration-B2C.json
```diff
@@ -0,0 +1,17 @@
+﻿{
```
bgavrilMS, line 1: These exist already but have been copied to ADAL and MSAL test projects.

PR #1244, opened 2018-08-31
Refactor and introdce platform specific tests

bgavrilMS:
1. Fix a race condition with test initialization

Fix initialization issues with tests - because tests have access to the internal structure of ADAL / MSAL, they should re-initialize the module (factories, statics etc.) - otherwise things like CorePlatformInformation is initialized either with net45PlatformInformation or with testPlatformInformation, depending on what test runs first.

2. Reference core tests in ADAL / MSAL tests. This has a few advantages:
a. it matches the structure of the product code
b. it avoids estranged objects, e.g. TokenCacheAccessor exists in MSAL but also in core (as a convenience). If a test method in CoreTests like Foo(TokenCacheAccessor) exists, it will use the type from core - meaning that the method cannot be used in MSALTests

3. Introduce netcore tests in msal


core/src/MsalIdHelper.cs
```diff
@@ -66,38 +66,32 @@ internal static class MsalIdParameter
...
-            corePlatformInformationBaseInstance = CorePlatformInformationBase.Instance;
-        }
-
         public static IDictionary<string, string> GetMsalIdParameters()
         {
-            if (corePlatformInformationBaseInstance == null)
+            if (CorePlatformInformationBase.Instance == null)
             {
-                throw CoreExceptionFactory.Instance.GetClientException(CoreErrorCodes.PlatformNotSupported, CoreErrorMessages.PlatformNotSupported);
+                throw new InvalidOperationException("Internal error - PlatformInformation not initialized");
```
henrik-me: >throw new InvalidOperationException("Internal error - PlatformInformation not initialized"); [](start = 16, length = 92)

I suggest to use the error codes/error messages made for this purpose. 
Aside: not sure this should even be done here... but it was the best place to discover something was wrong... perhaps it should be removed completely this check.

core/src/MsalIdHelper.cs
```diff
@@ -66,38 +66,32 @@ internal static class MsalIdParameter
     /// </summary>
     internal static class MsalIdHelper
```
henrik-me: >MsalIdHelper [](start = 26, length = 12)

IMO, this thing is static and should be caching everything. Having to even redo the values here is odd and only needed for testing. What about optimizing it for run-time and use from reflection or internals visible to clean it up (e.g. via the ForceInitialize in the module initializer)...? This thing is right now waste of CPU cycles

henrik-me:
>     <TargetFramework>netcoreapp2.0</TargetFramework>

if build machines supports 2.1 now, this can be updated.

---
Refers to: msal/tests/Test.MSAL.NET.Unit.netcore/Test.MSAL.NET.Unit.netcore.csproj:4 in 108185d. [](commit_id = 108185df43707864a10f5d843898e42662e03b44, deletion_comment = False)

henrik-me (changes requested):
:clock1:

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/ModuleInitializer.cs
```diff
@@ -44,27 +44,49 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal
...
         }
+
+
+        private static void InitializeModule()
+        {
+            CoreLoggerBase.Default = new AdalLogger(Guid.Empty);
+            CoreTelemetryService.InitializeCoreTelemetryService(Telemetry.GetInstance() as ITelemetry);
+            // Several statics in the library depends on platform information being timely initialized. 
+            // The static initializer on PlatformInformationBase will ensure this gets done.
+            CorePlatformInformationBase.Instance = new PlatformInformation();
```
henrik-me, line 88: >CorePlatformInformationBase.Instance = new PlatformInformation(); [](start = 12, length = 65)

If we are doing it this way then you can remove the comments above

jennyf19: ok.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Platform/PlatformInformationBase.cs
```diff
@@ -74,19 +58,13 @@ public virtual bool GetCacheLoadPolicy(IPlatformParameters parameters)
             return true;
         }
 
-        public virtual Uri ValidateRedirectUri(Uri redirectUri, RequestContext requestContext)
+        public override void ValidateRedirectUri(Uri redirectUri, RequestContext requestContext)
```
henrik-me, line 61: >void [](start = 24, length = 4)

IMO: this should still return the URI, now the caller has to know nothing is done to the input param. Alternatively the method should change name: ThrowIfRedirectUriIsNotValid ??? Then it is more obvious what the method does and that the flow will be terminated in case something fails.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenInteractiveHandler.cs
```diff
@@ -58,8 +59,8 @@ internal class AcquireTokenInteractiveHandler : AcquireTokenHandlerBase
             UserIdentifier userId, string extraQueryParameters, IWebUI webUI, string claims)
             : base(requestData)
         {
-            this.redirectUri = platformInformation.ValidateRedirectUri(redirectUri, RequestContext);
-
+            platformInformation.ValidateRedirectUri(redirectUri, RequestContext);
```
henrik-me, line 62: >ValidateRedirectUri [](start = 32, length = 19)

see comment in PlatformInformationBase

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/PlatformInformation.cs
```diff
@@ -70,7 +71,7 @@ public override string GetDeviceModel()
         public override string GetAssemblyFileVersionAttribute()
         {
             // TODO: Check if assembly file version can be read in iOS assembly as well or not. For now, we use assembly version instead.
-            return typeof(AdalIdHelper).GetTypeInfo().Assembly.GetName().Version.ToString();
+            return typeof(UriParamsHelper).GetTypeInfo().Assembly.GetName().Version.ToString();
```
henrik-me, line 74: >typeof(UriParamsHelper).GetTypeInfo().Assembly.GetName().Version.ToString(); [](start = 19, length = 76)

IMO: should only be "calculated" once and then return the calculated value, this thing is wasting lots of CPU cycles

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Platform/PlatformInformationBase.cs
```diff
@@ -29,41 +29,25 @@
...
-
-        public abstract string GetOperatingSystem();
-
-        public abstract string GetDeviceModel();
-
-        public virtual string GetAssemblyFileVersionAttribute()
+        public override string GetAssemblyFileVersionAttribute()
         {
-            var assemblyFileVersion = typeof(AdalIdHelper).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>();
+            var assemblyFileVersion = typeof(UriParamsHelper).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>();
```
henrik-me, line 42: >assemblyFileVersion [](start = 16, length = 19)

cache and return instead of continously calculating the value

core/src/CorePlatformInformationBase.cs
```diff
@@ -67,5 +71,30 @@ public virtual string GetDefaultRedirectUri(string correlationId)
         {
             return DefaultRedirectUri;
         }
+
+        public static string GetClientVersion()
```
henrik-me, line 75: >static [](start = 15, length = 6)

statics are evil: should be instance based and values should be cached.

core/src/CorePlatformInformationBase.cs
```diff
@@ -67,5 +71,30 @@ public virtual string GetDefaultRedirectUri(string correlationId)
...
+            if (match.Success)
+            {
+                string[] version = match.Groups[0].Value.Split(new[] { '=' }, StringSplitOptions.None);
+                return version[1];
+            }
+
+            return null;
+        }
+
+        public static string GetAssemblyFileVersion()
```
henrik-me, line 88: >static [](start = 15, length = 6)

statics are evil: should be instance based and values should be cached.

core/src/CorePlatformInformationBase.cs
```diff
@@ -67,5 +71,30 @@ public virtual string GetDefaultRedirectUri(string correlationId)
...
+        {
+            return CorePlatformInformationBase.Instance.GetAssemblyFileVersionAttribute();
+        }
+
+        public static string GetAssemblyInformationalVersion()
```
henrik-me, line 93: >static [](start = 15, length = 6)

statics are evil: should be instance based and values should be cached.

core/src/Exceptions/CoreErrorCodes.cs
```diff
@@ -54,6 +54,7 @@ internal class CoreErrorCodes
         public const string InvalidAuthority = "invalid_authority";
 
         public const string PlatformNotSupported = "platform_not_supported";
+        public const string PlatformNotInitialized = "platform_not_initialized";
```
henrik-me, line 57: >PlatformNotInitialized [](start = 28, length = 22)

Why not the one above? (same for the messages). I believe it was made for the same reason (and most likely this change is removing the use of the one above)

core/src/Exceptions/CoreErrorMessages.cs
```diff
@@ -69,5 +69,7 @@ internal class CoreErrorMessages
             "Cannot find the principal name for the user. Possible causes: the application does not request the User Account Information and the Enterprise Authentication capabilities or the user is not AD or AAD joined.";
 
         public const string PlatformNotSupported = "Platform Not Supported";
+
+        public const string PlatformNotInitialized = "Internal error - PlatformInformation not initialized";
```
henrik-me, line 73: >PlatformNotInitialized [](start = 28, length = 22)

see CodeErrorMessages (ad the line above)

core/src/Telemetry/DefaultEvent.cs
```diff
@@ -35,7 +35,7 @@ public DefaultEvent(string clientId) : base((string) (EventBase.EventNamePrefix
         {
             this[EventNamePrefix + "client_id"] = clientId;
             this[EventNamePrefix + "sdk_platform"] = CorePlatformInformationBase.Instance.GetProductName()?.ToLowerInvariant();
-            this[EventNamePrefix + "sdk_version"] = MsalIdHelper.GetMsalVersion();
+            this[EventNamePrefix + "sdk_version"] = CorePlatformInformationBase.GetClientVersion();
```
henrik-me, line 38: >CorePlatformInformationBase [](start = 52, length = 27)

Keep it instance based (see my comment on the use of statics) . those are super evil!

msal/src/Microsoft.Identity.Client/Exceptions/MsalClientException.cs
```diff
@@ -129,6 +129,12 @@ public class MsalClientException : MsalException
         /// The library is loaded on a platform which is not supported.
         /// </summary>
         public const string PlatformNotSupported = "platform_not_supported";
+
+        /// <summary>
+        /// The library is loaded on a platform which has not been initialized.
+        /// </summary>
+        public const string PlatformNotInitialized = "platform_not_initialized";
```
henrik-me, line 136: >PlatformNotInitialized [](start = 28, length = 22)

see above and other related comments

msal/src/Microsoft.Identity.Client/Internal/PlatformInformationBase.cs
```diff
@@ -43,7 +43,7 @@ static PlatformInformationBase()
         public override string GetAssemblyFileVersionAttribute()
         {
             return
-                typeof (MsalIdHelper).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>().Version;
+                typeof (UriParamsHelper).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>().Version;
```
henrik-me, line 46: >typeof (UriParamsHelper).Ge [](start = 16, length = 27)

cache

msal/tests/Test.MSAL.NET.Unit.net45/Test.MSAL.NET.Unit.net45.csproj
```diff
@@ -29,6 +29,7 @@
     <Compile Include="..\..\..\core\tests\Test.Microsoft.Identity.Core.Unit\TestConstants.cs"  LinkBase="CoreUT" />
     <Compile Include="..\..\..\core\tests\Test.Microsoft.Identity.Core.Unit\TestPlatformInformation.cs"  LinkBase="CoreUT" />
     <Compile Include="..\..\..\core\tests\Test.Microsoft.Identity.Core.Unit\TestLogger.cs"  LinkBase="CoreUT" />
+    <Compile Include="..\..\..\core\tests\Test.Microsoft.Identity.Core.Unit\MiscTests\*.cs"  LinkBase="CoreUT\MiscTests" />
```
henrik-me, line 32: >MiscTests [](start = 76, length = 9)

Not a fan of this folder name. 
Could possibly be platform instead? (e.g. net45)

henrik-me, line 32: >MiscTests [](start = 110, length = 9)

see other comment on naming

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Platform/PlatformInformationBase.cs
jennyf19: Or "ThrowIfRedirectUriIsNull" as this is just checking that the redirectUri is not null

---
In reply to: [215738440](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1244#discussion_r215738440) [](ancestors = 215738440)

core/src/Exceptions/CoreErrorCodes.cs
bgavrilMS: I believe this should actually be InternalError - if the users sees this they should log a bug.

jennyf19: sounds good.

---
In reply to: [215757011](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1244#discussion_r215757011) [](ancestors = 215757011)

PR #1246, opened 2018-09-07
silent call authority null - use app authority

SomkaPe:
[VSTS - 530576](https://identitydivision.visualstudio.com/DevEx/_workitems/edit/530576)
AcquireTokenSilent without passing authority does not work as expected when migrating from ADAL V4 to MSAL V2.

### DIFF
--- a/core/src/Instance/AadAuthority.cs
+++ b/core/src/Instance/AadAuthority.cs
@@ -51,6 +51,8 @@ namespace Microsoft.Identity.Core.Instance
 
         private const string AadInstanceDiscoveryEndpoint = "https://login.microsoftonline.com/common/discovery/instance";
 
+        public const string AADCanonicalAuthorityTemplate = "https://{0}/{1}/";
+
         internal AadAuthority(string authority, bool validateAuthority) : base(authority, validateAuthority)
         {
             AuthorityType = AuthorityType.Aad;
@@ -114,7 +116,7 @@ namespace Microsoft.Identity.Core.Instance
             Uri authorityUri = new Uri(CanonicalAuthority);
 
             CanonicalAuthority = 
-                string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/", authorityUri.Authority, tenantId);
+                string.Format(CultureInfo.InvariantCulture, AADCanonicalAuthorityTemplate, authorityUri.Authority, tenantId);
         }
     }
 }
--- a/core/src/Instance/AadAuthority.cs
+++ b/core/src/Instance/AadAuthority.cs
@@ -27,6 +27,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Identity.Core.OAuth2;
@@ -107,6 +108,13 @@ namespace Microsoft.Identity.Core.Instance
         {
             return GetFirstPathSegment(CanonicalAuthority);
         }
-        
+
+        internal override void UpdateTenantId(string tenantId)
+        {
+            Uri authorityUri = new Uri(CanonicalAuthority);
+
+            CanonicalAuthority = 
+                string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/", authorityUri.Authority, tenantId);
+        }
     }
 }
--- a/core/src/Instance/Authority.cs
+++ b/core/src/Instance/Authority.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Identity.Core.Instance
 {
     internal abstract class Authority
     {
-        private static readonly HashSet<string> TenantlessTenantNames =
+        internal static readonly HashSet<string> TenantlessTenantNames =
             new HashSet<string>(new[] {"common", "organizations"});
         private bool _resolved;
 
@@ -242,6 +242,8 @@ namespace Microsoft.Identity.Core.Instance
         protected abstract string GetDefaultOpenIdConfigurationEndpoint();
 
         internal abstract string GetTenantId();
+
+        internal abstract void UpdateTenantId(string tenantId);
 
         private async Task<TenantDiscoveryResponse> DiscoverEndpointsAsync(string openIdConfigurationEndpoint,
             RequestContext requestContext)
--- a/core/src/Instance/B2CAuthority.cs
+++ b/core/src/Instance/B2CAuthority.cs
@@ -71,8 +71,11 @@ namespace Microsoft.Identity.Core.Instance
             Uri authorityUri = new Uri(CanonicalAuthority);
             var segments = authorityUri.Segments;
 
+            var b2cPrefix = segments[1].TrimEnd('/');
+            var b2cPolicy = segments[3].TrimEnd('/');
+
             CanonicalAuthority = string.Format(CultureInfo.InvariantCulture, B2CCanonicalAuthorityTemplate, 
-                authorityUri.Authority, segments[1].TrimEnd('/'), tenantId, segments[3].TrimEnd('/'));
+                authorityUri.Authority, b2cPrefix, tenantId, b2cPolicy);
         }
     }
 }
--- a/core/src/Instance/B2CAuthority.cs
+++ b/core/src/Instance/B2CAuthority.cs
@@ -34,6 +34,7 @@ namespace Microsoft.Identity.Core.Instance
     internal class B2CAuthority : AadAuthority
     {
         public const string Prefix = "tfp"; // The http path of B2C authority looks like "/tfp/<your_tenant_name>/..."
+        public const string B2CCanonicalAuthorityTemplate = "https://{0}/{1}/{2}/{3}/";
 
         internal B2CAuthority(string authority, bool validateAuthority) : base(authority, validateAuthority)
         {
@@ -44,8 +45,9 @@ namespace Microsoft.Identity.Core.Instance
                 throw new ArgumentException(CoreErrorMessages.B2cAuthorityUriInvalidPath);
             }
 
-            CanonicalAuthority = string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/{2}/{3}/", authorityUri.Authority,
+            CanonicalAuthority = string.Format(CultureInfo.InvariantCulture, B2CCanonicalAuthorityTemplate, authorityUri.Authority,
                 pathSegments[0], pathSegments[1], pathSegments[2]);
+
             AuthorityType = AuthorityType.B2C;
         }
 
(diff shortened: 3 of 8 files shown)


asnow003:
sorry accidentally hit the close

core/src/Instance/AadAuthority.cs
```diff
@@ -107,6 +108,13 @@ internal override string GetTenantId()
...
             return GetFirstPathSegment(CanonicalAuthority);
         }
-        
+
+        internal override void UpdateTenantId(string tenantId)
+        {
+            Uri authorityUri = new Uri(CanonicalAuthority);
+
+            CanonicalAuthority = 
+                string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/", authorityUri.Authority, tenantId);
```
henrik-me: >"https://{0}/{1}/" [](start = 60, length = 18)

for b2c you made this a const... perhaps you should do the same here for consistency

core/src/Instance/B2CAuthority.cs
```diff
@@ -44,8 +45,9 @@ internal B2CAuthority(string authority, bool validateAuthority) : base(authority
                 throw new ArgumentException(CoreErrorMessages.B2cAuthorityUriInvalidPath);
             }
 
-            CanonicalAuthority = string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/{2}/{3}/", authorityUri.Authority,
+            CanonicalAuthority = string.Format(CultureInfo.InvariantCulture, B2CCanonicalAuthorityTemplate, authorityUri.Authority,
                 pathSegments[0], pathSegments[1], pathSegments[2]);
```
henrik-me, line 49: this behaves differently from UpdateTenentId, is that on purpose? Could you simply call UpdateTenentId here and get the expected CanonicalAuthority out?

core/src/Instance/Authority.cs
```diff
@@ -37,7 +37,7 @@ namespace Microsoft.Identity.Core.Instance
 {
     internal abstract class Authority
     {
-        private static readonly HashSet<string> TenantlessTenantNames =
+        internal static readonly HashSet<string> TenantlessTenantNames =
```
henrik-me: >internal [](start = 8, length = 8)

is this needed?

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -325,9 +325,21 @@ internal async Task<MsalAccessTokenCacheItem> FindAccessTokenAsync(Authenticatio
...
+                requestParams.RequestContext.Logger.Warning(msg);
+                requestParams.RequestContext.Logger.WarningPii(msg);
+
+                return null;
+            }
+
             lock (LockObject)
             {
-                string msg = "Looking up access token in the cache..";
+                msg = "Looking up access token in the cache..";
```
henrik-me: >. [](start = 60, length = 1)

nit: remove extra dot #Closed

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -325,9 +325,21 @@ internal async Task<MsalAccessTokenCacheItem> FindAccessTokenAsync(Authenticatio
         private MsalAccessTokenCacheItem FindAccessTokenCommon
             (AuthenticationRequestParameters requestParams, string preferredEnvironmentAlias, ISet<string> environmentAliases)
         {
+            string msg;
+
+            //no authority passed
+            if (environmentAliases.Count == 0)
+            {
+                msg = "No authority provided. Skipping cache lookup ";
```
henrik-me, line 333: >msg [](start = 16, length = 3)

While issue exists everywhere, pls. consider making such strings consts.

We are passing strings around which means it will continue to be copied into memory causing extra CPU cycles #Closed

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -394,107 +406,46 @@ private MsalAccessTokenCacheItem FindAccessTokenCommon
...
-                    else
-                    {
-                        msg = "No tokens found for matching client_id, user and scopes.";
-                        requestParams.RequestContext.Logger.Info(msg);
-                        requestParams.RequestContext.Logger.InfoPii(msg);
+                //filter by authority
+                IEnumerable<MsalAccessTokenCacheItem> filteredByPrefferedAlias =
+                    filteredItems.Where
+                    (item => item.Environment.Equals(preferredEnvironmentAlias, StringComparison.OrdinalIgnoreCase))
+                    .ToList();
```
henrik-me: >.ToList() [](start = 20, length = 9)

can we avoid the conversion toList ? seems like we are wasting CPU and memory?

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -394,107 +406,46 @@ private MsalAccessTokenCacheItem FindAccessTokenCommon
...
-                    {
-                        msg = "Multiple tokens found for matching authority, client_id, user and scopes.";
-                        requestParams.RequestContext.Logger.Error(msg);
-                        requestParams.RequestContext.Logger.ErrorPii(msg);
-                        
-                        throw new MsalClientException(MsalClientException.MultipleTokensMatchedError,
-                            MsalErrorMessage.MultipleTokensMatched);
-                    }
+                    throw new MsalClientException(MsalClientException.MultipleTokensMatchedError,
+                        MsalErrorMessage.MultipleTokensMatched);
```
henrik-me, line 446: Can we pls. inform the developer on what they should do as part of the error? e.g. that they have to clean the cache? #Closed

henrik-me (changes requested):
:clock1:

core/src/Instance/B2CAuthority.cs
```diff
@@ -63,5 +65,14 @@ internal override string GetTenantId()
...
             return new Uri(CanonicalAuthority).Segments[2].TrimEnd('/');
         }
+
+        internal override void UpdateTenantId(string tenantId)
+        {
+            Uri authorityUri = new Uri(CanonicalAuthority);
+            var segments = authorityUri.Segments;
+
+            CanonicalAuthority = string.Format(CultureInfo.InvariantCulture, B2CCanonicalAuthorityTemplate, 
+                authorityUri.Authority, segments[1].TrimEnd('/'), tenantId, segments[3].TrimEnd('/'));
```
bgavrilMS: It's pretty hard to read segments[1], segments[3] etc. - maybe create local variables or give an example  #Closed

core/src/Instance/B2CAuthority.cs
SomkaPe: UpdateTenantId in Authority class is applicable only for aad authority it does not work for b2c authortiy (url template is diffirent) , it should not be called for b2c authority because with b2c common authority should no be used. But if it will be called we need b2c specific function

core/src/Instance/AadAuthority.cs
SomkaPe: ok #Closed

msal/src/Microsoft.Identity.Client/TokenCache.cs
SomkaPe: ok #Closed

msal/src/Microsoft.Identity.Client/TokenCache.cs
SomkaPe: ok

msal/src/Microsoft.Identity.Client/TokenCache.cs
SomkaPe: It should never happen, combination of authority clientId user and scope is unique key in the cache, 
if it happened it means that cache is corrupted, due to some bug or error...  current  message MultipleTokensMatched = "The cache contains multiple tokens satisfying the requirements. Call AcquireToken again providing more requirements like authority"  is not relevant any more since authotiy is mandatory now, i can change it to "The cache contains multiple tokens satisfying the requirements. Try to clear token cache and report the issue"

core/src/Instance/B2CAuthority.cs
SomkaPe: good point - will change 
 #Closed

msal/src/Microsoft.Identity.Client/TokenCache.cs
SomkaPe: it is just local log message it is is not reused

msal/src/Microsoft.Identity.Client/Exceptions/MsalErrorMessage.cs
```diff
@@ -70,7 +70,7 @@ internal static class MsalErrorMessage
         public const string MissingAuthenticateHeader = "WWW-Authenticate header was expected in the response";
 
         public const string MultipleTokensMatched =
-            "The cache contains multiple tokens satisfying the requirements. Call AcquireToken again providing more requirements like authority";
+            "The cache contains multiple tokens satisfying the requirements. Try to clear token cache and report the issue";
```
henrik-me: >and report the issue [](start = 102, length = 20)

perhaps the above also needs to be there. or the clear cache. I would not like to see a report the issue in this text.

core/src/Instance/B2CAuthority.cs
henrik-me: this I do not get. then why is it implemented in this class?

---
In reply to: [216100716](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1246#discussion_r216100716) [](ancestors = 216100716)

msal/src/Microsoft.Identity.Client/TokenCache.cs
henrik-me: It may or may only be used once, but right now you are passing it to 2 different methods both constructing a copy of the string. had you however used a stringbuilder it would have been by reference. alternatively you can use a const, if I recall correctly it will not create the extra copy (which uses memory and CPU cycles)

---
In reply to: [216103745](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1246#discussion_r216103745) [](ancestors = 216103745)

core/src/Instance/B2CAuthority.cs
SomkaPe: that part should be refactored ))

msal/src/Microsoft.Identity.Client/TokenCache.cs
henrik-me: (not a fan of report the issue part, kindly remove that) if anything point to a page in the wiki where we can describe to the developer how to handle the situation the best.

---
In reply to: [216103209](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1246#discussion_r216103209) [](ancestors = 216103209)

msal/src/Microsoft.Identity.Client/Exceptions/MsalErrorMessage.cs
henrik-me: consider adding an aka.ms/msal-multipletokensfoundincache

---
In reply to: [216107714](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1246#discussion_r216107714) [](ancestors = 216107714)

core/src/Instance/B2CAuthority.cs
SomkaPe: I can do that if we have time, do we ?

henrik-me: I will stop by. I do not understand this.

henrik-me (approved):
:shipit:

PR #1250, opened 2018-09-11
using common authority if account with null id is passed to silent api

SomkaPe:
- Using common authority in silent api if passed account object has null id -->it can happens if cache contains entity in adalv3 format ( without client info)
- Adding "consumers" to the list of tenantless tenant names


core/src/Instance/Authority.cs
```diff
@@ -38,7 +38,7 @@ namespace Microsoft.Identity.Core.Instance
     internal abstract class Authority
     {
         internal static readonly HashSet<string> TenantlessTenantNames =
-            new HashSet<string>(new[] {"common", "organizations"});
+            new HashSet<string>(new[] {"common", "organizations", "consumers"});
```
jmprieur, line 41: thanks for adding this @SomkaPe. I guess consumer implies the MSA tenant Id, but I also guess we'd have it in the IAccount.

jmprieur (commented):
Thanks!

jmprieur (commented):
LGTM
Good job;
I tested the silentAccountIdNull branch, and indeed, I don't get the NullReferenceException any longer (that was with common)
This also works well with a proper tenant.

My only question is, organizations and consumers don't make sense for ADAL, so I guess they won't be in the authority anyway. It's ok to have the TenantlessTenantNames in core?

Good job @SomkaPe

henrik-me (dismissed):
@SomkaPe : was ADAL tested?

msal/tests/Test.MSAL.NET.Unit/UnifiedCacheTests.cs
```diff
@@ -153,5 +153,23 @@ public void UnifedCache_MsalStoresToAndReadRtFromAdalCache()
 
             Assert.IsNotNull(result1);
         }
+
+        [TestMethod]
+        [Description("Test unified token cache")]
+        public void UnifedCache_MsalSilentApiCommonAuthorityUsedForAdalAccountsWithNullAccountId()
```
henrik-me: Consider adding a similar test for ADAL?

msal/src/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -282,7 +282,8 @@ public async Task RemoveAsync(IAccount account)
                 authority = Core.Instance.Authority.CreateAuthority(Authority, ValidateAuthority);
                 var tenantId = authority.GetTenantId();
 
-                if (Core.Instance.Authority.TenantlessTenantNames.Contains(tenantId))
+                if (Core.Instance.Authority.TenantlessTenantNames.Contains(tenantId) 
+                    && account.HomeAccountId?.TenantId != null)
```
jennyf19: I tried this out w/a null account (silent flow, no cache) and got a null reference here because the account object is null. We check for a null account next in the SilentRequest and throw an MsalUiRequiredException there. we need to ensure account is not null before moving forward.  #Closed

msal/tests/Test.MSAL.NET.Unit/UnifiedCacheTests.cs
```diff
@@ -153,5 +153,23 @@ public void UnifedCache_MsalStoresToAndReadRtFromAdalCache()
...
+        {
+            PublicClientApplication app = new PublicClientApplication(TestConstants.ClientId);
+            try
+            {
+                var res = app.AcquireTokenSilentAsync(TestConstants.Scope, 
+                    new Account(null, TestConstants.Name, TestConstants.ProductionPrefNetworkEnvironment)).Result;
```
jennyf19: What if you have a null account object?

henrik-me (dismissed):
:clock1:

msal/src/Microsoft.Identity.Client/ClientApplicationBase.cs
SomkaPe: account is mandatory parameter for this api #Closed

henrik-me:
Seems like the build is failing on the initialization step. Removing block.

msal/src/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -274,18 +274,26 @@ public async Task RemoveAsync(IAccount account)
...
             if (authority == null)
             {
-                authority = Core.Instance.Authority.CreateAuthority(Authority, ValidateAuthority);
-                var tenantId = authority.GetTenantId();
-
-                if (Core.Instance.Authority.TenantlessTenantNames.Contains(tenantId))
-                {
-                    authority.UpdateTenantId(account.HomeAccountId.TenantId);
-                }
+                authority = GetAuthority(account);
```
henrik-me, line 301: >authority [](start = 16, length = 9)

Could this still be null?

PR #1251, opened 2018-09-12
Jennyf/duplicate key fix

jennyf19:
**Manual testing:** ⛄️ 🐱 🚀 🤖 #!
ADAL:
> Xamarin Android, iOS, UWP (Interactive/silent)
> Desktop Test App

MSAL:
> Xamarin Android, iOS, UWP (Interactive/silent)
> Desktop Test App


core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -96,11 +96,6 @@ private string GetTeamId()
                 match = SecKeyChain.QueryAsRecord(queryRecord, out resultCode);
             }
 
-            if (resultCode != SecStatusCode.Success)
```
bgavrilMS: Why are we removing this? Won't we get the null ref because match is null?

msal/src/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -39,12 +39,6 @@ internal class SilentRequest : RequestBase
         public SilentRequest(AuthenticationRequestParameters authenticationRequestParameters, bool forceRefresh)
             : base(authenticationRequestParameters)
         {
-            if (authenticationRequestParameters.Account == null)
-            {
-                throw new MsalUiRequiredException(MsalUiRequiredException.UserNullError, "Null user was passed in AcquiretokenSilent API. Pass in " +
```
bgavrilMS, line 44: You can add the error message to a constant file and it will be easier to test for it as well.

jennyf19: yes...will do.

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -63,12 +61,13 @@ enum CredentialAttrType
         private readonly string TeamIdKey = "teamIDHint";
 
         private string keychainGroup;
+        private RequestContext _requestContext;
```
bgavrilMS, line 64: [NIT] I know we don't enforce coding style yet and that we sometimes use underscore for private members, but it would be good to be consistent with the MS coding style, i.e. class members should use camelCasing (not _usederscoreCamel and not PascalCasing)

core/src/Platforms/iOS/TokenCacheAccessor.cs
jennyf19: removing this because this is not the fix. so just going with what we have currently in dev until we find a solution.

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -97,11 +96,14 @@ private string GetTeamId()
                 match = SecKeyChain.QueryAsRecord(queryRecord, out resultCode);
             }
 
+            if (resultCode != SecStatusCode.Success)
```
bgavrilMS: I think it would be safer to also add OR match==null to the condition

bgavrilMS: Also, do we want to add a bit of logging here?

core/src/Platforms/iOS/TokenCacheAccessor.cs
jennyf19: I agree...will fix.

bgavrilMS (approved):
LGTM - minor changes only

core/src/Platforms/iOS/TokenCacheAccessor.cs
henrik-me: fun thing is that the middleware team likes the _ as it is what .netcore does...

---
In reply to: [216906861](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1251#discussion_r216906861) [](ancestors = 216906861)

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -390,7 +387,7 @@ private SecRecord CreateRecord(string account, string service, string generic, i
                 Synchronizable = _defaultSyncSetting,
             };
         }
-        
+
         private SecStatusCode Remove(string account, string service, int type)
```
henrik-me, line 388: nit: seems like a number of lines in a few places like here.

henrik-me (approved):
:shipit:

core/src/Platforms/iOS/TokenCacheAccessor.cs
jennyf19: this should be removed...maybe it's showing from an earlier commit?

msal/src/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
jennyf19: done.

core/src/Platforms/iOS/TokenCacheAccessor.cs
jennyf19: Leaving this as-is.

core/src/Platforms/iOS/TokenCacheAccessor.cs
jennyf19: I think I removed all the extra lines. #!*

PR #1258, opened 2018-09-14
Update changelog for MSAL release 2.0.1 Preview

trwalke:
Update changelog for MSAL release 2.0.1

### DIFF
--- a/msal/changelog.txt
+++ b/msal/changelog.txt
@@ -3,7 +3,7 @@
 This release contains:
 - Resolved logger issue in B2C pm sample for msal with object null reference.
 - Adding required claims to error messages.
-- Resolving issue where B2C access and refresh tokens were not cached due to missing claims in the id Token.
+- Resolved issue where B2C access and refresh tokens were not cached due to missing claims in the id Token.
 - Resolved issue with AcquireTokenSilent without passing authority when migrating from ADAL V4 to MSAL V2.
 - Resolved issue with duplicate key when caching.
 - MSAL will now use the common authority if account with null id is passed to silent api
--- a/msal/changelog.txt
+++ b/msal/changelog.txt
@@ -1,3 +1,14 @@
+2.0.1-preview
+=============
+This release contains:
+- Resolved logger issue in b2 cpm sample for msal with object null reference.
+- Adding required claims to error messages.
+- Resolving issue where B2C access and refresh tokens were not cached due to missing claims in the id Token.
+- Resolved issue with AcquireTokenSilent without passing authority when migrating from ADAL V4 to MSAL V2.
+- Resolved issue with duplicate key when caching.
+- MSAL will now use the common authority if account with null id is passed to silent api
+- Fixed issue with accounts not being removed from cache.
+
 2.0.0-preview
 =============
 This release contains:
--- a/msal/changelog.txt
+++ b/msal/changelog.txt
@@ -1,7 +1,6 @@
 2.0.1-preview
 =============
 This release contains:
-- Adding required claims to error messages.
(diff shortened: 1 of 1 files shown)


msal/changelog.txt
```diff
@@ -1,3 +1,14 @@
+2.0.1-preview
+=============
+This release contains:
+- Resolved logger issue in B2C pm sample for msal with object null reference.
```
jennyf19: I would remove this.  Henrik made a fix because netcore2 apps were not loading netcore1, and we got the exception in the logger. but all this is internal. I don't think we need to mention it here.

msal/changelog.txt
```diff
@@ -1,3 +1,14 @@
...
+- Adding required claims to error messages.
```
jennyf19: I don't think we need this either. it's just an update to the log message.

msal/changelog.txt
```diff
@@ -1,3 +1,14 @@
...
+- Resolved issue where B2C access and refresh tokens were not cached due to missing claims in the id Token.
```
jennyf19: Maybe something like "Implemented client side workaround for handling missing claims in id_token returned from B2C"...something like that  @jmprieur

msal/changelog.txt
```diff
@@ -1,3 +1,14 @@
...
+- Resolved issue with AcquireTokenSilent without passing authority when migrating from ADAL V4 to MSAL V2.
+- Resolved issue with duplicate key when caching.
+- MSAL will now use the common authority if account with null id is passed to silent api
```
jennyf19: Is there a better way to word this? @jmprieur ideas?

msal/changelog.txt
```diff
@@ -1,3 +1,14 @@
...
+- Fixed issue with accounts not being removed from cache.
```
jennyf19: "Removed caching of B2C tokens in ADAL"

msal/changelog.txt
```diff
@@ -1,3 +1,14 @@
+2.0.1-preview
+=============
+This release contains:
+- Resolved logger issue in B2C pm sample for msal with object null reference.
+- Adding required claims to error messages.
+- Resolved issue where B2C access and refresh tokens were not cached due to missing claims in the id Token.
+- Resolved issue with AcquireTokenSilent without passing authority when migrating from ADAL V4 to MSAL V2.
+- Resolved issue with duplicate key when caching.
```
jennyf19: No customer reported this. we found it, i don't think it needs to be mentioned.

jennyf19 (approved):
🦑 maybe get a 2nd opinion, but approving so it's not blocked or anything.

msal/changelog.txt
jmprieur: I would write (to be customer focused):
This release contains bug fixes on top of MSAL 2.0.0-preview
- When using MSAL 2.0.0-preview with Azure AD B2C, the cache was never hit. (See MSAL#[604](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/604)), and the accounts were not removed correctly (See MSAL #[613](613))
- The `TokenCacheExtensions.Deserialize` was throwing if a null array of bytes was passed as arguments instead of silently not doing anything. See MSAL #[603](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/603) 
- When migrating a token cache from ADAL v3 or ADAL v4 to MSAL 2.0, the override of acquire token silent without authority used to work incorrectly (cache was missed)

jmprieur (changes requested):
Please see my proposed comment

jmprieur (approved):
LGTM :)

PR #1259, opened 2018-09-20
Add unit tests for CacheFallbackOperations and better null protection

bgavrilMS:


core/src/Cache/AdalTokenCacheKey.cs
```diff
@@ -51,6 +53,7 @@ internal enum TokenSubjectType
     /// <summary>
     /// <see cref="AdalTokenCacheKey"/> can be used with Linq to access items from the TokenCache dictionary.
     /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
```
bgavrilMS, line 56: Convenience method for debugging. nq = no quotes. #Closed

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -96,6 +102,14 @@ internal class CacheFallbackOperations
                     return;
                 }
 
+                Debug.Assert(
```
bgavrilMS: These should only be hit in tests #Closed

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -186,16 +206,25 @@ internal class CacheFallbackOperations
...
+            string accountOrUserId)
         {
             try
             {
                 IDictionary<AdalTokenCacheKey, AdalResultWrapper> adalCache =
                 AdalCacheOperations.Deserialize(legacyCachePersistance.LoadCache());
 
-                RemoveEntriesWithMatchingId(environmentAliases, identifier, adalCache);
-                RemoveEntriesWithMatchingName(environmentAliases, displayableId, adalCache);
+                if (!string.IsNullOrEmpty(accountOrUserId))
```
bgavrilMS, line 231: We are now deleting too much. The logic for deleting is documented in the md file and in the comments of this method.  
@jennyf19 - I was under the impression that the original code had this IF statement in here as well. Did you find use cases where we have to remove based on account_id AND based on username as well? I was pretty convinced that if we have an account_id we can stop there.  #Closed

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -242,27 +272,27 @@ internal class CacheFallbackOperations
...
 
             foreach (KeyValuePair<AdalTokenCacheKey, AdalResultWrapper> kvp in adalCache)
             {
-                string environment = new Uri(kvp.Key.Authority).Host;
-                string cachedAccountId = ClientInfo.CreateFromJson(kvp.Value.RawClientInfo).ToAccountIdentifier();
+                string rawClientInfo = kvp.Value.RawClientInfo;
 
-                if (environmentAliases.Contains(environment, StringComparer.OrdinalIgnoreCase) &&
-                    string.Equals(identifier, cachedAccountId, StringComparison.OrdinalIgnoreCase))
+                if (!String.IsNullOrEmpty(rawClientInfo))
```
bgavrilMS, line 299: Yet another place where a silent null ref can occur. We should reconsider catching all exceptions. #Closed

core/src/Cache/CacheFallbackOperations.cs
henrik-me: not convinced, will discuss "in-person"

---
In reply to: [219185322](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1259#discussion_r219185322) [](ancestors = 219185322)

core/src/Cache/CacheFallbackOperations.cs
henrik-me: perhaps at a higher level though than here? looking forward to more general discussions on this topic

---
In reply to: [219186752](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1259#discussion_r219186752) [](ancestors = 219186752)

core/src/Properties/InternalsVisibleTo.cs
```diff
@@ -27,5 +27,8 @@
 
 using System.Runtime.CompilerServices;
 
-// The core test project is the only one allowed to reference Core.dll for convenience. Do not add any other entries here. 
+// The core test project is the only one allowed to reference Core.dll for convenience. Do not add any other entries here, except for test.
 [assembly: InternalsVisibleTo("Test.Microsoft.Identity.Core.Unit, PublicKey=0024000004800000940000000602000000240000525341310004000001000100B5FC90E7027F67871E773A8FDE8938C81DD402BA65B9201D60593E96C492651E889CC13F1415EBB53FAC1131AE0BD333C5EE6021672D9718EA31A8AEBD0DA0072F25D87DBA6FC90FFD598ED4DA35E44C398C454307E8E33B8426143DAEC9F596836F97C8F74750E5975C64E2189F45DEF46B2A2B1247ADC3652BF5C308055DA9")]
+//[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100B5FC90E7027F67871E773A8FDE8938C81DD402BA65B9201D60593E96C492651E889CC13F1415EBB53FAC1131AE0BD333C5EE6021672D9718EA31A8AEBD0DA0072F25D87DBA6FC90FFD598ED4DA35E44C398C454307E8E33B8426143DAEC9F596836F97C8F74750E5975C64E2189F45DEF46B2A2B1247ADC3652BF5C308055DA9")]
```
henrik-me: >//[ [](start = 0, length = 3)

? #Closed

core/tests/Test.Microsoft.Identity.Core.Unit/CacheTests/CacheFallbackOperationsTests.cs
```diff
@@ -0,0 +1,311 @@
...
+using NSubstitute;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Test.Microsoft.Identity.Core.Unit.Mocks;
+
+namespace Test.Microsoft.Identity.Core.Unit.CacheTests
+{
+    [TestClass]
+    public class CacheFallbackOperationsTests
```
henrik-me, line 14: >CacheFallbackOperationsTests [](start = 17, length = 28)

Great! #Closed

core/tests/Test.Microsoft.Identity.Core.Unit/Mocks/MockHelpers.cs
```diff
@@ -60,6 +60,11 @@ public static string CreateClientInfo()
 
         public static string CreateClientInfo(string uid, string utid)
         {
+            if (String.IsNullOrEmpty(uid) && String.IsNullOrEmpty(utid))
```
henrik-me: >if (String.IsNullOrEmpty(uid) && String.IsNullOrEmpty(utid)) [](start = 12, length = 60)

does clientinfo exists with out both uid and utid? at least i would expect uid should always be there... but actually I would expect both. thus should this be an || instead of &&? #Closed

bgavrilMS: Yes, I would expect both as well. We don't really construct it like this through the code, only via deserialization from json. Theoretically you can have one of those 2 members be empty / null. I'll move this logic to the test #Closed

core/src/Properties/InternalsVisibleTo.cs
bgavrilMS: Removed. #Closed

core/src/Cache/CacheFallbackOperations.cs
MarkZuber: Don't we also run our tests in RELEASE mode?  If so, these won't be hit in test code, and then what?   #Closed

bgavrilMS: No, tests run in Debug mode. Also, when we perform manual testing using the dev apps, we use the debug version of the library. 

The reason for using Debug.Assert is that I am not sure there aren't cases where those conditions don't actually hold. The implications in this case are minimal for the user (e.g. minor loss of SSO), so I am hesitant to crash the app with an exception. I am also hesitant to add a lot of validation code in hot paths of the library when it is used in production. #Closed

MarkZuber: IMHO, we should run tests in RELEASE mode.  That's where our customers are using us and that's what we should be testing.  Testing in debug mode should be "yup, we do that too for sanity" but we should be testing what customers are running.

I'll need to look more at this particular scenario for what you're trying to achieve with the asserts, but we should be able to find a way to validate debug/release equally without un-naturally burdening users with exceptions/errors that aren't errors/exceptions. #Closed

core/src/Cache/cache.readme.md
```diff
@@ -0,0 +1,39 @@
...
+- GetAccounts and RemoveAccount do not work for ClientCredentail Grant
+
+
+
+
+## Other notes
+
+MSAL is able able to retrieve objects of type IAccount from the MSAL cache. 
+The token cache is scoped on Env and ClientId.
+When upgrading from ADALv3, the ClinetInfo is null because we do not request an IdToken;
```
henrik-me, line 36: we could make an update to ADAL v3 which would get the ClientInfo/IdToken ?

core/src/Cache/CacheFallbackOperations.cs
henrik-me: Petro asked us to do removal both based on account and on displayable id.

---
In reply to: [219186124](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1259#discussion_r219186124) [](ancestors = 219186124)

core/src/Properties/InternalsVisibleTo.cs
henrik-me: I still see it :)

---
In reply to: [219531456](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1259#discussion_r219531456) [](ancestors = 219531456)

core/src/Cache/CacheFallbackOperations.cs
henrik-me: I would prefer we identify a way to test with-out Debug.Asserts

---
In reply to: [219557012](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1259#discussion_r219557012) [](ancestors = 219557012)

bgavrilMS: There is no functional difference between DEBUG and RELEASE modes and .net provides ample support for adding Debug only code. Agreed that we should never perf test against Debug. 

VS logic uses this technique a lot, especially to protect against changes from other parts of the app. 

it's also a way of "documenting" the code by identifying invariants. It took me a long time to realize that those conditions always hold true. #Closed

core/src/Cache/CacheFallbackOperations.cs
bgavrilMS: I don't think the current implementation is ok. It deletes too much. I could remove line 197 (the call to RemoveEntriesWithMatchingId) and get the same result because  RemoveEntriesWithMatchingName ignores the client id althogheter. 

For example in the scenario where: 

a. Several users for which we do not have preferredUsername get tokens
b. One user activates a token cache removal (e.g. signs out)

Actual: all users get signed out of the adal cache 


I think maybe what Petro was trying to say is to scope the deletion both by accountId and by username at the same time ?

 #Closed

henrik-me: if all tests passes we are fine...

---
In reply to: [219568047](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1259#discussion_r219568047) [](ancestors = 219568047)

core/src/Cache/CacheFallbackOperations.cs
henrik-me: How about logging?

---
In reply to: [219556166](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1259#discussion_r219556166) [](ancestors = 219556166)

henrik-me:
When validating this pls. ensure to use the B2C sample app

core/src/Cache/CacheFallbackOperations.cs
bgavrilMS: Discussed offline, will do logging and bring this back at another time.  #Closed

core/src/Cache/cache.readme.md
bgavrilMS: :) Yes, might take a bit more time to do this though.

core/src/Cache/CacheFallbackOperations.cs
henrik-me: Have we seen any issues like this ever occur? Is it something critical? Can we consider only doing this validation in case there is not token returned or an error happens?

---
In reply to: [220294785](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1259#discussion_r220294785) [](ancestors = 220294785)

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -96,9 +107,21 @@ internal class CacheFallbackOperations
...
+                    CoreLoggerBase.Default.Error(DifferentEnvError);
+                    CoreLoggerBase.Default.ErrorPii(DifferentEnvError);
+                }
+
+                if (!String.Equals(rtItem?.Environment, (new Uri(authority)).Host, StringComparison.OrdinalIgnoreCase))
+                {
+                    CoreLoggerBase.Default.Error(DifferentAuthorityError);
+                    CoreLoggerBase.Default.ErrorPii(DifferentAuthorityError);
+                }
+
```
henrik-me: Have we seen any issues like this ever occur? Is it something critical? Can we consider only doing this validation in case there is not token returned or an error happens?
Goal: Only do these string operations if really needed (if things are found we should be good)
e.g. is it possible to only do this work as part of the exception handler?

bgavrilMS:
Will merge manually

PR #1260, opened 2018-09-20
Update telemetry start/stop to be in TelemetryHelper disposable

MarkZuber:
Simple refactor to remove try/finally for start/stop telemetry and move to a disposable helper class within a using block.


core/src/Telemetry/TelemetryTokenCacheAccessor.cs
```diff
@@ -36,114 +36,74 @@ internal class TelemetryTokenCacheAccessor : TokenCacheAccessor
...
-            }
         }
 
         public void SaveIdToken(MsalIdTokenCacheItem item, RequestContext requestContext)
         {
-            var cacheEvent = new CacheEvent(CacheEvent.TokenCacheWrite) { TokenType = CacheEvent.TokenTypes.ID };
-            CoreTelemetryService.GetInstance().StartEvent(requestContext.TelemetryRequestId, cacheEvent);
-            try
+            using (CoreTelemetryService.CreateTelemetryHelper(requestContext.TelemetryRequestId,
+                new CacheEvent(CacheEvent.TokenCacheWrite) { TokenType = CacheEvent.TokenTypes.ID }))
```
henrik-me, line 58: >new CacheEvent(CacheEvent.TokenCacheWrite) { TokenType = CacheEvent.TokenTypes.ID }) [](start = 16, length = 84)

suggestion: pass in only the params here and create the cache object in the helper method?
Will simplify the line. 
Perhaps also create a factory method for creating the cacheevent which can be used in the CreateTelemetryHelper method?

core/src/CoreTelemetryService.cs
```diff
@@ -47,5 +42,14 @@ public static ITelemetry GetInstance()
...
         }
+
+        public static TelemetryHelper CreateTelemetryHelper(
+            string requestId,
+            EventBase eventToStart,
+            EventBase eventToEnd = null,
+            bool shouldFlush = false)
+        {
+            return new TelemetryHelper(GetInstance(), requestId, eventToStart, eventToEnd ?? eventToStart, shouldFlush);
+        }
```
henrik-me, line 53: I understand why it's here now, but... I'm having some debates with myself... does this belong here?

henrik-me:
utests?

core/src/Telemetry/TelemetryTokenCacheAccessor.cs
MarkZuber: The CacheEvent is not related directly to CreateTelemetryHelper, that class only knows about BaseEvent so I'm not sure it's appropriate to add knowledge of the specific events to the CreateTelemetryHelper method.  Maybe I'm missing your point here.

core/src/CoreTelemetryService.cs
MarkZuber: As a stepping stone, I'd say it belongs here.  I also updated other parts of the code to all route into CoreTelemetryService as the central choke point since some were going to Telemetry.GetInstance() as well.  it looks like CoreTelemetryService exists so that we can override the value of the ITelemetry global instance, since you can't set it on the Telemetry class directly.  I'd like to look at our global object factory/singleton mechanisms more broadly in later work.

core/src/Telemetry/TelemetryTokenCacheAccessor.cs
henrik-me: Not good at explaining myself...
True, but given you are introducing a helper. Then I think we can help even more. Could be in another help where this is wrapped in though if that is better. I suppose a helper that took the full responsibility of creating the event and sending the event, and disposing it... is what I was looking for. No strong opinions on this though.

MarkZuber: Only the calling code knows which start/stop event(s) to actually log.  So I think it should be the calling code's responsibility to create the event and pass it in.

henrik-me: But there could be a helper, now you are newing up another object to get this. my objectives are: 1. code simplicity and 2. SRP :) - No strong opinions btw. as this was a suggestion :)

henrik-me (approved):
:shipit:

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Telemetry.cs
```diff
@@ -68,5 +68,14 @@ public void StopEvent(string requestId, EventBase eventToStop)
         {
             
         }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="requestId"></param>
```
bgavrilMS, line 75: Nit: we usually remove empty comments;

core/src/Microsoft.Identity.Core.csproj
```diff
@@ -13,7 +13,7 @@
 
 <!-- Important note: this assembly should not be shipped to customers; it acts only as a placeholder for files that are included
 in ADAL and MSAL and to ease unit testing. No projects should take a dependency on Core.dll either -->
-  
+
```
bgavrilMS, line 16: No changes were done to this file.

core/src/Telemetry/TelemetryHelper.cs
```diff
@@ -0,0 +1,79 @@
...
+            _requestId = requestId;
+            _eventToEnd = eventToEnd;
+            _shouldFlush = shouldFlush;
+            _telemetry.StartEvent(_requestId, eventToStart);
+        }
+
+        #region IDisposable Support
+        private bool _disposedValue = false; // To detect redundant calls
+
+        protected virtual void Dispose(bool disposing)
```
bgavrilMS: Cool, there is definitely a lack of Disposables in our codebase, and of the using { } pattern.

core/src/Telemetry/TelemetryHelper.cs
```diff
@@ -0,0 +1,79 @@
...
+                }
+
+                _disposedValue = true;
+            }
+        }
+
+        // This code added to correctly implement the disposable pattern.
+        public void Dispose()
+        {
+            Dispose(true);
```
bgavrilMS, line 75: You can also call GC.SuppressFinalize(this); here to tell the Garbage Collector that the object does not need to be cleanup anymore because we did it manually. 

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern

core/tests/Test.Microsoft.Identity.Core.Unit/Telemetry/TelemetryHelperTests.cs
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        private class _TestEvent : EventBase
+        {
+            public _TestEvent(string eventName) : base(eventName)
+            {
+            }
+        }
+
+        private class _TestTelem : ITelemetry
```
bgavrilMS, line 67: We have a dependency on NSubstitute that we use to to mock objects, count calls etc. It makes the code a bit more succinct, although creating the mock manually does make debugging tests a lot easier. Just a thought / suggestion, this is perfectly fine and easy to read as is.

core/src/Telemetry/TelemetryHelper.cs
MarkZuber: This is only needed if you implement a Finalizer, which you should only do if you are disposing unmanaged resources.  I used the VS "implement disposable pattern" and it generates a bunch of comments to that effect, and we had this same concern on a previous team.  But if you're only using managed resources (which we are in this case since we're just using IDisposable to call something on scope exit) then we don't need to suppress the finalizer since there isn't one.

core/tests/Test.Microsoft.Identity.Core.Unit/Telemetry/TelemetryHelperTests.cs
MarkZuber: Yeah, I saw that dependency and its usage with the Logger infra.  I've used Moq in the past which is super helpful but we didn't have a dependency on it yet.  Is NSubstitute a full mocking framework like Moq?  If so I'd rather use something like that.  I've just only used Moq previously...

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Telemetry.cs
bgavrilMS, line 75: We tend to remove empty comments and / or empty comment sections.

core/src/CoreTelemetryService.cs
```diff
@@ -47,5 +42,14 @@ public static ITelemetry GetInstance()
         {
             return instance;
         }
+
+        public static TelemetryHelper CreateTelemetryHelper(
+            string requestId,
+            EventBase eventToStart,
+            EventBase eventToEnd = null,
+            bool shouldFlush = false)
```
bgavrilMS, line 50: It would be useful to have a comment on this method with regards to how the shoudFlush param should be used - it is not clear from reading the method signature when I should make this true. Or alternatively do not use a boolean param at all and instead create 2 public methods to convey more meaning with the name.

core/src/Telemetry/TelemetryHelper.cs
```diff
@@ -0,0 +1,79 @@
...
+        public TelemetryHelper(ITelemetry telemetry, string requestId, EventBase eventToStart, EventBase eventToEnd, bool shouldFlush)
+        {
+            _telemetry = telemetry;
+            _requestId = requestId;
+            _eventToEnd = eventToEnd;
+            _shouldFlush = shouldFlush;
+            _telemetry.StartEvent(_requestId, eventToStart);
+        }
+
+        #region IDisposable Support
```
bgavrilMS, line 52: Very cool, there are not enough Disposables in our code where there should be... Also, I believe this is the first proper implementation of the DIsposable pattern.

core/src/Telemetry/TelemetryHelper.cs
bgavrilMS: You should call GC.SupressFinalize(this) here - see details https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern

bgavrilMS: Ah, cool, I didn't realize that. Also, appologies for some duplicate comments, my Internet connection went down exactly when I was hitting submit :)

core/tests/Test.Microsoft.Identity.Core.Unit/Telemetry/TelemetryHelperTests.cs
bgavrilMS: Yeah, afaik NSubstitute does everything that Moq does but plays a bit better with C#. The important thing is that they both target netstandard1.3, which means we can use these libraries for all target framework tests (except ADAL netstandard1.1 which we should probably deprecate sooner or later). Currently we have only net45 tests, but I have a work item that enables other tests as well.

core/src/CoreTelemetryService.cs
MarkZuber: I agree with you, however I also don't know why flush is used in the one case that it's used in RequestBase.cs and not anywhere else.  Let's sync up and see if we can figure out why this is happening and I'll update a comment here.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Telemetry.cs
MarkZuber: Added a comment here like in the others.

bgavrilMS (approved):

bgavrilMS:
@MarkZuber  - I believe this can go it as the build ADAL .NET Unified CI & PR & Nightly AppCenter should probably not be blocking PRs at this point

PR #1263, opened 2018-09-21
Implement DeviceCode flow in MSAL

MarkZuber:
This PR currently has several TODOs in it and not complete unit test coverage.  I'm starting the PR to start getting feedback on the TODOs as well as any other placement/usage issues in the existing code.


msal/src/Microsoft.Identity.Client/PublicClientApplication.cs
```diff
@@ -95,7 +95,33 @@ public PublicClientApplication(string clientId, string authority)
...
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="scopes"></param>
+        /// <param name="extraQueryParameters"></param>
+        /// <param name="deviceCodeResultCallback"></param>
+        /// <returns></returns>
+        public async Task<AuthenticationResult> AcquireTokenAsync(
```
henrik-me: >AcquireTokenAsync [](start = 48, length = 17)

not sure this flow should be made available on mobile (as those have browsers)... though not sure I can see a big issue in having it there...

henrik-me: >public async Task<AuthenticationResult> AcquireTokenAsync( [](start = 8, length = 58)

naming, AquireTokenWithDeviceCodeAsync .... and per other comment, this might have to move to specific platforms.

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -0,0 +1,125 @@
...
+            TimeSpan timeRemaining = _deviceCodeResult.ExpiresOn - DateTimeOffset.UtcNow;
+
+            while (timeRemaining.TotalSeconds > 0.0)
+            {
+                try
+                {
+                    await base.SendTokenRequestAsync().ConfigureAwait(false);
+                    return;
+                }
+                catch (MsalServiceException ex)
```
henrik-me: >catch [](start = 16, length = 5)

we may have to also handle the "429" here, unless we handle that either at lower level or at a higher layer... #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -0,0 +1,125 @@
...
+                {
+                    if (ex.ErrorCode.Equals("authorization_pending", StringComparison.OrdinalIgnoreCase))
+                    {
+                        await Task.Delay(TimeSpan.FromSeconds(_deviceCodeResult.Interval)).ConfigureAwait(false);
```
henrik-me: >_deviceCodeResult.Interval [](start = 62, length = 26)

should probably evaluate the value of this before using it ... and perhaps we should have a min, and some "exponential" function to it? #Closed

msal/tests/Test.MSAL.NET.Unit/RequestsTests/DeviceCodeRequestTests.cs
```diff
@@ -0,0 +1,202 @@
...
+            Assert.AreEqual("https://microsoft.com/devicelogin", actualDeviceCodeResult.VerificationUrl);
+
+            Assert.IsTrue(HttpMessageHandlerFactory.IsMocksQueueEmpty, "All mocks should have been consumed");
+        }
+
+        // todo: this is my test sample app that I built as a standalone console app.
+        // Need to figure out where we can check this in to do manual testing 
+        // of real device code flow login until we get automation in place on the
+        // MSAL side.
+        private static class __Program
```
henrik-me: >__Program [](start = 29, length = 9)

seems like we typically would put this in our developer sample apps?

msal/src/Microsoft.Identity.Client/PublicClientApplication.cs
```diff
@@ -95,7 +95,33 @@ public PublicClientApplication(string clientId, string authority)
...
+            IEnumerable<string> scopes, 
+            string extraQueryParameters, 
+            Action<DeviceCodeResult> deviceCodeResultCallback)
```
henrik-me: >Action<DeviceCodeResult> deviceCodeResultCallback [](start = 12, length = 49)

not sure about this one, but it makes sense. right now though we do not allow developers control on the flow thus we should at least introduce a way to cancel the polling

henrik-me: (btw. is an issue today in ADAL.NET)

---
In reply to: [219634068](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1263#discussion_r219634068) [](ancestors = 219634068)

msal/src/Microsoft.Identity.Client/DeviceCodeResult.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// </summary>
+        public string ClientId { get; }
+
+        /// <summary>
+        /// Identifier of the target resource that would be the recipient of the token.
+        /// </summary>
+        public string Resource { get; }
+        // todo: this should be SCOPE(S) and not resource for v2 endpoint...
+    }
+}
```
jennyf19, line 112: I know this is public, but looks to be similar to what is in ADAL. Is there a way to move the shared code to Core and surface the public api to ADAL and MSAL respectively?

msal/src/Microsoft.Identity.Client/Internal/DeviceCodeResponse.cs
```diff
@@ -0,0 +1,74 @@
...
+                DeviceCode, 
+                VerificationUrl, 
+                DateTime.UtcNow.AddSeconds(ExpiresIn), 
+                Interval, 
+                Message, 
+                clientId, 
+                resource);
+        }
+    }
+}
```
jennyf19, line 74: If this class is internal, can it be moved to Core? If the ADAL code for this is similar, can they both share the code in core? Removes duplicate code across the libraries.

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -0,0 +1,125 @@
...
+                        timeRemaining = _deviceCodeResult.ExpiresOn - DateTimeOffset.UtcNow;
+                    }
+                    else
+                    { 
+                        throw;
+                    }
+                }
+            }
+
+            throw new MsalServiceException("code_expired", "Verification code expired before contacting the server");
```
jennyf19: @henrik-me will say these need to be consts. :) #Closed

msal/src/Microsoft.Identity.Client/DeviceCodeResult.cs
bgavrilMS: Yes, good point. You can eliminate duplicate code by adding it to core. However, we do not currently have any public code in core, so this warrants a discussion. Please note that core.dll assembly is just a convenience to hold files and run tests against - ADAL and MSAL directly include all *.cs files from core. 

So we could move DeviceCodeResult file into core. We don't want to introduce breaking changes, so the ADAL namespaces cannot change. Therefore, when adding it to core, we have do something like: 

#if ADAL
#else 
#endif

bgavrilMS: Alternatively, if the only commonality are data classes, we could live with the duplication.

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -0,0 +1,125 @@
...
+            SupportADFS = false;
+        }
+
+        internal override async Task PreTokenRequestAsync()
+        {
+            await base.PreTokenRequestAsync().ConfigureAwait(false);
+
+            OAuth2Client client = new OAuth2Client();
+
+            var deviceCodeScopes = new SortedSet<string>();
```
bgavrilMS: Does it need to be a sorted set? I'd expect to see HashSet everywhere and when a sorted set is needed a comment should explain why (i.e. how do you modify the dictionary later on etc.)

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -0,0 +1,125 @@
...
+            deviceCodeScopes.UnionWith(AuthenticationRequestParameters.Scope);
+            deviceCodeScopes.Add("offline_access");
```
bgavrilMS: I think you added constants for these, but possibly the code changes were not in dev when you created this.

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -0,0 +1,125 @@
...
+            deviceCodeScopes.Add("profile");
+            deviceCodeScopes.Add("openid");
+
+            client.AddBodyParameter(OAuth2Parameter.ClientId, AuthenticationRequestParameters.ClientId);
+            client.AddBodyParameter(OAuth2Parameter.Scope, deviceCodeScopes.AsSingleString());
+
+            // todo: THIS IS A MAJOR HACK.  Work with Shiung/Henrik on proper way to determine the device code endpoint
+            string deviceCodeEndpoint = AuthenticationRequestParameters.Authority.TokenEndpoint
+                .Replace("token", "devicecode")
+                .Replace("common", "organizations");
```
bgavrilMS: In Integrated Windows Auth you get a pretty good error message when the user tries to use /common - is it not the case here?

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -0,0 +1,125 @@
...
+    // * do we need to set SaveToCache in the constructor?
+    internal class DeviceCodeRequest : RequestBase
+    {
+        private readonly Action<DeviceCodeResult> _deviceCodeResultCallback;
+        private DeviceCodeResult _deviceCodeResult;
+
+        public DeviceCodeRequest(AuthenticationRequestParameters authenticationRequestParameters, Action<DeviceCodeResult> deviceCodeResultCallback) 
+            : base(authenticationRequestParameters)
+        {
+            _deviceCodeResultCallback = deviceCodeResultCallback;
```
bgavrilMS: Arg null check

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
bgavrilMS: The library should thow *ClientExceptions. We use ServiceExceptions to wrap the server side exceptions.  #Closed

msal/src/Microsoft.Identity.Client/PublicClientApplication.cs
bgavrilMS: Async offers a rather clean way of providing cancellation, you just need to allow to pass in a cancellation token which the user can then call "Cancel()" on. In fact, it is recommened that all public async APIs offer an overload with a CancelationToken.

msal/src/Microsoft.Identity.Client/PublicClientApplication.cs
```diff
@@ -95,7 +95,33 @@ public PublicClientApplication(string clientId, string authority)
...
+            // todo:  Need to work with Bogdan/JM/Henrik on API semantics and get this exposed
+            // on the IPublicClientApplication interface as well once I determine the final appropriate
+            // placement for this call.
+
+            Authority authority = Core.Instance.Authority.CreateAuthority(Authority, ValidateAuthority);
+
+            var requestParams = CreateRequestParameters(authority, scopes, null, UserTokenCache);
+            requestParams.ExtraQueryParameters = extraQueryParameters;
+
+            var handler = new DeviceCodeRequest(requestParams, deviceCodeResultCallback);
```
bgavrilMS, line 139: I like the whole callback approach, although it is quite different than what was done in ADAL. A few thoughts: 

1. We should check that the user actually passes in a callback and provide a good error message if he does not
2. The callback could be async. Sync code is quite frowned upon, especially in ASP.NET scenarios. You can easily wrap sync code in an async callback, but not so much the opposite way around.

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -0,0 +1,125 @@
...
+                client.CreateFullEndpointUri(new Uri(deviceCodeEndpoint)), 
+                HttpMethod.Post,
+                AuthenticationRequestParameters.RequestContext).ConfigureAwait(false);
+
+            // todo: fix resource/scopes here for how we want to invoke this...
+            _deviceCodeResult = response.GetResult(AuthenticationRequestParameters.ClientId, "todo resource");
+            _deviceCodeResultCallback(_deviceCodeResult);
+        }
+
+        protected override async Task SendTokenRequestAsync()
```
bgavrilMS: This logic seems ideantical to the one in ADAL - we could consolidate it in core.

msal/tests/Test.MSAL.NET.Unit/RequestsTests/DeviceCodeRequestTests.cs
```diff
@@ -0,0 +1,202 @@
...
+            var authenticationResult = task.Result;
+            Assert.IsNotNull(authenticationResult);
+            Assert.IsNotNull(actualDeviceCodeResult);
+
+            Assert.AreEqual("client_id", actualDeviceCodeResult.ClientId);
+            Assert.AreEqual("BAQABAAEAAADXzZ3ifr-GRbDT45zNSEFEfU4P-bZYS1vkvv8xiXdb1_zX2xAcdcfEoei1o-t9-zTB9sWyTcddFEWahP1FJJJ_YVA1zvPM2sV56d_8O5G23ti5uu0nRbIsniczabYYEr-2ZsbgRO62oZjKlB1zF3EkuORg2QhMOjtsk-KP0aw8_iAA", actualDeviceCodeResult.DeviceCode);
+            Assert.AreEqual("client_id", actualDeviceCodeResult.ClientId);
+
+            // todo: figure out how to unit test this one since it's calculated from current time based on the datetime offset retrieved in the response
+            // Assert.AreEqual(5, actualDeviceCodeResult.ExpiresOn);
```
bgavrilMS, line 140: There are a series of tests in DeviceProfileTests that check the timeout exceptions. But, as stated above, we if we centralize this logic in core, we might be able to reuse those tests.

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
henrik-me: current recommendations from evo is that we do not use common. the error message returned is not guaranteed to be brilliant and is currently subject to change.

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
MarkZuber: Exponential doesn't make sense here.  The user has a fixed time to respond with their device code login and the expectation is that once they've entered their code that they'll be signed in.  Exponential fallback here would present a very strange user experience where the user may have to wait longer than expected to complete their login experience.

msal/tests/Test.MSAL.NET.Unit/RequestsTests/DeviceCodeRequestTests.cs
MarkZuber: yes, this is for the developer sample app.  I just didn't want to lose this snippet while I worked out getting the sample in place and whether there's a place for this in our internal dev/test apps instead of just an external sample.  since this makes testing the app for real actually possible.  would appreciate guidance here.

msal/src/Microsoft.Identity.Client/PublicClientApplication.cs
MarkZuber: I've added CancellationToken support to the top level public API (by adding 2nd versions of the methods) and passing it along as needed.  I didn't wire it up deeply throughout the entire stack but we now have the ability to.  I've wired it up in the case of DeviceCodeRequest so we have the ability to cancel the in progress token request operation.

msal/src/Microsoft.Identity.Client/DeviceCodeResult.cs
MarkZuber: I actually think the data classes _may_ end up varying since there are protocol changes in the v2 endpoint.  This is why I just made them copies so that we can vary them with the final V2 endpoint versions.

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
MarkZuber: i've made them consts. #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
MarkZuber: Nope, I saw this as SortedSet elsewhere and used it but looking closer this can be HashSet.

msal/tests/Test.MSAL.NET.Unit/RequestsTests/DeviceCodeRequestTests.cs
henrik-me: Add it to the one of the dev sample apps. My recommendation would be to start with the desktop app.

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
MarkZuber: fixed

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
MarkZuber: fixed #Closed

msal/src/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -197,13 +198,14 @@ public async Task<IAccount> GetAccountAsync(string accountId)
...
-
+        }
+
         /// <summary>
         /// Attempts to acquire an access token for the <paramref name="account"/> from the user token cache. 
         /// </summary> 
         /// <param name="scopes">Scopes requested to access a protected API</param>
-        /// <param name="account">Account for which the token is requested. <see cref="IAccount"/></param>
+        /// <param name="account">Account for which the token is requested. <see cref="IAccount"/></param>
+        /// <param name="cancellationToken"></param>
```
henrik-me, line 208: Should provide some explanation. Also pls. provide a link to the wiki which talks about how to use the cancellation Token. Also pls. ensure that the usage of Code Flow including the cancelation token is in a sample. Until release you can validate with internal build and publish the sample when the release is out.

henrik-me: aside: Perhaps we should wait adding this everywhere until we are ready to implement it everywhere and then in this change only include changes related to Device Code now?

---
In reply to: [220018243](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1263#discussion_r220018243) [](ancestors = 220018243)

msal/src/Microsoft.Identity.Client/IPublicClientApplication.cs
```diff
@@ -168,8 +169,35 @@ public interface IPublicClientApplication : IClientApplicationBase
...
+        // endif !ANDROID
 #endif
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="scopes"></param>
+        /// <param name="extraQueryParameters"></param>
+        /// <param name="deviceCodeResultCallback"></param>
+        /// <returns></returns>
```
henrik-me, line 181: as mentioned everywhere, should have texts for both interfaces and implementations. But really great to see the cancellation stuff making it's way!

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -0,0 +1,133 @@
...
+    // * sample app / test dev app as a console application for validation
+    internal class DeviceCodeRequest : RequestBase
+    {
+        private readonly Action<DeviceCodeResult> _deviceCodeResultCallback;
+        private DeviceCodeResult _deviceCodeResult;
+
+        public DeviceCodeRequest(AuthenticationRequestParameters authenticationRequestParameters, Action<DeviceCodeResult> deviceCodeResultCallback) 
+            : base(authenticationRequestParameters)
+        {
+            _deviceCodeResultCallback = deviceCodeResultCallback ?? throw new ArgumentNullException("A deviceCodeResultCallback must be provided for Device Code authentication to work properly");
```
henrik-me, line 52: >deviceCodeResultCallback [](start = 103, length = 24)

not sure if we should use an overload which calls out the arg specifically. 
Aside: When an exception like this bubbles up it would be meaningless to the developer (unless they read our internal code), thus the actual place to do validation is at input time in the public API?

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
henrik-me: I agree on exponential, however I put "exponential" in my question to indicate something different from linear. but it might be fine with something linear for the duration (expected about 15 mins of polling). We should though still ensure that we have to deal with 429's correctly throughout the stack (currently expecting that will be lower in the stack than here). 

---
In reply to: [219945965](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1263#discussion_r219945965) [](ancestors = 219945965)

msal/src/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        protected override async Task SendTokenRequestAsync(CancellationToken cancellationToken)
+        {
+            TimeSpan timeRemaining = _deviceCodeResult.ExpiresOn - DateTimeOffset.UtcNow;
+
+            while (timeRemaining.TotalSeconds > 0.0)
+            {
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    throw new OperationCanceledException();
```
henrik-me, line 102: >OperationCanceledException [](start = 30, length = 26)

thinking, we might want to have a Client exception which is an operation Canceled exception, though not sure about this as it will not be standard (but something for the PM when considering the programming model for the developer)

msal/tests/Test.MSAL.NET.Unit/RequestsTests/InteractiveRequestTests.cs
```diff
@@ -325,7 +326,7 @@ public void DuplicateQueryParameterErrorTest()
 
             try
             {
-                request.PreTokenRequestAsync().Wait();
+                request.PreTokenRequestAsync(CancellationToken.None).Wait();
```
henrik-me, line 329: >CancellationToken.None [](start = 45, length = 22)

these are defaults, right? Shouldn't the API keep working as is if there are no changes made to the client code?

henrik-me:
shouldn't this change have been rebased instead? Seems like now I got additional 40+ commits to this PR which is already in dev.

henrik-me (changes requested):
:clock1:

msal/tests/Test.MSAL.NET.Unit/RequestsTests/InteractiveRequestTests.cs
MarkZuber: Yes, these are the defaults.  You just can't make CancellationToken.None a default parameter to a function since it's not a compile time constant.

MarkZuber:
@henrik-me -- i'm also confused as to the extra commits.  I did do a rebase from dev, but it may be an artifact of the iwa change coming into dev maybe?  seems odd to me but I did notice you guys do a rebase model and not a merge to feature branch model like we did in msal c++.

MarkZuber:
Something in the merge/rebase was messed up here.  Abandoning this PR and I'll have an updated one with more/most fixes in it from another PR.

PR #1266, opened 2018-09-24
Jennyf/uwp

jennyf19:
The silent call in the Xamarin MSAL apps does not work because the below line was removed. Adding it back.


henrik-me (approved):
:shipit:

PR #1268, opened 2018-09-24
Ui automation app updates

trwalke:
Adding button to export authentication result to file to workaround a limitation in the appium framework.

### DIFF
--- a/adal/automation/WinFormsAutomationApp/MainForm.cs
+++ b/adal/automation/WinFormsAutomationApp/MainForm.cs
@@ -21,7 +21,7 @@ namespace WinFormsAutomationApp
         }
 
         private Command _commandToRun = null;
-        private string AuthResultFile = "AuthResult.txt";
+        private const string AuthResultFile = "AuthResult.txt";
 
         public MainForm()
         {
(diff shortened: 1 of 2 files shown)


adal/automation/WinFormsAutomationApp/MainForm.cs
```diff
@@ -99,6 +101,14 @@ private void acquireDeviceCode_Click(object sender, EventArgs e)
         {
             _commandToRun = AuthenticationHelper.AcquireDeviceCodeAsync;
             pageControl1.SelectedTab = dataInputPage;
-        }        
+        }
+
+        private void writeToFile_Click(object sender, EventArgs e)
+        {
+            if (File.Exists(AuthResultFile))
```
henrik-me, line 108: >AuthResultFile [](start = 28, length = 14)

want to add it to the location of the running process? Not sure this is what File.Write does by default. Consider making the location of the file explicit.

henrik-me: approving, however pls. take a look at this first.

henrik-me (approved):

adal/automation/WinFormsAutomationApp/MainForm.cs
trwalke: i verified that the default behavior is to use the current running process location which will make it easier for the agent to find it.

PR #1273, opened 2018-09-26
Workaround for UWP bug in HttpStack when cloning the http con…

bgavrilMS:
…tent

### DIFF
--- a/adal/devApps/AdalUAPTestApp/MainPage.xaml.cs
+++ b/adal/devApps/AdalUAPTestApp/MainPage.xaml.cs
@@ -27,8 +27,10 @@
 
 using Microsoft.IdentityModel.Clients.ActiveDirectory;
 using System;
+using System.Collections.Generic;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
+using Windows.Web.Http;
 
 // The Blank Page item template is documented at https://go.microsoft.com/fwlink/?LinkId=402352&clcid=0x409
 
@@ -107,11 +109,6 @@ namespace UAPTestApp
             }
         }
 
-        private void Button_Click_1(object sender, RoutedEventArgs e)
-        {
-            this.AccessToken.Text = string.Empty;
-        }
-
 #pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
         private async void Button_Click_2(object sender, RoutedEventArgs e)
 #pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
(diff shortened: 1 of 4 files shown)


core/src/Http/HttpRequest.cs
```diff
@@ -236,5 +238,32 @@ private static async Task<HttpResponse> CreateResponseAsync(HttpResponseMessage
...
+                }
+            }
+
+            // WORKAROUND 
+            // On UWP there is a bug in the Http stack that causes an exception to be thrown when moving around a stream.
+            // https://stackoverflow.com/questions/31774058/postasync-throwing-irandomaccessstream-error-when-targeting-windows-10-uwp
+            // LoadIntoBufferAsync is necessary to buffer content for multiple reads - see https://stackoverflow.com/questions/26942514/multiple-calls-to-httpcontent-readasasync
+            // Documentation is sparse, but it looks like loading the buffer into memory avoids the bug, without 
+            // replacing the System.Net.HttpClient with Windows.Web.Http.HttpClient, which is not exactly a drop in replacement
+            await clone.LoadIntoBufferAsync().ConfigureAwait(false);
```
bgavrilMS, line 264: This line is the actual fix

core/src/Helpers/CoreHelpers.cs
```diff
@@ -233,20 +233,5 @@ private static void AddKeyValueString(StringBuilder messageBuilder, string key,
             messageBuilder.Append(value);
         }
 
-        internal static async Task<HttpContent> DeepCopyAsync(HttpContent content)
```
bgavrilMS, line 236: Not needed anywhere else but in HttpRequest below.

jennyf19 (approved):
LGTM

core/src/Http/HttpRequest.cs
henrik-me: nice find! should there be an #if UWP for this then?

---
In reply to: [220696902](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1273#discussion_r220696902) [](ancestors = 220696902)

adal/devApps/AdalUAPTestApp/MainPage.xaml.cs
```diff
@@ -164,5 +161,41 @@ private async void AccessTokenSilentButton_Click(object sender, RoutedEventArgs
...
+        {
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
+
+            this.AccessToken.Text = string.Empty;
+            AuthenticationContext context = new AuthenticationContext("https://login.microsoftonline.com/common");
+            try
+            {
+                AuthenticationResult authResult = await context.AcquireTokenAsync(
+                    "https://graph.microsoft.com",
+                    ClientId, new UserCredential("bogavril@microsoft.com"))
```
henrik-me, line 178: >"bogavril@microsoft.com") [](start = 49, length = 25)

should this be an input via the UI instead? (perhaps the UI should be grabbing this from the user of the system by default?)

henrik-me:
Need any updates for MSAL apps?

bgavrilMS:
@henrik-me - I made an update so that the app fetches the username from the OS.

henrik-me:
thanks

core/src/Http/HttpRequest.cs
bgavrilMS: Yes, I guess so. I'll address this in a separate PR.

PR #1276, opened 2018-09-27
MSAL UWP test app and minor fixes

bgavrilMS:


msal/src/Microsoft.Identity.Client/Platforms/uap/PlatformInformation.cs
```diff
@@ -49,6 +49,71 @@ public override string GetProductName()
             return "MSAL.UAP";
         }
 
+        /// <summary>
+        /// Get a principal name that can be used for WIA
```
henrik-me: >WIA [](start = 54, length = 3)

nit: IWA #Resolved

msal/src/Microsoft.Identity.Client/Platforms/uap/PlatformInformation.cs
```diff
@@ -49,6 +49,71 @@ public override string GetProductName()
...
+        /// </summary>
+        /// <remarks>
+        /// Win10 allows several identities to be logged in at once; 
+        /// select the first principal name that can be used
+        /// </remarks>
+        /// <returns></returns>
```
henrik-me: returns is probably at least part of what you have in remarks. #Resolved

msal/src/Microsoft.Identity.Client/Platforms/uap/PlatformInformation.cs
```diff
@@ -49,6 +49,71 @@ public override string GetProductName()
...
+        public override async Task<string> GetUserPrincipalNameAsync()
+        {
+            IReadOnlyList<User> users = await User.FindAllAsync();
+            if (users == null || !users.Any())
+            {
+                throw new MsalClientException(
+                 CoreErrorCodes.CannotAccessUserInformationOrUserNotDomainJoined,
+                 CoreErrorMessages.UapCannotFindDomainUser);
```
henrik-me, line 67: nit: should the names of the code and the message be the same? 
aside: we could have a class which held both the code and the message... #Resolved

msal/src/Microsoft.Identity.Client/Platforms/uap/PlatformInformation.cs
```diff
@@ -49,6 +49,71 @@ public override string GetProductName()
...
+            if (userDetailWithPn != null)
+            {
+                return userDetailWithPn.PrincipalName;
+            }
+
+            // user has domain name, but no upn -> missing Enterprise Auth capability
+            if (userDetails.Any(d => !String.IsNullOrWhiteSpace(d.Domain)))
+            {
+                throw new MsalClientException(
+                 CoreErrorCodes.CannotAccessUserInformationOrUserNotDomainJoined,
```
henrik-me: imo, there should be a unique code for each of the cases an exception is thrown. Also, does each of the messages include a link to read more about troubleshooting user lookup or simply to the generic page on IWA, which can then talk about challenges getting user information on UWP/Desktop, etc.  #Resolved

henrik-me:
Looks great. (I got about halfway through doing something whihc is not hearly as good as this). Main ask from me here is to ensure the error messages are improved a bit. Can potentially be another PR if you prefer. #Resolved

msal/tests/dev apps/UWP/MainPage.xaml.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+
+        private async void AcquireTokenIWA_ClickAsync(object sender, RoutedEventArgs e)
+        {
+            AuthenticationResult result = null;
+            try
+            {
+                //result = await _pca.AcquireTokenByIntegratedWindowsAuthAsync(Scopes);
+                 result = await _pca.AcquireTokenByIntegratedWindowsAuthAsync(Scopes, "bogavril@microsoft.com"); // can also use this overload
```
henrik-me: >"bogavril@microsoft.com" [](start = 86, length = 24)

move to field on the form and make a selector on the form to try either with or with-out the username (upn) #WontFix

msal/tests/dev apps/UWP/MainPage.xaml.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        private async void AccessTokenSilentButton_ClickAsync(object sender, RoutedEventArgs e)
+        {
+            var accounts = await _pca.GetAccountsAsync().ConfigureAwait(false);
+
+            AuthenticationResult result = null;
+            try
+            {
+                result = await _pca.AcquireTokenSilentAsync(Scopes, accounts.FirstOrDefault());
```
henrik-me, line 83: >accounts.FirstOrDefault() [](start = 68, length = 25)

perhaps we should do an attempt to actually find the account mentioned in the upn field, instead of grabbing what comes first? #WontFix

henrik-me (changes requested):
:clock1:

msal/src/Microsoft.Identity.Client/Platforms/uap/PlatformInformation.cs
bgavrilMS: Will use the same error message, good point. Will look at 1 class for code + message later.

---
In reply to: [220986182](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1276#discussion_r220986182) [](ancestors = 220986182)

henrik-me:
removing block, expecting an update to the errors, etc. in another PR #Resolved

henrik-me (approved):

jennyf19 (approved):
:shipit:

PR #1279, opened 2018-09-27
Update changelog

bgavrilMS:


adal/changelog.txt
```diff
@@ -1,3 +1,17 @@
+Version 4.1.0-preview
+=============
+
+This release includes: 
+
+- Integrated Windows Auth and Username / Password minor changes, such as improved error messages and http call resilience
+
+Version 4.0.1-preview
```
jennyf19: Do we have to mention this even if we didn't release it to nuget? Would be something to add to the release process if we need to make note of msal changes here.

jennyf19 (approved):
🚀

PR #1281, opened 2018-09-27
Fix handshake w/lab

jennyf19:
The lab was not accepting the handshake, so added in ServicePointManger.

### DIFF
--- a/msal/tests/Test.MSAL.NET.Integration/UsernamePasswordIntegrationTests.cs
+++ b/msal/tests/Test.MSAL.NET.Integration/UsernamePasswordIntegrationTests.cs
@@ -42,6 +42,12 @@ namespace Test.MSAL.NET.Integration
         public const string Authority = "https://login.microsoftonline.com/organizations/";
         public string[] Scopes = { "User.Read" };
         AuthHelper authHelper = new AuthHelper();
+        
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12 | SecurityProtocolType.Ssl3;
+        }
 
         [TestMethod]
         [TestCategory("UsernamePasswordIntegrationTests")]
--- a/msal/tests/Test.MSAL.NET.Integration/UsernamePasswordIntegrationTests.cs
+++ b/msal/tests/Test.MSAL.NET.Integration/UsernamePasswordIntegrationTests.cs
@@ -42,11 +42,11 @@ namespace Test.MSAL.NET.Integration
         public const string Authority = "https://login.microsoftonline.com/organizations/";
         public string[] Scopes = { "User.Read" };
         AuthHelper authHelper = new AuthHelper();
-        
-        [TestInitialize]
-        public void TestInitialize()
+
+        [ClassInitialize]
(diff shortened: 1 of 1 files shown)


msal/tests/Test.MSAL.NET.Integration/UsernamePasswordIntegrationTests.cs
```diff
@@ -42,6 +42,12 @@ public class UsernamePasswordIntegrationTests
         public const string Authority = "https://login.microsoftonline.com/organizations/";
         public string[] Scopes = { "User.Read" };
         AuthHelper authHelper = new AuthHelper();
+        
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12 | SecurityProtocolType.Ssl3;
```
henrik-me: >SecurityProtocolType.Ssl3 [](start = 80, length = 25)

Can we remove this one? It should only be using TLS 1.2+ #Resolved

henrik-me (changes requested):
:clock1:

msal/tests/Test.MSAL.NET.Integration/UsernamePasswordIntegrationTests.cs
```diff
@@ -42,6 +42,12 @@ public class UsernamePasswordIntegrationTests
         public const string Authority = "https://login.microsoftonline.com/organizations/";
         public string[] Scopes = { "User.Read" };
         AuthHelper authHelper = new AuthHelper();
+        
+        [TestInitialize]
+        public void TestInitialize()
```
bgavrilMS: >TestInitialize [](start = 20, length = 14)

It seems to me that you want to run this code once for all tests in this assembly? In that case, note that there are 2 other init attributes in mstest - ClassInitialize and AssemblyInitialize. The last seems more appropriat for your needs: [https://stackoverflow.com/questions/2382552/is-it-possible-to-execute-code-once-before-all-tests-run](https://stackoverflow.com/questions/2382552/is-it-possible-to-execute-code-once-before-all-tests-run) #Resolved

bgavrilMS (approved):
:shipit:

msal/tests/Test.MSAL.NET.Integration/UsernamePasswordIntegrationTests.cs
jennyf19: very cool. You're right, i only need this once for the class, so I changed to classInitialize. Thanks for the suggestion. 

---
In reply to: [221246985](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1281#discussion_r221246985) [](ancestors = 221246985)

msal/tests/Test.MSAL.NET.Integration/UsernamePasswordIntegrationTests.cs
jennyf19: had some issues w/keyvault, so this took a little longer to fix.

---
In reply to: [221143624](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1281#discussion_r221143624) [](ancestors = 221143624)

henrik-me (approved):
:shipit:

PR #1284, opened 2018-09-28
Extract platform specific logic to core

bgavrilMS:
During last release we had 1-2 bugs due to PlatformInformation. One part of refactoring this class is to extract common logic (operating system related logic) to core,


core/src/Platforms/Android/PlatformProxy.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        /// <summary>
+        /// Get the user logged in 
+        /// </summary>
+        /// <returns>The username or throws</returns>
+        public async Task<string> GetUserPrincipalNameAsync()
+        {
+            return await Task.Factory.StartNew(() => string.Empty).ConfigureAwait(false);
+
+        }
```
jennyf19: Add line below #Resolved

core/src/Platforms/iOS/PlatformProxy.cs
```diff
@@ -0,0 +1,79 @@
...
+//
+//------------------------------------------------------------------------------
+
+using System.Threading.Tasks;
+using UIKit;
+
+namespace Microsoft.Identity.Core
+{
+    /// <summary>
+    /// Platform / OS specific logic.  No library (ADAL / MSAL) specific code should go in here.
```
jennyf19: >No library (ADAL / MSAL) specific code should go in here. [](start = 39, length = 57)

This comment should go in the Android PlatformProxy as well. #Resolved

core/src/Platforms/iOS/PlatformProxy.cs
```diff
@@ -0,0 +1,79 @@
...
+            return UIDevice.CurrentDevice.SystemVersion;
+        }
+
+        public  string GetDeviceModel()
+        {
+            return UIDevice.CurrentDevice.Model;
+        }
+
+
+    }
```
jennyf19: remove extra lines #Resolved

core/src/Platforms/iOS/PlatformProxy.cs
```diff
@@ -0,0 +1,79 @@
...
+
+using System.Threading.Tasks;
+using UIKit;
+
+namespace Microsoft.Identity.Core
+{
+    /// <summary>
+    /// Platform / OS specific logic.  No library (ADAL / MSAL) specific code should go in here. 
+    /// </summary>
+    internal class PlatformProxy : IPlatformProxy
```
MarkZuber: Could we start naming the classes more specifically?  The intellisense/build break navigation logic gets all confused with 5+ implementations of exactly the same class based on the configuration.  

So, could we call this IosPlatformProxy as a class (and Net45PlatformProxy, etc for others).  Then we can have a single factory class called PlatformProxyFactory() that returns an IPlatformProxy and within that factory we have the #ifdef code to look for the platform we're building for and return the right one.  I think this would be really helpful for our own code navigation and debugging...   #Resolved

henrik-me: +1: imo a good idea, probably means we will have to use interfaces a few places where we are not doing so today. And probably we will have to have a factory...

---
In reply to: [221390351](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1284#discussion_r221390351) [](ancestors = 221390351)

henrik-me:
nit: a few stylecop issues #Resolved

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Constants.cs
```diff
@@ -167,6 +166,9 @@ internal static class AdalErrorMessage
         public const string DeviceCertificateNotFoundTemplate = "Device Certificate was not found for {0}";
 
         public const string InteractionRequired = "interaction_required";
+
+        public const string GetUserNameFailed = "Failed to get user name";
```
henrik-me, line 170: >public const s [](start = 8, length = 14)

not sure if this change is needed (at least I do not understand it) #WontFix

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Platform/PlatformInformationBase.cs
```diff
@@ -42,10 +42,10 @@ public override string GetAssemblyFileVersionAttribute()
...
         }
 
-        public async override Task<bool> IsUserLocalAsync(RequestContext requestContext)
-        {
-            return await Task.Factory.StartNew(() => false).ConfigureAwait(false);
-        }
+        //public async override Task<bool> IsUserLocalAsync(RequestContext requestContext)
+        //{
+        //    return await Task.Factory.StartNew(() => false).ConfigureAwait(false);
+        //}
```
henrik-me: ? #Resolved

bgavrilMS: Dead code, removed (that method is platform specific, so it goes in PlatformProxy)

---
In reply to: [221416098](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1284#discussion_r221416098) [](ancestors = 221416098)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Http/AdalHttpClient.cs
```diff
@@ -200,7 +200,7 @@ private async Task<T> HandleDeviceAuthChallengeAsync<T>(IHttpWebResponse respons
 
         private static string CheckForExtraQueryParameter(string url)
         {
-            string extraQueryParameter = new PlatformInformation().GetEnvironmentVariable("ExtraQueryParameter");
+            string extraQueryParameter = PlatformProxyFactory.GetPlatformProxy().GetEnvironmentVariable("ExtraQueryParameter");
```
henrik-me, line 200: >GetPlatformProxy [](start = 62, length = 16)

this thing creates a new object every time. is this really needed? seems to be waste of cpu cycles #Resolved

core/src/PlatformProxyFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+namespace Microsoft.Identity.Core
+{
+    internal class PlatformProxyFactory
+    {
+        public static IPlatformProxy GetPlatformProxy()
```
henrik-me: >ublic static IPlatformProxy [](start = 9, length = 27)

mentioned elsewhere... imo, it's overhead to new these up everytime as we do this a lot... #Resolved

henrik-me (changes requested):
:clock1:

core/src/PlatformProxyFactory.cs
```diff
@@ -0,0 +1,64 @@
...
+    internal class PlatformProxyFactory
+    {
+        private IPlatformProxy _platformProxy;
+
+        /// <summary>
+        /// Gets the platform proxy, which can be used to perform platform specific operations
+        /// </summary>
+        public IPlatformProxy GetOrCreatePlatformProxy()
+        {
+            if (_platformProxy == null)
```
MarkZuber: This isn't thread safe if we have multiple threads/calls setting this up at the same time.   #Resolved

core/src/PlatformProxyFactory.cs
henrik-me: not sure it's fully there as you are now newing the platformproxyfactory every time... Do you have a next step of this which passes the platformproxy along or are you planning on setting it up in the moduleinitilizer or? or perhaps in each of the places where it's used it needs to be "cached" as well.  but that seems to be a sub-optimal pattern... 

---
In reply to: [221690145](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1284#discussion_r221690145) [](ancestors = 221690145)

core/tests/Test.Microsoft.Identity.Core.Unit/PlatformProxyFactoryTests.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.Identity.Core;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Test.Microsoft.Identity.Core.Unit
+{
+    [TestClass]
+    public class PlatformProxyFactoryTests
+    {
+        [TestMethod]
+        public void PlatformProxyFactoryCachesTheProxy()
```
henrik-me, line 37: >PlatformProxyFactoryCachesTheProxy [](start = 20, length = 34)

imo, we need an object which holds the proxy during the lifetime of the client applications. (perhaps something setup in the module initilizer).  #Resolved

core/src/PlatformProxyFactory.cs
bgavrilMS: I made this a singleton which is thread safe. I know singletons are defeating the SRP principle, but some responsabilities are cross cutting, i.e. logging, exceptions and accessing platform APIs, so this is a good compromise. 

---
In reply to: [221832318](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1284#discussion_r221832318) [](ancestors = 221832318,221690145)

core/tests/Test.Microsoft.Identity.Core.Unit/PlatformProxyFactoryTests.cs
bgavrilMS: Made it a singleton. It doesn't need to be initialized in the module initializer because there is no library specific logic. 

---
In reply to: [221832547](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1284#discussion_r221832547) [](ancestors = 221832547)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Constants.cs
bgavrilMS: git conflict resolution ....

---
In reply to: [221416051](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1284#discussion_r221416051) [](ancestors = 221416051)

PR #1287, opened 2018-10-01
Limit the token cache key to 255 chars on UWP

bgavrilMS:
Bug fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/612
It is easier to review commit by commit!


core/src/Cache/Items/MsalIdTokenCacheItem.cs
```diff
@@ -76,5 +76,16 @@ internal MsalIdTokenCacheKey GetKey()
         {
             return new MsalIdTokenCacheKey(Environment, TenantId, HomeAccountId, ClientId);
         }
+
```
bgavrilMS: I'll remove this. #Resolved

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
```diff
@@ -0,0 +1,167 @@
...
+//
+//------------------------------------------------------------------------------
+
+using System;
+using System.Text;
+
+namespace Microsoft.Identity.Core.Cache
+{
+    /// <summary>
+    /// An object representing the key of the token cache AT dictionary. The
```
bgavrilMS, line 34: I will add a link to the spec here. #WontFix

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
```diff
@@ -0,0 +1,167 @@
...
+using System;
+using System.Text;
+
+namespace Microsoft.Identity.Core.Cache
+{
+    /// <summary>
+    /// An object representing the key of the token cache AT dictionary. The 
+    /// format of the key is not important for this library, as long as it is unique.
+    /// </summary>
+    internal class MsalAccessTokenCacheKey
```
bgavrilMS, line 38: This is just a refactoring + move of an existing class. I decided to remove the inheritance hierarchy because the Keys are just data classes: 

- the 3 level hierarchy makes it more difficult to understand what fields go in each key
- we had a "type" enum in the base class. this is a code smell that inheritance is not a good decision here. 

The spec does mention that all keys share "env" and "clientid". We don't use this detail in our code, but we can just as easy extract an interface. #Resolved

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
```diff
@@ -0,0 +1,167 @@
...
+    {
+        private readonly string _environment;
```
bgavrilMS, line 40: The most important part of refactoring was to make these fields private. As such, our library does NOT use these fields anywhere else. This makes the usage scenarios more clear: 

Supported: 
- construct a key 
- query the store to see if an entry exists based on the key

Not Supported: 
- load all store entries
- filter those that have specific criteria (e.g. a given subset of scopes)
Note that we still perform the above scenario, but we look directly at the entries.  #Resolved

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
```diff
@@ -0,0 +1,167 @@
...
+        private readonly string _homeAccountId;
+        private readonly string _clientId;
+        private readonly string _scopes;
```
bgavrilMS: Todo: I'll add a comment here stating that scopes and env are not added in clear in the UWP key. #Resolved

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
```diff
@@ -0,0 +1,167 @@
...
+        /// <remarks>
+        /// accountId - two guids plus separator - 73 chars        
+        /// "accesstoken" string - 11 chars
+        /// clientid - a guid - 36 chars
+        /// tenantid - a guid - 36 chars
+        /// scopes and env - a sha256 string - 44 chars
+        /// delimiters - 4 chars
+        /// total: 204 chars
+        /// </remarks>
+        public string GetFixedSizeKey()
```
bgavrilMS: This is the actual bug fix. I decided to hash both scopes and env, although we could probably get away with hashing only the scopes. I am not hashing anything in the other keys, since there are no scopes in RT / IdT.

@MarkZuber  @oldalton  - please advise if this is acceptable. I am testing to see the impact on ADAL -> MSAL upgrade, but this is going to be a breaking change in the cache schema.  #Resolved

core/tests/Test.Microsoft.Identity.Core.Unit/AssertException.cs
```diff
@@ -0,0 +1,223 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Diagnostics;
+using System.Globalization;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Test.Microsoft.Identity.Core.Unit
+{
+    public static class AssertException
```
bgavrilMS, line 37: Helper test class - moved from ADAL. Will refactor and remove duplication as part of the work to have multi-platform tests. #Closed

core/tests/Test.Microsoft.Identity.Core.Unit/CacheTests/MsalTokenCacheKeysTests.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        [TestMethod]
+        public void MsalRefreshTokenCacheKey()
+        {
+            MsalRefreshTokenCacheKey key = new MsalRefreshTokenCacheKey("env", "cid", "uid");
+
+            Assert.AreEqual("uid-env-refreshtoken-cid-", key.ToString());
+
+            Assert.AreEqual("uid-env", key.GetiOSAccountKey());
+            Assert.AreEqual("refreshtoken-cid--", key.GetiOSServiceKey()); // not a bug?
```
bgavrilMS: Confirmed with Olga that this is not a bug, I'll remove this comment. #Resolved

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
bgavrilMS: Alternative fix

The way we store the cache on UWP is smth like: 

LocalData.Container["MSAL_AT"][key] = value

We could however store it like (this is what ADAL does):

LocalData.Container["MSAL_AT"] = dictonary_with_all_keys_and_values

Pros: no key schema change
Cons: slower lookup in the cache, but should not be a problem on UWP / Public Client #WontFix

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
```diff
@@ -0,0 +1,167 @@
...
+        {
+            var stringBuilder = new StringBuilder();
+
+            stringBuilder.Append(MsalCacheConstants.AccessToken);
+            stringBuilder.Append(MsalCacheConstants.CacheKeyDelimiter);
+
+            stringBuilder.Append(_clientId);
+            stringBuilder.Append(MsalCacheConstants.CacheKeyDelimiter);
+
+            stringBuilder.Append(_tenantId ?? "");
```
henrik-me: >"" [](start = 46, length = 2)

nit: should these be using string.empty? #WontFix

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
oldalton: Hashing of scopes make sense to me, see alternative proposal below, and let's agree on approach across SDKs. #Resolved

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
```diff
@@ -0,0 +1,167 @@
...
+        {
+            var stringBuilder = new StringBuilder();
+            stringBuilder.Append((_homeAccountId ?? "") + MsalCacheConstants.CacheKeyDelimiter);
+            
+            stringBuilder.Append(MsalCacheConstants.AccessToken + MsalCacheConstants.CacheKeyDelimiter);
+            stringBuilder.Append(_clientId + MsalCacheConstants.CacheKeyDelimiter);
+            stringBuilder.Append((_tenantId ?? "") + MsalCacheConstants.CacheKeyDelimiter);
+
+            // can't use scopes and env because they are of variable length
+            stringBuilder.Append(CoreCryptographyHelpers.CreateSha256Hash(_scopes + _environment));
```
oldalton: I assume input scopes are always in the same order? We previously agreed to use the order that's returned from the server side.  #WontFix

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
```diff
@@ -0,0 +1,167 @@
...
+
+            return stringBuilder.ToString();
+        }
+        #endregion
+
+
+        #region iOS
+
+        public string GetiOSAccountKey()
```
oldalton, line 112: Consider moving type also into here, like GetiOSAccountType #ByDesign

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
```diff
@@ -0,0 +1,167 @@
...
+        /// "accesstoken" string - 11 chars
+        /// clientid - a guid - 36 chars
+        /// tenantid - a guid - 36 chars
+        /// scopes and env - a sha256 string - 44 chars
+        /// delimiters - 4 chars
+        /// total: 204 chars
+        /// </remarks>
+        public string GetFixedSizeKey()
+        {
+            var stringBuilder = new StringBuilder();
```
oldalton: I think this means you're also making a breaking change for Android. While there's no real SSO for Android, Android key components must be consistent across SDKs, so you shouldn't make a breaking change for other platforms. #Resolved

core/src/Cache/Keys/MsalAccountCacheKey.cs
```diff
@@ -26,30 +26,62 @@
...
 
         public override string ToString()
         {
             var stringBuilder = new StringBuilder();
 
-            stringBuilder.Append(HomeAccountId + CacheKeyDelimiter);
-            stringBuilder.Append(Environment + CacheKeyDelimiter);
-            stringBuilder.Append(TenantId);
+            stringBuilder.Append(_homeAccountId + MsalCacheConstants.CacheKeyDelimiter);
+            stringBuilder.Append(_environment + MsalCacheConstants.CacheKeyDelimiter);
```
oldalton: Environment part of the key is now inconsistent between access token cache key and account cache key. 
Let's either only hash scopes, or hash environment consistently across all keys.

Additional alternative for the fix - follow the documented cache key format everywhere, but hash the whole key only for UWP in platform specific code.

In any case, we should agree on the change across SDKs and document the change in schema. #Resolved

core/src/Cache/Keys/MsalAccountCacheKey.cs
```diff
@@ -26,30 +26,62 @@
...
+            stringBuilder.Append(_tenantId);
 
             return stringBuilder.ToString();
         }
+
+        #region iOS
+
+        public string GetiOSAccountKey()
```
oldalton, line 68: Account is also missing generic field #ByDesign

core/src/Cache/Keys/MsalRefreshTokenCacheKey.cs
```diff
@@ -0,0 +1,125 @@
...
+            stringBuilder.Append(MsalCacheConstants.CacheKeyDelimiter);
+
+            stringBuilder.Append(this._environment);
+            stringBuilder.Append(MsalCacheConstants.CacheKeyDelimiter);
+
+            stringBuilder.Append(MsalCacheConstants.RefreshToken);
+            stringBuilder.Append(MsalCacheConstants.CacheKeyDelimiter);
+
+            stringBuilder.Append(_clientId);
+            stringBuilder.Append(MsalCacheConstants.CacheKeyDelimiter);
```
oldalton: you're missing one additional key delimiter here in the end for schema compliance #Resolved

core/src/Cache/Keys/MsalRefreshTokenCacheKey.cs
```diff
@@ -0,0 +1,125 @@
...
+            return stringBuilder.ToString();
+        }
+
+        #region iOS
+
+        public string GetiOSAccountKey()
+        {
+            var stringBuilder = new StringBuilder();
+
+            stringBuilder.Append(_homeAccountId ?? "");
```
oldalton: why would we not have homeAccountId for MSAL tokens (new schema)?  #Resolved

core/src/Cache/Keys/MsalRefreshTokenCacheKey.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            var stringBuilder = new StringBuilder();
+
+            stringBuilder.Append(MsalCacheConstants.RefreshToken);
+            stringBuilder.Append(MsalCacheConstants.CacheKeyDelimiter);
+
+            stringBuilder.Append(_clientId);
+            stringBuilder.Append(MsalCacheConstants.CacheKeyDelimiter);
+
+            // no tenant id, but add the deliminter anyway
```
oldalton: typo: deliminter #Resolved

core/src/Cache/Keys/MsalIdTokenCacheKey.cs
```diff
@@ -0,0 +1,140 @@
...
+            stringBuilder.Append(_clientId);
+            stringBuilder.Append(MsalCacheConstants.CacheKeyDelimiter);
+
+            stringBuilder.Append(_tenantId ?? "");
+            stringBuilder.Append(MsalCacheConstants.CacheKeyDelimiter);
+
+            return stringBuilder.ToString();
+        }
+
+        public string GetiOSGenericKey()
```
oldalton, line 92: same for credentials - consider moving iOS type into here too for consistency #WontFix

core/tests/Test.Microsoft.Identity.Core.Unit/CacheTests/MsalTokenCacheKeysTests.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        [TestMethod]
+        public void MsalAccessTokenCacheKey()
+        {
+            MsalAccessTokenCacheKey key = new MsalAccessTokenCacheKey("env", "tid", "uid", "cid", "scopes");
+
+            Assert.AreEqual("uid-env-accesstoken-cid-tid-scopes", key.ToString());
+
+            Assert.AreEqual("uid-env", key.GetiOSAccountKey());
+            Assert.AreEqual("accesstoken-cid-tid-scopes-", key.GetiOSServiceKey());
```
oldalton: There should be no delimiter in the end, so "accesstoken-cid-tid-scopes" #Closed

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
bgavrilMS: That might be a bit difficult because it is used for other things than keys and I can't reference the TokenAccessor from here because it is compiled only on iOS.

---
In reply to: [221719634](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1287#discussion_r221719634) [](ancestors = 221719634)

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
bgavrilMS: VS doesn't complain and I'd argue it makes the code more readable like this.

---
In reply to: [221705262](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1287#discussion_r221705262) [](ancestors = 221705262)

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
bgavrilMS: This key computation will only be used on UWP. The rest of the platforms will use the long key. I renamed the method to make it clearer.

---
In reply to: [221720078](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1287#discussion_r221720078) [](ancestors = 221720078)

core/src/Cache/Keys/MsalAccountCacheKey.cs
bgavrilMS: Good point. I will hash only the scopes - this will give us a bit more flexibility in case we want to filter keys by env or client id etc. This only happen on UWP, there is no reason to change the logic for other platforms.

 In any case, since all the key components are also in the value, we can always write a one time migration code in case we need to make further changes. 

---
In reply to: [221720606](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1287#discussion_r221720606) [](ancestors = 221720606)

core/src/Cache/Keys/MsalRefreshTokenCacheKey.cs
bgavrilMS: Existing code was like this. I see 2 possibilities for this confidential client flows that does not involve client and ADAL v3 where we do not have the idtoken.

---
In reply to: [221721468](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1287#discussion_r221721468) [](ancestors = 221721468)

core/src/Cache/Keys/MsalRefreshTokenCacheKey.cs
bgavrilMS: Yes, good catch. 

---
In reply to: [221721226](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1287#discussion_r221721226) [](ancestors = 221721226)

core/src/Cache/Keys/MsalIdTokenCacheKey.cs
bgavrilMS: That type seems to be used in many places and the current implementation (an enum) is pretty readable as is. 

---
In reply to: [221722252](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1287#discussion_r221722252) [](ancestors = 221722252)

core/src/Cache/Keys/MsalAccountCacheKey.cs
bgavrilMS: It is because it is set to local_account_id which is not part of the account key object.

---
In reply to: [221720840](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1287#discussion_r221720840) [](ancestors = 221720840)

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
bgavrilMS: Thanks for bringing this up, because the current implementation is not in line with your agreement. The scopes are lowercased and ordered. Assuming the iOS version has shipped, this means we are incompatible :(. We should follow this up separately. 

---
In reply to: [221718990](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1287#discussion_r221718990) [](ancestors = 221718990)

core/src/Cache/Keys/MsalAccountCacheKey.cs
oldalton: Hmm, I think I saw generic in some credential cache keys. Why we would have it differently between accounts and credentials? #Resolved

core/src/Cache/Keys/MsalAccountCacheKey.cs
oldalton: @bgavrilMS Although, it only affects UWP, we should still document the change. Previous spec assumed same keys for all platforms unless described separately. #Resolved

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
oldalton: We have found that inconsistency during the compatibility testing before and Petro fixed it, if I remember correctly. Let's follow up on that. #Resolved

core/src/Cache/Keys/MsalAccessTokenCacheKey.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// <summary>
+        /// Gets a key that is smaller than 255 characters, which is a limitation for 
+        /// UWP storage. This is done by hashing the scopes and env.
+        /// </summary>
+        /// <remarks>
+        /// accountId - two guids plus separator - 73 chars        
+        /// "accesstoken" string - 11 chars
+        /// clientid - a guid - 36 chars
+        /// tenantid - a guid - 36 chars
+        /// scopes and env - a sha256 string - 44 chars
```
oldalton, line 93: you should update this comment if you only hash scopes #Resolved

core/src/Cache/Keys/MsalAccountCacheKey.cs
bgavrilMS: @oldalton@microsoft.com - already updated the spec. 

---
In reply to: [222163772](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1287#discussion_r222163772) [](ancestors = 222163772)

core/src/Cache/Keys/MsalAccountCacheKey.cs
bgavrilMS: Yes, you are right, added.

---
In reply to: [222163627](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1287#discussion_r222163627) [](ancestors = 222163627)

bgavrilMS:
Merged manually, closing this PR.

PR #1288, opened 2018-10-01
(554545) Update logging infra to work towards reducing double logging

MarkZuber:
Currently, we have a lot of code that looks like this:

                msg = "Client Info is missing. Skipping MSAL refresh token cache write";
                CoreLoggerBase.Default.Info(msg);
                CoreLoggerBase.Default.InfoPii(msg);

This results in:
•	The developer having to write this twice (more code to write/review maintain)
•	Two writes / callbacks for the same content (performance)
•	Lots of double-logs printed out in the log files (makes logs harder to read)

There are two cases I see in our code for logging:
1.	We have only non-PII data to log
2.	We have PII data to log and a backup message to show that has the PII stripped out of it

This PR is the API surface change and root implementation.  I'll have follow up pull requests to fix up the code to use the updated API and then reduce the API so that *PII methods always take two strings.


bgavrilMS (approved):
:shipit:

jennyf19 (approved):
:shipit:

PR #1289, opened 2018-10-01
(554545) Remove double logging

MarkZuber:
Please pay special attention to changes made in LoggerCallbackTests.cs and AdalLoggerBase::Log() to ensure behavior is consistent with legacy.

### DIFF
--- a/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationParameters.cs
+++ b/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationParameters.cs
@@ -151,8 +151,6 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
                 await request.GetResponseAsync().ConfigureAwait(false);
 
                 var ex = new AdalException(AdalError.UnauthorizedResponseExpected);
-                string noPiiMsg = CoreExceptionFactory.Instance.GetPiiScrubbedDetails(ex);
-                CoreLoggerBase.Default.Error(noPiiMsg);
                 CoreLoggerBase.Default.ErrorPii(ex);
                 throw ex;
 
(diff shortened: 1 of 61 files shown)


bgavrilMS:
While having a look at the MSALLogger, I noticed that one of the methods throws an exception - is this expected? If not, probably missing a unit test as well.. Please check ADALLogger as well. Otherwise the PR looks good!

 public override void InfoPii(string messageWithPii, string messageScrubbed)
        {
            throw new NotImplementedException();
        }

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationParameters.cs
```diff
@@ -163,8 +161,6 @@ private static async Task<AuthenticationParameters> CreateFromResourceUrlCommonA
                 if (response == null)
                 {
                     var serviceEx = new AdalServiceException(AdalErrorMessage.UnauthorizedHttpStatusCodeExpected, ex);
-                    string noPiiMsg = CoreExceptionFactory.Instance.GetPiiScrubbedDetails(serviceEx);
-                    CoreLoggerBase.Default.Error(noPiiMsg);
```
bgavrilMS, line 167: >noPiiMsg [](start = 49, length = 8)

Tedious work, thanks for doing this! #ByDesign

bgavrilMS (approved):
:shipit:

MarkZuber:
>           While having a look at the MSALLogger, I noticed that one of the methods throws an exception - is this expected? If not, probably missing a unit test as well.. Please check ADALLogger as well. Otherwise the PR looks good!
> public override void InfoPii(string messageWithPii, string messageScrubbed)
> {
> throw new NotImplementedException();
> }
You may have been looking at an intermediary commit.  I've re-reviewed the code and I don't see any NotImplementedExceptions being thrown within any of the loggers.

PR #1290, opened 2018-10-01
First round at removing static httprequest layer, moving to HttpManager

MarkZuber:
Create a new class, HttpManager and its corresponding interface, IHttpManager.  The intent is to replace the static HttpRequest class with this object and bound usage of this to IHttpManager so we can simplify our http layer mocking.  This currently gets us a non-static version of HttpManager and places its implementation under HttpRequest's static calls as the first step.

### DIFF
--- a/core/src/Exceptions/CoreExceptionFactory.cs
+++ b/core/src/Exceptions/CoreExceptionFactory.cs
@@ -29,11 +29,42 @@ using System;
 
 namespace Microsoft.Identity.Core
 {
+    internal interface ICoreExceptionFactory
+    {
+        Exception GetClientException(
+            string errorCode,
+            string errorMessage,
+            Exception innerException = null);
+
+        Exception GetServiceException(
+            string errorCode,
+            string errorMessage);
+
+        Exception GetServiceException(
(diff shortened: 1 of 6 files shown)


core/src/Exceptions/CoreExceptionFactory.cs
```diff
@@ -29,11 +29,42 @@
 
 namespace Microsoft.Identity.Core
 {
+    internal interface ICoreExceptionFactory
```
henrik-me: >internal interface ICoreExceptionFactory [](start = 4, length = 40)

separate file?

henrik-me:
aside: as we move away from non-static (good) then we should still ensure the httpc;lient is re-used for the lifetime of our own clientapplication objects

henrik-me (approved):
:shipit:

bgavrilMS (approved):
:shipit:

core/src/Exceptions/CoreExceptionFactory.cs
MarkZuber: fixed

PR #1292, opened 2018-10-02
Jennyf/ios webview

jennyf19:
**Tested Interactive and Cancel with**
> XS iOS 12 Simulator
> SE iOS 11 Simulator
> 6s iOS 9.2 Simulator

Corresponding [MSAL issue](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/489)


core/src/Platforms/iOS/SystemWebview/SystemWebUI.cs
```diff
@@ -73,23 +74,77 @@ public void Authenticate(Uri authorizationUri, Uri redirectUri, UIViewController
...
+                    asWebAuthenticationSession = new AuthenticationServices.ASWebAuthenticationSession(new NSUrl(authorizationUri.AbsoluteUri),
+                        redirectUri.Scheme, (callbackUrl, error) =>
+                        {
+                            if (error != null)
+                            {
+                                ProcessCompletionHandlerError(error);
+                            }
+                            else
+                            {
+                                vc.InvokeOnMainThread(() =>
```
MarkZuber: If we're always calling ContinueAuthentication on the main thread, should we move the vc.InvokeOnMainThread() to within the ContinueAuthentication function?

core/src/Platforms/iOS/SystemWebview/SystemWebUI.cs
```diff
@@ -73,23 +74,77 @@ public void Authenticate(Uri authorizationUri, Uri redirectUri, UIViewController
...
-                    "Failed to invoke SFSafariViewController", 
+                    CoreErrorCodes.AuthenticationUiFailedError,
+                    "Failed to invoke SFSafariViewController",
                     ex);
             }
         }
 
+        public void ProcessCompletionHandlerError(NSError error)
+        {
+            if (returnedUriReady != null)
```
MarkZuber, line 135: learning for me:  I see that this sets the authorizationResult onto the class, so someone can look at it.  But how does this trigger an error/exception callback to the caller that their async operation has failed?

also, what is returnedUriReady, how/when can it be null, and does the user not get notified of a failure if it is?

core/src/Platforms/iOS/SystemWebview/SystemWebUI.cs
jennyf19: I moved the ViewController to webview base so the vc.InvokeOnMainThread() can be moved to CoutinueAuthentication. I tested it w/all three OS versions, and it works fine. Let me know what you think.

core/src/Platforms/iOS/SystemWebview/SystemWebUI.cs
jennyf19: the authResult gets pushed back through the code to the InteractiveRequest, where it is verified (VerifiyAuthorizationResult) and if the result has AuthorizationStatus.UserCancel, we throw an MsalClientException there.

returnedUriReady is a semaphoreSlim, and if it's null, it goes back to the app in the AppDelegate, OpenUrl.

MarkZuber: Awesome.  It's not clear just from reading this method how the error propagates.  Perhaps adding a comment here explaining where this propagates to the exception throw would be helpful.

core/src/Platforms/iOS/WebviewBase.cs
```diff
@@ -28,14 +28,19 @@
...
+using AuthenticationServices;
+using UIKit;
 
 namespace Microsoft.Identity.Core.UI
 {
     internal abstract class WebviewBase : NSObject, IWebUI, ISFSafariViewControllerDelegate
     {
         protected static SemaphoreSlim returnedUriReady;
         protected static AuthorizationResult authorizationResult;
+        protected static UIViewController viewController;
```
MarkZuber, line 40: should viewController be static?  the other view controllers and sessions are instance-level and not static.

core/src/Platforms/iOS/SystemWebview/SystemWebUI.cs
jennyf19: done.

core/src/Platforms/iOS/WebviewBase.cs
jennyf19: if I make it not static...I have to go back through all the static methods and classes and fix those as well. lots of things blowing up. I could make a task for this though as a separate PR. let's discuss further.

MarkZuber: I only asked because the other view controllers weren't static.  if this blows things up based on current design, ignore for now and you've done the right thing here.

MarkZuber (approved):

core/src/Platforms/iOS/WebviewBase.cs
jennyf19: ok...sounds good. I'll leave for now. Thanks.

PR #1293, opened 2018-10-03
Adfs msal changes

reed1995:
Adfs authority enabled in MSAL


henrik-me:
Please remove all formatting changes. And also pls. ensure that you only change what is required for ADFS. If you are up for formatting and other changes please do so in a separate PR. It's difficult to find the important changes. #Closed

henrik-me:
Aside: Expect some bigger changes coming very soon thus the sooner you get your changes in to the repo the better #Closed

core/src/Instance/AdfsAuthority.cs
```diff
@@ -66,28 +62,10 @@ protected override async Task<string> GetOpenIdConfigurationEndpointAsync(string
...
-                        CoreErrorCodes.MissingPassiveAuthEndpoint,
-                        CoreErrorMessages.CannotFindTheAuthEndpont, 
-                        ExceptionDetail.FromDrsResponse(drsResponse));                        
-                }
-
                 string resource = string.Format(CultureInfo.InvariantCulture, CanonicalAuthority);
                 string webfingerUrl = string.Format(CultureInfo.InvariantCulture,
-                    "https://{0}/adfs/.well-known/webfinger?rel={1}&resource={2}",
-                    drsResponse.IdentityProviderService.PassiveAuthEndpoint.Host,
+                    "https://{0}/.well-known/webfinger?rel={1}&resource={2}",
```
henrik-me, line 74: >https://{0}/.well-known/webfinger?rel={1}&resource={2}", [](start = 21, length = 56)

is this backwards compatible? same goes for the other changes in this file. #Closed

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -345,13 +365,25 @@ private MsalAccessTokenCacheItem FindAccessTokenCommon
...
+                   tokenCacheItems.Where(
+                           item => item.ScopeSet.ScopeContains(requestParams.Scope));
+
+                //Adfs does not return scopes in resource/scope format
+                if (requestParams.Authority.AuthorityType == Core.Instance.AuthorityType.Adfs && filteredItems.Count()==0)
+                {
+                    SortedSet<string> scopes = new SortedSet<string>();
+                    foreach (string scope in requestParams.Scope)
+                    {
+                        scopes.Add(scope.Substring(scope.LastIndexOf("/") + 1));
```
henrik-me: >scope.Substring(scope.LastIndexOf("/") + 1) [](start = 35, length = 43)

this looks magical. Can you pls. show an example of what the resource/scope then looks like in ADFS? (could be made more readable by having a function which does this and which is named elegantly, perhaps with related xml comments if the name is not enough.

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -345,13 +365,25 @@ private MsalAccessTokenCacheItem FindAccessTokenCommon
...
-                    tokenCacheItems.Where(
-                            item =>
-                                item.ScopeSet.ScopeContains(requestParams.Scope));
-
-                requestParams.RequestContext.Logger.Info("Matching entry count after filtering by scopes - " + filteredItems.Count());
+    
+
+                IEnumerable<MsalAccessTokenCacheItem> filteredItems =
+                   tokenCacheItems.Where(
+                           item => item.ScopeSet.ScopeContains(requestParams.Scope));
```
henrik-me: Should then only do this in case it's not ADFS.... seems like waste of CPU cycles. #Closed

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -345,13 +365,25 @@ private MsalAccessTokenCacheItem FindAccessTokenCommon
                 }
 
                 requestParams.RequestContext.Logger.Info("Matching entry count -" + tokenCacheItems.Count);
-
-                IEnumerable<MsalAccessTokenCacheItem> filteredItems =
-                    tokenCacheItems.Where(
-                            item =>
-                                item.ScopeSet.ScopeContains(requestParams.Scope));
-
-                requestParams.RequestContext.Logger.Info("Matching entry count after filtering by scopes - " + filteredItems.Count());
```
henrik-me: > requestParams.RequestContext.Logger.Info("Matching entry count after filtering by scopes - " + filteredItems.Count()); [](start = 15, length = 119)

what's the reason to remove the log statement?

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -366,7 +398,8 @@ private MsalAccessTokenCacheItem FindAccessTokenCommon
                 {
                     filteredItems = filteredItems.Where(
                         item => environmentAliases.Contains(item.Environment) &&
-                        item.TenantId.Equals(requestParams.Authority.GetTenantId(), StringComparison.OrdinalIgnoreCase));
+                        ((item.TenantId==null && requestParams.Authority.GetTenantId()==null) ||
```
henrik-me: >((item.TenantId==null && requestParams.Authority.GetTenantId()==null) || [](start = 24, length = 72)

this is what it is for ADFS if  I read the code correctly. I believe the missing tenant id will cause misses on the cache, is that expected for ADFS?

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -195,9 +195,10 @@ private void SaveTokenResponseToCache()
 
             ClientInfo fromServer = null;
 
-            if (!AuthenticationRequestParameters.IsClientCredentialRequest)
+            if (!AuthenticationRequestParameters.IsClientCredentialRequest && AuthenticationRequestParameters.Authority.AuthorityType != Core.Instance.AuthorityType.Adfs)
             {
                 //client_info is not returned from client credential flows because there is no user present.
+                //Adfs does not return client_info
```
henrik-me: >//Adfs does not return client_info [](start = 16, length = 34)

as mentioned elsewhere, not sure how the cache will work... I bet it will not.

msal/tests/Test.MSAL.NET.Unit/CacheTests/TokenCacheTests.cs
```diff
@@ -956,6 +956,38 @@ public void CacheB2CTokenTest()
...
+                RequestContext = requestContext,
+                Authority = authority,
+                ClientId = TestConstants.ClientId,
+                TenantUpdatedCanonicalAuthority = TestConstants.OnPremiseAuthority
+            };
+
+            AdfsCache.SaveAccessAndRefreshToken(requestParams, response);
+
+            Assert.AreEqual(1, AdfsCache.tokenCacheAccessor.RefreshTokenCacheDictionary.Count);
+            Assert.AreEqual(1, AdfsCache.tokenCacheAccessor.AccessTokenCacheDictionary.Count);
```
henrik-me: > Assert.AreEqual(1, AdfsCache.tokenCacheAccessor.AccessTokenCacheDictionary.Count); [](start = 11, length = 83)

Can the token which is now in the cache actually be used? Did you test on an ADFS environment? Which versions of ADFS?

core/src/Instance/AdfsAuthority.cs
```diff
@@ -50,12 +50,8 @@ public AdfsAuthority(string authority, bool validateAuthority) : base(authority,
 
         protected override bool ExistsInValidatedAuthorityCache(string userPrincipalName)
         {
-            if (string.IsNullOrEmpty(userPrincipalName))
-            {
-                throw CoreExceptionFactory.Instance.GetClientException(
-                    CoreErrorCodes.UpnRequired,
-                    CoreErrorMessages.UpnRequiredForAuthroityValidation);
-            }
+            if (userPrincipalName == null)
```
MarkZuber: nit: please enclose all statements in {}, even one liners like this one. #Closed

MarkZuber: should use string.IsNullOrEmpty() here instead of just a null comparison against a string. #Closed

core/src/Instance/AdfsAuthority.cs
reed1995: There will only be compatability with Adfs 2019...older versions do not support the resource/scope format

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
reed1995: Caching worked in my tested scenarios. I populate the HomeAccountID with the subjec tin the ID token and leave the tenantId null

msal/src/Microsoft.Identity.Client/TokenCache.cs
reed1995: All adfs instances will have null tenant Id's, and I beleive this was accounted for in cachine logic. Basically, a null tenant id in the cahce will match the null tenant id on the incoming request

reed1995:
Not sure why weird formatting changes are being caught...removing them now #Closed

henrik-me:
@bgavrilMS @SomkaPe : pls. help review, especially the changes around the cache.

core/src/Instance/AdfsAuthority.cs
henrik-me: hmm, why not have support for previous versions? Almost no-one uses ADFS 2019 at this point in time. At least 2016 should be supported imo.

---
In reply to: [223864165](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1293#discussion_r223864165) [](ancestors = 223864165)

core/src/Instance/Authority.cs
```diff
@@ -162,6 +160,27 @@ internal static AuthorityType GetAuthorityType(string authority)
             }
         }
 
+        private static Authority CreateInstance(CorePlatformInformationBase platformInformation, string authority, bool validateAuthority)
```
henrik-me: where is this used?

core/tests/Test.Microsoft.Identity.Core.Unit/InstanceTests/AdfsAuthorityTests.cs
```diff
@@ -388,9 +287,9 @@ public void FailedTenantDiscoveryMissingEndpointsTest()
                 }).GetAwaiter().GetResult();
                 Assert.Fail("validation should have failed here");
             }
-            catch (TestServiceException exc)
+            catch (Exception exc)
```
henrik-me: >Exception [](start = 19, length = 9)

can this be a more specific exception? Things could fail for other reasons. Also the assert seems irrelevant.

core/src/Instance/AdfsAuthority.cs
```diff
@@ -66,28 +64,10 @@ protected override async Task<string> GetOpenIdConfigurationEndpointAsync(string
         {
             if (ValidateAuthority)
             {
-                DrsMetadataResponse drsResponse = await GetMetadataFromEnrollmentServerAsync(userPrincipalName, requestContext).ConfigureAwait(false);
```
henrik-me: >DrsMetadataResponse drsResponse = await GetMetadataFromEnrollmentServerAsync( [](start = 16, length = 77)

why remove this?

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -58,8 +58,16 @@ internal AuthenticationResult(MsalAccessTokenCacheItem msalAccessTokenCacheItem,
             _msalIdTokenCacheItem = msalIdTokenCacheItem;
             if (_msalAccessTokenCacheItem.HomeAccountId != null)
             {
-                Account = new Account(AccountId.FromClientInfo(_msalAccessTokenCacheItem.ClientInfo),
-                    _msalIdTokenCacheItem?.IdToken?.PreferredUsername, _msalAccessTokenCacheItem.Environment);
+                if (_msalAccessTokenCacheItem.TenantId == null)
```
henrik-me: > if (_msalAccessTokenCacheItem.TenantId == null) [](start = 15, length = 48)

not sure how this will work in non ADFS cases. @somska @bgavrilMS, thoughts on this?

msal/tests/Test.MSAL.NET.Unit/ConfidentialClientApplicationTests.cs
```diff
@@ -264,6 +270,63 @@ public void ConfidentialClientUsingSecretTest()
...
+            Assert.AreEqual(1, app.AppTokenCache.tokenCacheAccessor.AccessTokenCacheDictionary.Count);
+            Assert.AreEqual(0, app.AppTokenCache.tokenCacheAccessor.RefreshTokenCacheDictionary.Count); //no refresh tokens are returned
+
+            Assert.IsTrue(HttpMessageHandlerFactory.IsMocksQueueEmpty, "All mocks should have been consumed");
+
+            //call AcquireTokenForClientAsync again to get result back from the cache
+            task = app.AcquireTokenForClientAsync(TestConstants.Scope.ToArray());
+            result = task.Result;
+            Assert.IsNotNull(result);
+            Assert.IsNotNull("header.payload.signature", result.AccessToken);
```
henrik-me: this seems like an odd assert. is this intentional?

msal/src/Microsoft.Identity.Client/AccountId.cs
```diff
@@ -80,6 +80,14 @@ public AccountId(string identifier, string objectId, string tenantId)
             ValidateId();
         }
 
+        /// <summary>
+        /// Constructor of an AccountId meant for Adfs scenarios since Adfs instances lack tennt id's.
```
bgavrilMS: >tennt id's [](start = 91, length = 10)

typos (tennt / ids) Also CC @jmprieur as he reviews all public docs. I wonder if the input param of this ctor should be adfsIdentifier to make intent more clear...

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
bgavrilMS: @somska can confirm, but AFAIK clients migrated from ADAL v3 will have ClientInfo null => tenantId is null. 
@reed1995 - will ADFS always have an UPN there? Maybe we could put UPN in the condition?

---
In reply to: [227041247](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1293#discussion_r227041247) [](ancestors = 227041247)

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -58,8 +58,16 @@ internal AuthenticationResult(MsalAccessTokenCacheItem msalAccessTokenCacheItem,
...
+                {
+                    //Null tenant is indicates Adfs authority
+                    Account = new Account(new AccountId(_msalAccessTokenCacheItem.HomeAccountId), _msalIdTokenCacheItem?.IdToken.Upn, _msalAccessTokenCacheItem.Environment);
```
bgavrilMS: >IdToken [](start = 121, length = 7)

THis means you are sure the IdToken cannot be null. If so, please leave as is, we'll get a NullRef exception which is easy to track

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -156,21 +158,30 @@ internal void OnBeforeWrite(TokenCacheNotificationArgs args)
...
             }
 
             lock (LockObject)
             {
                 try
                 {
+                    Account account = null;
+                    if(msalAccessTokenCacheItem.HomeAccountId != null)
+                    {
+                        if(requestParams.Authority.AuthorityType == Core.Instance.AuthorityType.Adfs)
```
bgavrilMS: >if [](start = 24, length = 2)

I know that this isn't your code, but could you: 

a. Move this block to a private method (this one is too large)
b. Use a switch (Authority) here instead of if / else. Similar to the switch in Authority.cs you've added.

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -346,9 +366,21 @@ private MsalAccessTokenCacheItem FindAccessTokenCommon
...
-                IEnumerable<MsalAccessTokenCacheItem> filteredItems =
-                    tokenCacheItems.Where(item => ScopeHelper.ScopeContains(item.ScopeSet, requestParams.Scope));
+    
+
+                IEnumerable<MsalAccessTokenCacheItem> filteredItems =
+                   tokenCacheItems.Where(
+                           item => ScopeHelper.ScopeContains(item.ScopeSet, requestParams.Scope));
+
+                //Adfs does not return scopes in resource/scope format
+                if (requestParams.Authority.AuthorityType == Core.Instance.AuthorityType.Adfs && filteredItems.Count()==0)
```
bgavrilMS: >Count [](start = 111, length = 5)

NIT: use Any()

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -365,7 +397,8 @@ private MsalAccessTokenCacheItem FindAccessTokenCommon
                 {
                     filteredItems = filteredItems.Where(
                         item => environmentAliases.Contains(item.Environment) &&
-                        item.TenantId.Equals(requestParams.Authority.GetTenantId(), StringComparison.OrdinalIgnoreCase));
+                        ((item.TenantId==null && requestParams.Authority.GetTenantId()==null) ||
```
bgavrilMS: >TenantId [](start = 31, length = 8)

I think the code would be clearer if we put an IsADFS { get { return tenantId==null }} property on the token cache items. We're making this assumption everywhere, so it's breaking SRP.

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
bgavrilMS: The problem is that an ADALv3 user will also have an empty client_info ... This will be a pretty complex test case...

---
In reply to: [223772308](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1293#discussion_r223772308) [](ancestors = 223772308)

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
```diff
@@ -32,17 +32,17 @@
...
-
-namespace Test.MSAL.NET.Integration
-{
-    [TestClass]
-    public class DeviceCodeIntegrationTests
-    {
+
+namespace Test.MSAL.NET.Integration
+{
+    [TestClass]
```
bgavrilMS: >TestClass [](start = 5, length = 9)

no changes, please remove

msal/tests/Test.MSAL.NET.Unit/ConfidentialClientApplicationTests.cs
bgavrilMS: assert.equals?

---
In reply to: [227043557](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1293#discussion_r227043557) [](ancestors = 227043557)

core/src/Instance/AdfsAuthority.cs
reed1995: I think support for 2016 is possible. I'd like to get these changes in ASAP (next couple of days) as frequent updates the library have made confilct resolution very painful, so maybe do 2016 compatability as a separate PR?

msal/src/Microsoft.Identity.Client/TokenCache.cs
reed1995: Condensed code, but didn't add new function

core/src/Instance/AdfsAuthority.cs
reed1995: This endpoint doesn't exist for adfs

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
reed1995: @bgavrilMS is that relevant for the purposes of this PR? Currently it looks like the client is reconstructed for all public clients

reed1995: So that case would also curentyl lead to failures (I was getting exceptions thrown since client info was null)

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
reed1995: Added an IsAdfs property on the token cache item and use that in the conditional.

msal/src/Microsoft.Identity.Client/TokenCache.cs
reed1995: The MSAL library adds the scopes in this format if none are present in the response; shouldn't happen, but technically is a possibility

msal/src/Microsoft.Identity.Client/TokenCache.cs
reed1995: Accidnet

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -1120,8 +1147,43 @@ internal void ClearAdalCache()
...
+            try
+            {
+                TokenCacheNotificationArgs args = new TokenCacheNotificationArgs
+                {
+                    TokenCache = this,
+                    ClientId = ClientId,
+                    Account = null
+                };
+
+                OnBeforeAccess(args);
```
bgavrilMS: >OnBeforeAccess [](start = 16, length = 14)

Why do you need this? Do you call ClearMsalCache on its own? I believe you should call Clear() that will call both ClearMsalCache and ClearAdalCache (this is to ensure that the unified cache is consistent). I agree that ClearMsalCache and ClearAdalCache should have been made private and not internal, as they are not meant to called from outside.

bgavrilMS: From my point of view, this is the only outstanding issue. 

---
In reply to: [229642843](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1293#discussion_r229642843) [](ancestors = 229642843)

reed1995: This is not my change...might be something from a merge or accidental formatting change

bgavrilMS (approved):
:shipit:

SomkaPe (commented):
did we do integration testing of this change? There is some code which is written based on assumption that 
account is is always derived from client_info so in particular format, so when existing code will try to read account with id in different format there will be null pointer exception.
Also account object for adfs should be persisted with authority_type = ADFS  according to unified schema.
It would be nice to have cache format unit test for adfs as well.
I do not think this change it ready for merge to dev

core/src/IdToken.cs
```diff
@@ -43,6 +43,7 @@ internal class IdTokenClaim
         public const string PreferredUsername = "preferred_username";
         public const string Name = "name";
         public const string HomeObjectId = "home_oid";
+        public const string Upn = "upn";
```
SomkaPe, line 46: This is v2 IdToken - it does not contain "upn" claim, 
that claim is from v1 idToken

core/src/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -93,6 +93,9 @@ internal MsalAccessTokenCacheItem
         [DataMember(Name = "access_token_type")]
         internal string TokenType { get; set; }
 
+        [DataMember(Name = "is_adfs")]
```
SomkaPe, line 96: unified schema does not contains this field, account property authority_type should be used instead

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -1121,9 +1146,26 @@ internal void ClearAdalCache()
...
-        {
-            TokenCacheAccessor.Clear();
+
+        internal void ClearMsalCache()
+        {
+            TokenCacheAccessor.Clear();
+        }
+
+
+        internal SortedSet<string> ParseScopesForAdfsToken(SortedSet<string> scopes)
```
SomkaPe, line 1156: Not sure that ADFS v4 scope support was tested comprehensively, i did not manage to get at for resource from Lab ADFS V4 env 

based on "Adfs tokens return provided scopes without using the resource/scope format of AAD" I assume that .default scope is not supported, Is it?
If yes it is not consistent with AAD and  will need additional guidance for developers.

Taking to account that ADFS support is relevant for all SDK we have to have design discussions , agree on design and implement it according to it.

SomkaPe (changes requested):

core/src/IdToken.cs
reed1995: Would unique_name be better?

msal/src/Microsoft.Identity.Client/TokenCache.cs
reed1995: What was the request you sent? There are some limitations regarding adfs supportability (only a single resource is allowed)

reed1995: Were u testing against Adfs 2016? Only Adfs 2019 is supported in this PR

core/src/Cache/Items/MsalAccessTokenCacheItem.cs
reed1995: Are you suggesting I just store the authority_type here?

core/src/IdToken.cs
SomkaPe: Msal is SDK for V2 endpoints, so it is supposed to receive idTokens in v2 format (which is more aligned with OpenID Connect spec) https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens

msal/src/Microsoft.Identity.Client/TokenCache.cs
SomkaPe: yes, i have used adfs v4

henrik-me:
@reed1995 : The MSAL repo moved here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet

henrik-me:
The change for MSAL is done here:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/834

PR #1294, opened 2018-10-03
Review service exceptions and add Http headers to them

bgavrilMS:


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Exceptions/AdalExceptionFactory.cs
```diff
@@ -40,44 +41,58 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
...
+            string errorMessage,
             Exception innerException = null)
         {
             ValidateRequiredArgs(errorCode, errorMessage);
             return new AdalException(errorCode, errorMessage);
         }
-      
-        public override Exception GetServiceException(string errorCode, string errorMessage)
+
+        public override Exception GetServiceException(string errorCode, string errorMessage, IHttpWebResponse httpResponse)
```
bgavrilMS, line 52: I decided that we are too lax in throwing service exceptions and we throw service exceptions where we should not. So I am making it mandatory to have an http response or an exception detail object when throwing a service exception. If those aren't available, it's probably a client exception :) #Closed

bgavrilMS: This also ensures that we put the headers in the exception. #Closed

jennyf19: that's a great change. we had a customer issue on this a month or two ago (throwing a service when it should have been client). nice addition.

---
In reply to: [222418975](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1294#discussion_r222418975) [](ancestors = 222418975)

core/src/Exceptions/CoreErrorMessages.cs
```diff
@@ -91,5 +91,7 @@ internal class CoreErrorMessages
 
         public const string GetUserNameFailed = "Failed to get user name from the operating system.";
 
+        public const string OAuthNonJsonError = "An error response was returned by the OAuth2 server. Please inspect the exception properties for details.";
```
jennyf19, line 94: >OAuthNonJsonError [](start = 28, length = 17)

When would we get this specific error? The "NonJson" part seems a little strange. #Closed

core/src/Exceptions/ICoreExceptionFactory.cs
```diff
@@ -25,37 +25,57 @@
...
+    /// <summary>
+    /// Factory for creating ADAL or MSAL exceptions.
+    /// ErrorCodes should be made public constants for users to reference them
+    /// </summary>
     internal interface ICoreExceptionFactory
     {
+        /// <summary>
+        /// Create a client exception, arising from logic within this library.
+        /// </summary>
+        /// <returns></returns>
```
jennyf19, line 42: >/// <returns></returns> [](start = 8, length = 23)

remove? #Resolved

core/src/Exceptions/CoreErrorMessages.cs
bgavrilMS: It looks like the server normally sends back a json with error details. We parse the json and give a nice message. But if the server sends a back some other error (and standard HTTP response bodies are not json), then we would have failed with a serialization exception. Now we fail with OAuthNonJsonError. 

Happy to change the name - what do you suggest? #Closed

jennyf19: that makes sense. I think it's fine.

---
In reply to: [222783088](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1294#discussion_r222783088) [](ancestors = 222783088)

MarkZuber: UnparseableError? (that's not awesome either but might spark ideas...) #Resolved

msal/tests/Test.MSAL.NET.Unit/RequestsTests/InteractiveRequestTests.cs
```diff
@@ -235,6 +235,98 @@ public void RedirectUriContainsFragmentErrorTest()
...
+            };
+            parameters.RedirectUri = new Uri("some://uri");
+
+            MockWebUI ui = new MockWebUI();
+
+            InteractiveRequest request = new InteractiveRequest(parameters,
+                TestConstants.ScopeForAnotherResource.ToArray(),
+                 TestConstants.DisplayableId,
+                UIBehavior.SelectAccount, ui);
+
```
MarkZuber, line 266: nit: remove extra blank line #Resolved

msal/tests/Test.MSAL.NET.Unit/RequestsTests/InteractiveRequestTests.cs
```diff
@@ -235,6 +235,98 @@ public void RedirectUriContainsFragmentErrorTest()
...
+            catch (Exception exc)
+            {
+                MsalServiceException serverEx = exc.InnerException as MsalServiceException;
+                Assert.IsNotNull(serverEx);
+                Assert.AreEqual(429, serverEx.StatusCode);
+                Assert.AreEqual(MockHelpers.TooManyRequestsContent, serverEx.ResponseBody);
+                Assert.AreEqual(MockHelpers.TestRetryAfterDuration, serverEx.Headers.RetryAfter.Delta);
+                Assert.AreEqual(CoreErrorCodes.OAuthNonJsonError, serverEx.ErrorCode);
+            }
+        }
```
jennyf19, line 283: check that all the mocks were consumed? #Resolved

msal/tests/Test.MSAL.NET.Unit/RequestsTests/InteractiveRequestTests.cs
```diff
@@ -235,6 +235,98 @@ public void RedirectUriContainsFragmentErrorTest()
...
+            }
+            catch (Exception exc)
+            {
+                MsalServiceException serverEx = exc.InnerException as MsalServiceException;
+                Assert.IsNotNull(serverEx);
+                Assert.AreEqual(429, serverEx.StatusCode);
+                Assert.AreEqual(MockHelpers.TestRetryAfterDuration, serverEx.Headers.RetryAfter.Delta);
+                Assert.AreEqual("Server overload", serverEx.ErrorCode);
+            }
+        }
```
jennyf19, line 328: check for consumed mocks #Resolved

msal/tests/Test.MSAL.NET.Unit/RequestsTests/InteractiveRequestTests.cs
MarkZuber: comment for future:  should we make the mock http handler stuff a disposable and auto-fail a test if all mocks aren't consumed?  might help us avoid having to add that every time..

MarkZuber (approved):

core/src/Exceptions/CoreErrorMessages.cs
jennyf19: Good point...the reason we return this is because the Json is not parseable..."nonjson" isn't clear on that by just looking at it. Or add that to the error message, like you have in the summary for MsalError. "cannot be parsed" something like that.

---
In reply to: [222785065](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1294#discussion_r222785065) [](ancestors = 222785065)

jennyf19 (approved):
:shipit:

msal/tests/Test.MSAL.NET.Unit/RequestsTests/InteractiveRequestTests.cs
jennyf19: yes...definitely.

---
In reply to: [222787298](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1294#discussion_r222787298) [](ancestors = 222787298)

bgavrilMS:
Merged manually

PR #1297, opened 2018-10-04
#supportablity for TeamId issue

jennyf19:
Check for success error code before creating match

### DIFF
--- a/core/src/Exceptions/CoreErrorCodes.cs
+++ b/core/src/Exceptions/CoreErrorCodes.cs
@@ -77,7 +77,9 @@ namespace Microsoft.Identity.Core
         public const string UapCannotFindUpn = "uap_cannot_find_upn";
 
         public const string GetUserNameFailed = "get_user_name_failed";
-
+#if iOS
+        public const string GetTeamIdReturnedNull = "null_TeamId";
+#endif
 
 #if ANDROID
         public const string FailedToCreateSharedPreference = "shared_preference_creation_failed";
--- a/core/src/Exceptions/CoreErrorCodes.cs
+++ b/core/src/Exceptions/CoreErrorCodes.cs
@@ -78,7 +78,7 @@ namespace Microsoft.Identity.Core
 
         public const string GetUserNameFailed = "get_user_name_failed";
 #if iOS
-        public const string GetTeamIdReturnedNull = "null_TeamId";
+        public const string CannotAccessPublisherKeyChain = "cannot_access_publisher_keychain";
 #endif
 
 #if ANDROID
--- a/core/src/Exceptions/CoreErrorMessages.cs
+++ b/core/src/Exceptions/CoreErrorMessages.cs
@@ -86,12 +86,14 @@ namespace Microsoft.Identity.Core
 
         public const string InternalErrorCacheEmptyUsername =
             "Internal error - trying to remove an ADAL user with an empty username. Possible cache corruption. See https://aka.ms/adal_token_cache_serialization";
-        public const string InternalErrorCacheEmptyIdentifier = 
+        public const string InternalErrorCacheEmptyIdentifier =
             "Internal error - trying to remove an ADAL user with an empty identifier. Possible cache corruption. See https://aka.ms/adal_token_cache_serialization";
 
         public const string GetUserNameFailed = "Failed to get user name from the operating system.";
 
-        public const string GetTeamIdReturnedNull =
-           "TeamId returned null from iOS keychain. See https://aka.ms/msal-net-enable-keychain-access for more details on enabling keychain access.";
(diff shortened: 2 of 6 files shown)


core/src/Exceptions/CoreErrorCodes.cs
```diff
@@ -77,7 +77,9 @@ internal class CoreErrorCodes
         public const string UapCannotFindUpn = "uap_cannot_find_upn";
 
         public const string GetUserNameFailed = "get_user_name_failed";
-
+#if iOS
+        public const string GetTeamIdReturnedNull = "null_TeamId";
```
jennyf19: suggestions on a better name or value?  #Resolved

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -96,7 +96,14 @@ private string GetTeamId()
...
 
-            return match.AccessGroup.Split('.')[0];
+            if (resultCode == SecStatusCode.Success)
+            {
+                return match.AccessGroup.Split('.')[0];
+            }
+
+            throw CoreExceptionFactory.Instance.GetClientException(
+                CoreErrorCodes.GetTeamIdReturnedNull,
+                CoreErrorMessages.GetTeamIdReturnedNull);
```
jennyf19: we're assuming if the code != success, then the reason is the null teamId, but it could be something else. Do we want to do something else here in that case? #Resolved

core/src/Exceptions/CoreErrorCodes.cs
jmprieur: Wouldn't the customer facing issue that the app cannot access the KeyChain for all the apps of the same publisher? What about `CannotAccessPublisherKeyChain` ? #Resolved

core/src/Exceptions/CoreErrorMessages.cs
```diff
@@ -91,5 +91,7 @@ internal class CoreErrorMessages
 
         public const string GetUserNameFailed = "Failed to get user name from the operating system.";
 
+        public const string GetTeamIdReturnedNull =
+           "TeamId returned null from iOS keychain. See https://aka.ms/msal-net-enable-keychain-access for more details on enabling keychain access.";
```
jmprieur: What about:
Why do you get this error?
the application cannot access iOS keychain for application publisher (TeamId is null), whereas it's needed to enable Single Sign On between applications of the same publisher

How to remediate?
his iOS app configuration issue. See https://aka.ms/msal-net-enable-keychain-access for more details on enabling keychain access. #Resolved

msal/src/Microsoft.Identity.Client/Exceptions/MsalClientException.cs
```diff
@@ -129,6 +129,20 @@ public class MsalClientException : MsalException
         /// The library is loaded on a platform which is not supported.
         /// </summary>
         public const string PlatformNotSupported = "platform_not_supported";
+
+#if iOS
+        /// <summary>
+        /// Xamarin.iOS specific. This erorr indicates that keychain access has not be enabled for the application.
```
jmprieur: Typo: erorr => error, otherwise excellent. Not sure we even want to mention the TeamId.
what about
From MSAL 2.x and ADAL 4.x, the keychain for the publisher need to be accessed in order to provide Single Sign On between applications of the same publisher.

the rest is perfect 👍 

note that the <CodesignEntitlements>Entitlements.plist</CodesignEntitlements> sentence will probably generate a build warning, I believe you need to use &amp;lt; for < and &amp;gt; for > #Resolved

jmprieur (changes requested):
This looks good
I've suggested changes for the name of the error, and a bit for the XML comment.

MarkZuber (approved):
Functionally, I think this change is righteous.  Please resolve naming/commenting issues with JM before merging this.

core/src/Exceptions/CoreErrorCodes.cs
jennyf19: agreed. made the change and am also changing the response.

---
In reply to: [222515585](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1297#discussion_r222515585) [](ancestors = 222515585)

msal/src/Microsoft.Identity.Client/Exceptions/MsalClientException.cs
jennyf19: great...thank you!

---
In reply to: [222516169](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1297#discussion_r222516169) [](ancestors = 222516169)

jmprieur (approved):
LGTM. Thanks for addressing the PR comments.

PR #1304, opened 2018-10-04
integ testig + bug fixes

SomkaPe:
while writing integration tests next bugs were found and fixed:

1. Clear cache API - on Msal clear cache api did not clear persistent version of  Adal cache because "onAfter" delegate was not called (only in memory representation was cleared).
as result adal cache was not cleared.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/651
2. Same problem with opposite direction Adal -> Msal.
https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1344
3. RemoveAdalUser api should remove users by accountId and by displayableId ( to cover scenario of
migrarion form adalv3 to adalv4 to msal2)
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/652
4. NullPoinetEx in  RemoveMsalAccount (Msal) - when adalv3 account passed
5. NullPoinetEx in  RemoveAdalUser (Msal) - when adalv3 account passed
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/653
6. TokenCacheSerializeHelper.DeserializeUnifiedCache   did not clear in memory version of cache so it was appending it 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/654


jennyf19:
@SomkaPe Can you tell us which bugs you fixed and where? will make the review easier. We've also fixed a few more in the past couple of weeks. Thanks :)

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
```diff
@@ -47,6 +47,7 @@ public class DeviceCodeIntegrationTests
 
         [TestMethod]
         [TestCategory("UsernamePasswordIntegrationTests")]
+        [Ignore]
```
bgavrilMS: I don't think this is needed, they should pass on the build machine as the user running the tests has access to KeyVault. @jennyf19 - the tests should fail more gracefully, i.e. they fail with a NullRef today, they should fail with a nice message  #Closed

core/src/Cache/CacheFallbackOperations.cs
```diff
@@ -214,10 +214,8 @@ internal class CacheFallbackOperations
                 {
                     RemoveEntriesWithMatchingId(environmentAliases, clientId, accountOrUserId, adalCache);
                 }
-                else
-                {
-                    RemoveEntriesWithMatchingName(environmentAliases, clientId, displayableId, adalCache);
-                }
+
+                RemoveEntriesWithMatchingName(environmentAliases, clientId, displayableId, adalCache);
```
bgavrilMS, line 218: Could you explain, and also add a comment in the code - how will this help the ADAL v3 -> ADAL v4 -> MSAL upgrade? 

Also, don't you think we will be deleting too much in the standard case? The RemoveEntriesWithMatchingName will delete entries with the same name even if they are from different accounts ...

 #Closed

core/src/Cache/TokenCacheSerializeHelper.cs
```diff
@@ -59,6 +59,8 @@ internal static void DeserializeUnifiedCache(TokenCacheAccessor tokenCacheAccess
                 return;
             }
 
+            tokenCacheAccessor.Clear();
```
bgavrilMS: Shouldn't this go to before if (cacheDict is empty) ?  #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/InternalsVisibleTo.cs
```diff
@@ -34,3 +34,4 @@
...
 [assembly: InternalsVisibleTo("Test.ADAL.NET.Integration, PublicKey=0024000004800000940000000602000000240000525341310004000001000100B5FC90E7027F67871E773A8FDE8938C81DD402BA65B9201D60593E96C492651E889CC13F1415EBB53FAC1131AE0BD333C5EE6021672D9718EA31A8AEBD0DA0072F25D87DBA6FC90FFD598ED4DA35E44C398C454307E8E33B8426143DAEC9F596836F97C8F74750E5975C64E2189F45DEF46B2A2B1247ADC3652BF5C308055DA9")]
 [assembly: InternalsVisibleTo("WinFormsAutomationApp, PublicKey=0024000004800000940000000602000000240000525341310004000001000100B5FC90E7027F67871E773A8FDE8938C81DD402BA65B9201D60593E96C492651E889CC13F1415EBB53FAC1131AE0BD333C5EE6021672D9718EA31A8AEBD0DA0072F25D87DBA6FC90FFD598ED4DA35E44C398C454307E8E33B8426143DAEC9F596836F97C8F74750E5975C64E2189F45DEF46B2A2B1247ADC3652BF5C308055DA9")]
+[assembly: InternalsVisibleTo("Test.MSAL.NET.Integration, PublicKey=00240000048000009400000006020000002400005253413100040000010001002D96616729B54F6D013D71559A017F50AA4861487226C523959D1579B93F3FDF71C08B980FD3130062B03D3DE115C4B84E7AC46AEF5E192A40E7457D5F3A08F66CEAB71143807F2C3CB0DA5E23B38F0559769978406F6E5D30CEADD7985FC73A5A609A8B74A1DF0A29399074A003A226C943D480FEC96DBEC7106A87896539AD")]
```
bgavrilMS, line 37: These are for the L1 tests, i.e. E2E headless. If you need internal relation, these are L0, and they should go to the unit test assembly. #Closed

msal/tests/Test.MSAL.NET.Integration/Test.MSAL.NET.Integration.csproj
```diff
@@ -4,6 +4,9 @@
     <TargetFramework>net4.5.2</TargetFramework>
 
     <IsPackable>false</IsPackable>
+
+    <SignAssembly>true</SignAssembly>
```
bgavrilMS, line 8: Good idea, we should sign all our test assemblies instead of applying the sn hack #Closed

msal/tests/Test.MSAL.NET.Integration/Test.MSAL.NET.Integration.csproj
```diff
@@ -19,8 +22,13 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\..\adal\src\Microsoft.IdentityModel.Clients.ActiveDirectory\Microsoft.IdentityModel.Clients.ActiveDirectory.csproj">
```
bgavrilMS, line 33: Please add a comment on why this is needed, i.e. qualified name collisions... #Closed

msal/tests/Test.MSAL.NET.Unit/AuthorityAliasesTests.cs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        internal void AddMockResponseForInstanceDisovery()
+        {
+            HttpMessageHandlerFactory.AddMockHandler(
+                MockHelpers.CreateInstanceDiscoveryMockHandler(
+                    TestConstants.GetDiscoveryEndpoint(TestConstants.AuthorityCommonTenant)));
+        }
+
+        class TestLegacyCachePersistance : ILegacyCachePersistance
```
bgavrilMS: This class also exists in UnifiedCacheTests - could you pull it out into the Mocks directory and use it from there? Alternatively, use NSubstitute to create on the fly. #Closed

msal/tests/Test.MSAL.NET.Unit/AuthorityAliasesTests.cs
```diff
@@ -0,0 +1,216 @@
...
+                try
+                {
+                    result = app.AcquireTokenSilentAsync(TestConstants.ScopeForAnotherResource,
+                        app.GetAccountsAsync().Result.First(),
+                        string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/", envAlias, TestConstants.Utid),
+                        false).Result;
+                }
+                catch (AggregateException ex)
+                {
+                    Assert.IsNotNull(ex.InnerException);
```
bgavrilMS: The intent of the test is not clear - do we expect a UIRequiredException or do we expect the result to be retured? Please refactor so that the requests that result in an exception fail if a result is returned and vice versa. #Closed

core/src/Cache/CacheFallbackOperations.cs
bgavrilMS, line 218: There are unit tests for this logic, if changes are made, there should be unit test changes ( a bit surprised if they do no fail) #Closed

bgavrilMS (changes requested):
:clock1:

msal/tests/Test.MSAL.NET.Integration/UsernamePasswordIntegrationTests.cs
```diff
@@ -24,9 +24,10 @@
 // THE SOFTWARE.
 //
 //------------------------------------------------------------------------------
+extern alias msal;
 
 using Test.Microsoft.Identity.LabInfrastructure;
-using Microsoft.Identity.Client;
+using msal::Microsoft.Identity.Client;
```
MarkZuber, line 30: can you please explain why this msal:: alias is now needed?   #Closed

core/src/Cache/TokenCacheSerializeHelper.cs
SomkaPe: oh, right - will fix it  #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/InternalsVisibleTo.cs
SomkaPe: not sure what you mean here, please explain  #Closed

msal/tests/Test.MSAL.NET.Integration/Test.MSAL.NET.Integration.csproj
SomkaPe: to cover unified cache test cases we need to be able create instances of Adal  and Msal in same test #Closed

msal/tests/Test.MSAL.NET.Unit/AuthorityAliasesTests.cs
SomkaPe: ok, will change #Closed

msal/tests/Test.MSAL.NET.Unit/AuthorityAliasesTests.cs
SomkaPe: good catch, will change #Closed

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: This is covered by UnifiedCache_Adalv3ToAdalV4ToMsal2MigrationIntegrationTest #Closed

msal/tests/Test.MSAL.NET.Integration/UsernamePasswordIntegrationTests.cs
SomkaPe: because of namespace collisions between adal and msal, I have added integ tests for unified cache which create msal and adal client in the same test #Closed

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
SomkaPe: this test more some kind of manual test since the step of authentication is not automated, so it is not integration test #Closed

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: this case is covered in UnifiedCache_Adalv3ToAdalV4ToMsal2MigrationIntegrationTest : 
if you get AT for res1 in  adal v3  than migrate to adal v4 and get token to res2 you will have next state of the cache:
for the same Account you have one entity in adal cache in adalv3 format (no client info)
another entity in adal v4 format (with client info).
So you have two Adal cache entities for the same account .

If you migrate to Msal and call get accounts api one Account object will be returned.
If you call Remove account it will remove only V4 adal entity , so next call to get accounts will return same account object (this time with id null)

So remove account api will not remove account .
Unless remove account api remove by displayble id also.

 #Closed

core/src/Cache/TokenCacheSerializeHelper.cs
SomkaPe: Good Catch #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/InternalsVisibleTo.cs
henrik-me: The basic idea is that integration tests should have no reason to have access to internals. If you need access to internals then perhaps the test is a unit test and not an integration test.

---
In reply to: [223127605](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r223127605) [](ancestors = 223127605)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -294,22 +294,24 @@ public virtual void Clear()
         {
             lock (cacheLock)
             {
+                TokenCacheNotificationArgs args = new TokenCacheNotificationArgs { TokenCache = this };
+                this.OnBeforeAccess(args);
```
henrik-me, line 300: is it needed to have the lock on the cache for the onbeforeaccess?

Alos the cache is also locked for every call to the eventlisteners. #Closed

core/src/Cache/CacheFallbackOperations.cs
henrik-me: We have been going back and forth on this. Last time we ended up saying we should not do both deletes as it have side effects. 

---
In reply to: [223143317](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r223143317) [](ancestors = 223143317)

core/src/Cache/TokenCacheSerializeHelper.cs
```diff
@@ -50,6 +50,8 @@ internal class TokenCacheSerializeHelper
         /// <param name="requestContext">call state to pass correlation id and logger instance</param>
         internal static void DeserializeUnifiedCache(TokenCacheAccessor tokenCacheAccessor, byte[] unifiedState, RequestContext requestContext)
         {
+            tokenCacheAccessor.Clear();
```
henrik-me, line 48: >tokenCacheAccessor [](start = 12, length = 18)

This seems important (an alternative is to throw if the tokenAccessor is not empty, or to simply create the tokenCacheAccessor in the helper method).... are unit tests covering this helper's behavior? #Pending

core/tests/Test.Microsoft.Identity.Core.Unit/Mocks/MockHelpers.cs
```diff
@@ -315,5 +316,19 @@ public static HttpMessageHandler CreateInstanceDiscoveryMockHandler(string url,
...
+            public byte[] LoadCache()
+            {
+                return data;
+            }
+
+            public void WriteCache(byte[] serializedCache)
+            {
+                data = serializedCache;
+            }
+        }
```
henrik-me: new file #Closed

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -901,6 +901,11 @@ internal async Task RemoveAsync(string authority, bool validateAuthority, IAccou
 
         internal void RemoveMsalAccount(IAccount account, ISet<string> environmentAliases, RequestContext requestContext)
         {
+            if (account.HomeAccountId == null)
```
henrik-me, line 917: >if (account.HomeAccountId == null) [](start = 12, length = 34)

is this properly covered by utests? 
Also should probably at least be logged as info? #Closed

msal/src/Microsoft.Identity.Client/Properties/InternalsVisibleTo.cs
```diff
@@ -10,4 +10,5 @@
...
\ No newline at end of file
+[assembly: InternalsVisibleTo("Test.MSAL.NET.Unit,       PublicKey=00240000048000009400000006020000002400005253413100040000010001002D96616729B54F6D013D71559A017F50AA4861487226C523959D1579B93F3FDF71C08B980FD3130062B03D3DE115C4B84E7AC46AEF5E192A40E7457D5F3A08F66CEAB71143807F2C3CB0DA5E23B38F0559769978406F6E5D30CEADD7985FC73A5A609A8B74A1DF0A29399074A003A226C943D480FEC96DBEC7106A87896539AD")]
+[assembly: InternalsVisibleTo("Test.MSAL.NET.Integration, PublicKey=00240000048000009400000006020000002400005253413100040000010001002D96616729B54F6D013D71559A017F50AA4861487226C523959D1579B93F3FDF71C08B980FD3130062B03D3DE115C4B84E7AC46AEF5E192A40E7457D5F3A08F66CEAB71143807F2C3CB0DA5E23B38F0559769978406F6E5D30CEADD7985FC73A5A609A8B74A1DF0A29399074A003A226C943D480FEC96DBEC7106A87896539AD")]
```
henrik-me, line 14: >InternalsVisibleTo("Test.MSAL.NET.Integration [](start = 11, length = 45)

see related comment. #Pending

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
```diff
@@ -25,8 +25,10 @@
 //
 //------------------------------------------------------------------------------
 
+extern alias msal;
```
henrik-me: not needed. you can specify alias'es as art of the using statement #Closed

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
henrik-me: pls. do not ignore tests. if it doesn't pass locally due to missing setup investigate how to best fix that.

---
In reply to: [222927993](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r222927993) [](ancestors = 222927993)

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+extern alias msal;
+extern alias adal;
```
henrik-me, line 29: add these alias's to the using statements instead. #Closed

msal/tests/Test.MSAL.NET.Unit/TestConstants.cs
```diff
@@ -42,14 +42,20 @@ class TestConstants
         public static readonly SortedSet<string> ScopeForAnotherResource = new SortedSet<string>(new[] { "r2/scope1", "r2/scope2" });
         public static readonly SortedSet<string> CacheMissScope = new SortedSet<string>(new[] { "r3/scope1", "r3/scope2" });
         public static readonly string ScopeForAnotherResourceStr = "r2/scope1 r2/scope2";
-        public static readonly string ProductionPrefNetworkEnvironment = "login.microsoftonline.com";
-        public static readonly string ProductionPrefCacheEnvironment = "login.windows.net";
+        public static readonly string Uid = "my-uid";
+        public static readonly string Utid = "my-utid";
+
+        public static readonly string ProdPrefNetworkEnv = "login.microsoftonline.com";
```
henrik-me: not sure I understand the need for shortening the consts. Please change it to meaningful names. reading the test code I found it hard to understand these abbreviations #Pending

msal/tests/Test.MSAL.NET.Unit/AuthorityAliasesTests.cs
```diff
@@ -0,0 +1,202 @@
...
+            {
+                Assert.AreEqual(expectedEnvironment, account.Environment);
+            }
+
+            IDictionary<AdalTokenCacheKey, AdalResultWrapper> adalCache =
+                AdalCacheOperations.Deserialize(cache.legacyCachePersistance.LoadCache());
+
+            foreach (KeyValuePair<AdalTokenCacheKey, AdalResultWrapper> kvp in adalCache)
+            {
+                Assert.AreEqual(expectedEnvironment, new Uri(kvp.Key.Authority).Host);
```
henrik-me, line 195: >new Uri(kvp.Key.Authority).Host [](start = 53, length = 31)

nit: would be great if it was not required to new up an object. #Closed

msal/tests/Test.MSAL.NET.Unit/AuthorityAliasesTests.cs
```diff
@@ -0,0 +1,202 @@
...
+                Assert.AreEqual(expectedEnvironment, rt.Environment);
+            }
+
+            var idTokens = cache.GetAllIdTokensForClient(new RequestContext(new MsalLogger(Guid.NewGuid(), null)));
+            foreach (var id in idTokens)
+            {
+                Assert.AreEqual(expectedEnvironment, id.Environment);
+            }
+
+            var accounts = cache.GetAllAccounts(new RequestContext(new MsalLogger(Guid.NewGuid(), null)));
```
henrik-me: >new MsalLogger(Guid.NewGuid(), null) [](start = 67, length = 36)

can you use the same logger for all? What about the same request context? #Pending

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
```diff
@@ -45,6 +47,7 @@ public class DeviceCodeIntegrationTests
 
         [TestMethod]
         [TestCategory("UsernamePasswordIntegrationTests")]
+        [Ignore]
         public async Task AcquireTokenWithManagedUsernamePasswordAsync()
```
henrik-me: >AcquireTokenWithManagedUsernamePasswordAsync [](start = 26, length = 44)

@jennyf19: Not sure why this test is in a test class for DeviceCodeIntegrationTests, thoughts? #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+using System.Threading.Tasks;
+using Test.MSAL.NET.Integration.Infrastructure;
+using System.IO;
+using System.Linq;
+using System;
+
+namespace Test.MSAL.NET.Integration
+{
+    [TestClass]
+    public class UnifiedCacheTests
```
henrik-me, line 45: >UnifiedCacheTests [](start = 17, length = 17)

These tests are not able to run in parallel. I'm thinking it should be possible to structure the tests in a way which doesn't prevent them from running in parallel. #Closed

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
jennyf19: Not sure. The actual u/p integration tests are in a [separate class](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/msal/tests/Test.MSAL.NET.Integration/UsernamePasswordIntegrationTests.cs#L39) - I think the method name and TestCategory name did not get changed on a copy/paste. just a guess. want me to fix it? 

---
In reply to: [223430145](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r223430145) [](ancestors = 223430145)

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+                    IsMfaUser = false,
+                    IsFederatedUser = false
+                });
+
+            new Test.Microsoft.Identity.LabInfrastructure.UserQueryParameters
+            {
+                IsMamUser = false,
+                IsMfaUser = false,
+                IsFederatedUser = false
+            };
```
henrik-me: what is this doing? #Pending

henrik-me (changes requested):
:clock1:

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
SomkaPe: it does not work because authentication is not implemented, so it is not automated test #Closed

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: not sure what side effects you mean here , In extremely rare case when few different accounts with same displaybleId/preferred_user_name were used in Adal, Msal remove Account api will remove SSO in adal cache format for both of them which is less evil that not working  removeAccount api for described scenario(which is quite possible scenario ) . If you want me to remove this change I can do it.  #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/InternalsVisibleTo.cs
SomkaPe: how would you check state of the cache or modify it (for example to expire AT to trigger RT usage)? #Closed

core/src/Cache/TokenCacheSerializeHelper.cs
SomkaPe: Why would you throw here ? #Closed

core/tests/Test.Microsoft.Identity.Core.Unit/Mocks/MockHelpers.cs
SomkaPe: ? #Closed

msal/src/Microsoft.Identity.Client/TokenCache.cs
SomkaPe: Adal v3 to Msal v2 migration is not covered by unit test, this integration project is only one which has references to adal an msal so this scenario can be tested  #Closed

msal/tests/Test.MSAL.NET.Unit/TestConstants.cs
SomkaPe: ok, will change back #Closed

msal/tests/Test.MSAL.NET.Unit/AuthorityAliasesTests.cs
SomkaPe: i need to extract host form url string here #Closed

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
SomkaPe: ) #Closed

SomkaPe: copy paste == evil #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: will remove #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
SomkaPe: probably no , in general onBeforeAccess call is not necessary here , we do not need to read it here.
But It is general pattern of accessing cache here (in this lib)  #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: ok, i will make in memory persisted cache instance specific so individual tests will be isolated.
 #Pending

core/src/Cache/CacheFallbackOperations.cs
bgavrilMS: I understand the scenario now, thank you for explaining Petro. Definitely needs to be in the code as a comment if we decided to go for it. Is this algorithm discussed with the other teams or is it only us that have this problem? #Closed

core/tests/Test.Microsoft.Identity.Core.Unit/Mocks/MockHelpers.cs
bgavrilMS: Please place this class in a new file. See other objects such as the TestLogger.cs #Closed

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
bgavrilMS: UsernamePassword is implemented and shipped on MSAL since 2.1.0. We even shipped Device Flow in 2.2.0 :). This test passes on the build machine, but fails on our machines. If you just delete the [Ignore] I am sure it will not cause issues on the build machine. We will look separately into making it pass or at least fail more gracefully on dev machines. #Closed

msal/tests/Test.MSAL.NET.Integration/Test.MSAL.NET.Integration.csproj
bgavrilMS: Sorry for not making myself clear. When I say "please add a comment", I mean "please add a comment in the code" not "please add a comment in the PR"  #Closed

msal/tests/Test.MSAL.NET.Integration/UsernamePasswordIntegrationTests.cs
bgavrilMS: Yeah, I have seen this as well. DeviceCodeResult now exists in 2 separate assemblies (ADAL and MSAL) and the current assemblies references both. As such it is a separate type, even though the fully qualified name (namespace + type name) is the same. So .net needs a way to disambiguate between the 2. 

I have a work in progress to get rid of "core" unit tests and just have the core tests run as part of ADAL and as part of MSAL. This will eliminate the need for assembly aliasing. Until then, it's probably the only way. #Closed

msal/tests/Test.MSAL.NET.Unit/AuthorityAliasesTests.cs
bgavrilMS: +1;  #Pending

bgavrilMS: On top of this, we should check that no errors are logged. This is quite important because a lot of the token cache logic does a try / catch / ignore exception and log, so it's the only way to tell if an exception was thrown... #Pending

core/src/Cache/TokenCacheSerializeHelper.cs
henrik-me: I'm suggesting 2 ways to deal with this differently, to identify the better of these options or potentially a third I would need to understand why you had to add this in the first place.

1) To me it's magic that we decide to clear the object, and if we do there is really no need to pass it in. We can just return a new accessor instead. 
2) Alternatively if it's very important that the object has been cleared in advance then we can throw an exception instead (should the object not be in the expected state coming in)

How did you discover that .clear was needed?
 #Pending

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
SomkaPe: it can not pass anywhere it is not even test , it makes no sense, it just some random copy-paste(test category, test name is not related to  device code flow, user object is not used , empty call back is passed)
It just wait 15 mins and than fails (because deviceCode result is just ignored - empty callback passed).

If you think it pass on build machine It just means that it is not get run there.

 #Closed

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: Discussed with Olga, she implemented something similar #Closed

core/src/Cache/TokenCacheSerializeHelper.cs
SomkaPe: For custom cache persistence scenarios,  every cache access preceded by before access delegate call,  which is the place were cache supposed to be read from some persistent storage and deserialized to in-memory representation.
This is important to reflect current state of the cache, persistent state of cache  is "golden source" of it , not in-memory version.
For example : 
1.  you share cache between few app
2.  in  app1 you create msalClient1, then acquire token (now you have tokens in the cache).
3.  switch to app2 create msalClient2 then clear cache
4. return to app1  and execute getAccount on  msalClient1 - during deserialization   in memory representation of the cache should be cleared and instead persistent state (which is empty now) should be used,
  #Closed

core/tests/Test.Microsoft.Identity.Core.Unit/Mocks/MockHelpers.cs
SomkaPe: ok #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/InternalsVisibleTo.cs
bgavrilMS: You don't. Unit tests do that.  #Closed

core/src/Cache/CacheFallbackOperations.cs
bgavrilMS: Sounds good to me then. Note that there are more granular unit tests in LCacheFallbackOperationsTests.cs - we should add smth for this case  #Closed

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
bgavrilMS: If I understood correctly, this is a copy paste error. Please delete the test. 

In general we should not [Ignore] tests because we forget about them and after a few months they become dead code. We should only Ignore if we create a bug / work item to track the work of un-ignoring.  #Closed

henrik-me: @MarkZuber : What was the intentions of this? Should it be deleted? How is the e2e of Device Codeflow validated now? Manually (does the test exist in VSTS)?

---
In reply to: [224132337](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r224132337) [](ancestors = 224132337)

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
henrik-me: any update on this change? (make all tests instance based and not depending on state captured in the test class.... there is one minor exception to that however I believe it should be possible to write these tests with-out that).

---
In reply to: [223480524](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r223480524) [](ancestors = 223480524)

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+        {
+            ClearCacheStorage();
+
+            user = authHelper.GetUser(
+                new Test.Microsoft.Identity.LabInfrastructure.UserQueryParameters
+                {
+                    IsMamUser = false,
+                    IsMfaUser = false,
+                    IsFederatedUser = false
+                });
```
henrik-me, line 72: @trwalke has a consolidation of std. users coming up with his change.   #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+            Assert.AreEqual(user.Upn, msalAuthResult.Account.Username);
+        }
+
+        private void ValidateAdalAuthResult()
+        {
+            Assert.IsNotNull(adalAuthResult);
+            Assert.IsNotNull(adalAuthResult.AccessToken);
+            Assert.IsNotNull(adalAuthResult.IdToken);
+            Assert.AreEqual(user.Upn, adalAuthResult.UserInfo.DisplayableId);
+        }
```
henrik-me, line 178: should specific values be validated not just that something exists? #Closed

msal/tests/Test.MSAL.NET.Integration/Test.MSAL.NET.Integration.csproj
henrik-me: can this be applied to all test and sample projects?

---
In reply to: [222932902](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r222932902) [](ancestors = 222932902)

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+
+        [TestMethod]
+        [TestCategory("UnifiedCache_IntegrationTests")]
+        public async Task UnifiedCache_Msal2_Adalv4_SharedRtGetUsed()
```
henrik-me: >SharedRtGetUsed [](start = 52, length = 15)

suggestion: 
AdalCache_CreatedBy_MSAL_Is_UsedBy_ADALandMSAL_When_MSALCache_Is_Empty #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+
+            ClearMsalCache();
+            AssertMsalCacheIsEmpty();
+
+            // passing empty password to make sure that token returned silenlty - using RT
+            adalAuthResult = await adal::Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationContextIntegratedAuthExtensions.
+                AcquireTokenAsync(adalContext, AdalResource1, ClientId, 
+                new adal::Microsoft.IdentityModel.Clients.ActiveDirectory.UserPasswordCredential(user.Upn, ""));
+
+            ValidateAdalAuthResult();
```
henrik-me, line 194: suggestion: Clear MSAL cache again and call msal.aquiretoken to ensure msal also uses the adal cache, this way this test will validate that both libraries can use items from just the adal cache created with MSAL #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+        }
+
+        [TestMethod]
+        [TestCategory("UnifiedCache_IntegrationTests")]
+        public async Task UnifiedCache_Msal2_Adalv4_MsalRtUsedByAdal()
```
henrik-me: >UnifiedCache_Msal2_Adalv4_MsalRtUsedByAdal [](start = 26, length = 42)

See above... naming change is both uses the MSAL cache... #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+            adalAuthResult = await adal::Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationContextIntegratedAuthExtensions.
+                AcquireTokenAsync(adalContext, AdalResource2, ClientId,
+                new adal::Microsoft.IdentityModel.Clients.ActiveDirectory.UserPasswordCredential(user.Upn, securePassword));
+            ValidateAdalAuthResult();
+
+            ClearAdalCache();
+            AssertAdalCacheIsEmpty();
+
+            var msalAccounts = await msalPublicClient.GetAccountsAsync();
+            Assert.AreEqual(1, msalAccounts.Count());
```
henrik-me, line 281: should initially assert that there are no accounts? #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+
+            msalAuthResult = await msalPublicClient.AcquireTokenSilentAsync(MsalScopes, msalAccounts.First());
+            ValidateMsalAuthResult();
```
henrik-me, line 284: I would suggest using the same with ADAL, clearing the ADAL cache first? #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+                AcquireTokenAsync(adalContext, AdalResource1, ClientId,
+                new adal::Microsoft.IdentityModel.Clients.ActiveDirectory.UserPasswordCredential(user.Upn, securePassword));
+            ValidateAdalAuthResult();
+
+            adalAuthResult = await adal::Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationContextIntegratedAuthExtensions.
+                AcquireTokenAsync(adalContext, AdalResource2, ClientId,
+                new adal::Microsoft.IdentityModel.Clients.ActiveDirectory.UserPasswordCredential(user.Upn, securePassword));
+            ValidateAdalAuthResult();
+            
+            // simulate adalV3 token cache state by setting client info in adal cache entities to null
```
henrik-me, line 402: >simulate adalV3 token cache state by setting client info in adal cache entities to null  [](start = 15, length = 88)

nice workaround :) #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+            // and clearing msal cache
+            UpdateAdalCacheSetClientInfoToNull();
+            ClearMsalCache();
+            AssertMsalCacheIsEmpty();
+
+            // Migration to AdalV4 - acquire adal tokens using adalV4
+            adalAuthResult = await adal::Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationContextIntegratedAuthExtensions.
+                AcquireTokenAsync(adalContext, AdalResource2, ClientId,
+                new adal::Microsoft.IdentityModel.Clients.ActiveDirectory.UserPasswordCredential(user.Upn, securePassword));
+            ValidateAdalAuthResult();
```
henrik-me, line 459: suggestion before this add a silent call using adal v4 to the resource you already requested an AT for... that way you will be able to validate that v4 works even with-out user_info... or perhaps that it doesn't work (expected?) #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+
+            var msalAccounts = await msalPublicClient.GetAccountsAsync();
+            Assert.AreEqual(1, msalAccounts.Count());
+            var account = msalAccounts.First();
+            Assert.AreEqual(user.Upn, account.Username);
+            // make sure for the same account RT in V4 format preffered over V3 format
+            Assert.IsNotNull(account.HomeAccountId);
+            Assert.IsNotNull(account.Environment);
+
+            // make sure that adal v4 RT is usable by Msal
```
henrik-me, line 478: >// make sure that adal v4 RT is usable by Msal [](start = 12, length = 46)

Not sure it is needed as you have already validted this in other tests, however would it make sense to delete the msal cache first here as well?  #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+            msalAuthResult = await msalPublicClient.AcquireTokenSilentAsync(MsalScopes, account);
+            ValidateMsalAuthResult();
+
+            // make sure Msal remove account api remove corresponding cache entities in all formats  
+            msalAccounts = await msalPublicClient.GetAccountsAsync();
+            Assert.AreEqual(1, msalAccounts.Count());
+            account = msalAccounts.First();
```
henrik-me, line 485: consider validating the number of cache items as well in each cache #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,491 @@
...
+            AssertMsalCacheIsEmpty();
+
+            // Migration to AdalV4 - acquire adal tokens using adalV4
+            adalAuthResult = await adal::Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationContextIntegratedAuthExtensions.
+                AcquireTokenAsync(adalContext, AdalResource2, ClientId,
+                new adal::Microsoft.IdentityModel.Clients.ActiveDirectory.UserPasswordCredential(user.Upn, securePassword));
+            ValidateAdalAuthResult();
+
+            // At this poing Adal cache contains RTs for the same account in diff format v3 and v4
+            Assert.IsTrue(adalCache.ReadItems().Count() == 2);
```
henrik-me, line 462: >Assert.IsTrue(adalCache.ReadItems().Count() == 2); [](start = 12, length = 50)

perhaps validate msal cache as well? (I assume one entry will be there) #Closed

msal/tests/Test.MSAL.NET.Integration/Test.MSAL.NET.Integration.csproj
henrik-me: I'm still confused as to why this is all needed. Is it because you can' t access core elements? ... is there a trick for this in the using statements, perhaps not but was hoping. 

---
In reply to: [223487288](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r223487288) [](ancestors = 223487288)

henrik-me (changes requested):
:clock1:

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/InternalsVisibleTo.cs
SomkaPe: This is integration testing of msal and adal coexistence  it is not unit test #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: we(adal/msal) do not validate tokens received from sts #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: ok #Pending

msal/tests/Test.MSAL.NET.Integration/DeviceCodeIntegrationTests.cs
MarkZuber: This was put in early in development of devicecode and wasn't fully fleshed out because we can't automate the browser side of things.  so i'll kill this test for now.  yes, Henrik, for now this test is manual since we don't automate the browser.

leave it as ignore in your PR (it's ignore on my end too) and i'll kill it with my refactor PR. #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: Good point , I have missed that case, will add and rename test names to better reflect test purpose #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: good point will add #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: it i actually validated in next lines by calling getAccounts api  #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: ok #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: i think no  - it is covered in prev tests #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,541 @@
...
+            ValidateAdalAuthResult();
+        }
+
+        [TestMethod]
+        [TestCategory("UnifiedCache_IntegrationTests")]
+        public async Task MsalRt_CreatedByMsalV2_UsedByMsalV2()
+        {
+            await AcquireTokensUsingMsal();
+
+            ClearAdalCache();
```
trwalke, line 236: Why clear the ADAL cache here? #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,541 @@
...
+            ValidateMsalAuthResult();
+        }
+
+        [TestMethod]
+        [TestCategory("UnifiedCache_IntegrationTests")]
+        public async Task AdalRt_CreatedByAdalV4_UsedByAdalV4()
+        {
+            await AcquireTokensUsingAdal();
+
+            ClearMsalCache();
```
trwalke, line 326: Why clear MSAL cache in ADAL cache test? #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,541 @@
...
+
+            var msalAccounts = await msalPublicClient.GetAccountsAsync();
+            Assert.AreEqual(1, msalAccounts.Count());
+            var account = msalAccounts.First();
+            Assert.AreEqual(user.Upn, account.Username);
+            // make sure for the same account RT in V4 format preffered over V3 format
+            Assert.IsNotNull(account.HomeAccountId);
+            Assert.IsNotNull(account.Environment);
+
+            // validate than Adal writes only Rt in Msal format  an
```
henrik-me: >format  an [](start = 57, length = 10)

this text update doesn't seem right... something is missing #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,541 @@
...
+        AuthHelper authHelper = new AuthHelper();
+
+        private const string AdalResource1 = "https://graph.windows.net";
+        private const string AdalResource2 = "https://graph.microsoft.com";
+
+        private Test.Microsoft.Identity.LabInfrastructure.IUser user;
+        private SecureString securePassword;
+
+        private byte[] AdalV3StateStorage;
+        private byte[] UnifiedStateStorage;
```
henrik-me, line 60: you will need to also have these ones in each test to ensure they can run completely independent of each other. for the token persistance handling, you can likely use delegates in each test.  #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,541 @@
...
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            user = authHelper.GetUser(
```
henrik-me, line 66: nit: you can probably do this only if null, the first test that runs will set it. 
@trwalker, have you added std. users to the test helper? #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
```diff
@@ -0,0 +1,541 @@
...
+                new Test.Microsoft.Identity.LabInfrastructure.UserQueryParameters
+                {
+                    IsMamUser = false,
+                    IsMfaUser = false,
+                    IsFederatedUser = false
+                });
+
+            securePassword = new NetworkCredential("", ((LabUser)user).GetPassword()).SecurePassword;
```
henrik-me: nit: likely only need to set this once, thus if already set it should be good. #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
henrik-me: true, but you are testing the entire result, I'm thinking that at least one value besides from the upn an be validated. I'm not asking you to look into the tokens.

---
In reply to: [224197546](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r224197546) [](ancestors = 224197546)

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: for every test method new instance of test class is instantiated, so it is not shared #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: ok, will do it in static constructor #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: ok, will do it in static constructor #Pending

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: fixed #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: because adal v4 and msal v2 now share sso - so can read and write each other  format #Closed

core/src/Cache/CacheFallbackOperations.cs
henrik-me: Did you cover this with tests Petro? Is that the test you referred to in another comment?

---
In reply to: [224130526](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r224130526) [](ancestors = 224130526)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/InternalsVisibleTo.cs
henrik-me: which is exactly why there should ideally not be a reference... but makes sense.

---
In reply to: [224193062](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r224193062) [](ancestors = 224193062)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
henrik-me: would have liked to see if we could loosen this a bit and have fewer locks

---
In reply to: [223477053](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r223477053) [](ancestors = 223477053)

msal/tests/Test.MSAL.NET.Unit/TestConstants.cs
henrik-me: Seems like these are still not changed.

---
In reply to: [223459786](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r223459786) [](ancestors = 223459786)

msal/tests/Test.MSAL.NET.Unit/AuthorityAliasesTests.cs
henrik-me: @SomskaPe ?

---
In reply to: [223489018](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r223489018) [](ancestors = 223489018)

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
henrik-me: Seems still to be around...

---
In reply to: [223460925](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r223460925) [](ancestors = 223460925)

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
henrik-me: Not sure about that... should still be a fairly easy change.

---
In reply to: [224570927](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r224570927) [](ancestors = 224570927)

msal/tests/Test.MSAL.NET.Unit/AuthorityAliasesTests.cs
bgavrilMS: You can setup a logger stub using NSubstitute and if anything is logged via Error or ErrorPii, fail the test. We do smth similar here [https://github.com:443/AzureAD/azure-activedirectory-library-for-dotnet.git/blob/dev/core/tests/Test.Microsoft.Identity.Core.Unit/CacheTests/CacheFallbackOperationsTests.cs#L188](https://github.com:443/AzureAD/azure-activedirectory-library-for-dotnet.git/blob/dev/core/tests/Test.Microsoft.Identity.Core.Unit/CacheTests/CacheFallbackOperationsTests.cs#L188)

---
In reply to: [224676750](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r224676750) [](ancestors = 224676750,223489018)

henrik-me: [https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/core/tests/Test.Microsoft.Identity.Core.Unit/CacheTests/CacheFallbackOperationsTests.cs#L188](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/core/tests/Test.Microsoft.Identity.Core.Unit/CacheTests/CacheFallbackOperationsTests.cs#L188)

---
In reply to: [224727703](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r224727703) [](ancestors = 224727703,224676750,223489018)

henrik-me: @SomskaPe  ?

---
In reply to: [224780601](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r224780601) [](ancestors = 224780601,224727703,224676750,223489018)

core/src/Cache/CacheFallbackOperations.cs
henrik-me: @bgavrilMS, are you ok with this change?

---
In reply to: [224675645](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r224675645) [](ancestors = 224675645,224130526)

henrik-me: @SomskaPe, are the tests in place for this?

---
In reply to: [226489008](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r226489008) [](ancestors = 226489008,224675645,224130526)

henrik-me:
Still there are a number of unaddressed comments in the tests, I marked those as pending. @SomskaPe, kindly answer the still active questions: there are 2 open items to close before this PR can move on. #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/InternalsVisibleTo.cs
SomkaPe: end to end test does not need access to internals, integration test can have access to it, discussed project called "Test.MSAL.NET.Integration" #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: yes it was commited #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: not sure what that means #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
SomkaPe: fewer locks ? There is only one lock for cache. Did you mean shorter critical sections ? Probably yes.
This deserves its own user story and its own pr. #Pending

msal/src/Microsoft.Identity.Client/Properties/InternalsVisibleTo.cs
SomkaPe: As I replied in prev comments Integration test can and in our case absolutely need access to internals. End to End test generally should not.  #Closed

msal/tests/Test.MSAL.NET.Unit/AuthorityAliasesTests.cs
bgavrilMS: All your integration tests invoke logic from the CacheFallbackOperations class. All the methods in this class follow the pattern try { ... } catch all exception and log them. As such, all your integration tests should check that no errors or warnings are ever logged.

---
In reply to: [226488693](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r226488693) [](ancestors = 226488693,224780601,224727703,224676750,223489018)

core/src/Cache/CacheFallbackOperations.cs
bgavrilMS: Yes, sounds good. Better to delete more than to not be able to remove accounts from cache...

---
In reply to: [226489100](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#discussion_r226489100) [](ancestors = 226489100,226489008,224675645,224130526)

henrik-me:
@SomkaPe pls. do the integration and ensure the PR is ready to commit. The most important issues has been addressed. #Closed

msal/tests/Test.MSAL.NET.Integration/UnifiedCacheTests.cs
SomkaPe: I am pretty sure about it - it is not shared #Closed

core/src/Cache/CacheFallbackOperations.cs
SomkaPe: I have added test to CacheFallbackOperationsTests, But I got puzzled after quick look at tests in that class,
@bgavrilMS Why PopulateLegacyCache function is shared by many tests ? some tests call it and add more entities to the cache - so state of the cache is messed up.  
in order to test logic in CacheFallbackOperations you need to consciously configure cache with some state, describe that state  Than run  some logic and check expectations. 
I mean  business logic completely depends on the state of the cache,
so it is crucial incoming parameter to any test. It can not be just some shared random entities 
 #Closed

SomkaPe:
> @SomkaPe pls. do the integration and ensure the PR is ready to commit. The most important issues has been addressed.

@bgavrilMS
@henrik-me 
It is ready to commit.
I merged with dev, and added test case to CacheFallbackOperationsTests.

Not sure how much more time you want to spend on this PR but please pay attention to the fact that all described bugs are pretty severe prod bugs. 

If you wait for some more changes please notify me and describe very specifically what changes you expect (but first check recent commits)

SomkaPe:
@bgavrilMS
@henrik-me 
Any update on this ? #Closed

henrik-me:
Please ensure you document what you consider product bugs as issues. You should link all the issues to this PR.

---
In reply to: [431527007](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#issuecomment-431527007) [](ancestors = 431527007)

henrik-me:
reviewing again now.

---
In reply to: [431973200](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#issuecomment-431973200) [](ancestors = 431973200)

henrik-me (approved):
:shipit:

SomkaPe:
> Please ensure you document what you consider product bugs as issues. You should link all the issues to this PR.
> 
> In reply to: [431527007](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1304#issuecomment-431527007) [](ancestors = 431527007)

Created corresponding issues and added to pr description

PR #1307, opened 2018-10-08
Add PCA with token cache ctor to netcore

bgavrilMS:
The PCA ctor that allows users to pass a custom token cache is not exposed.

### DIFF
--- a/msal/tests/dev apps/NetCoreTestApp/Program.cs
+++ b/msal/tests/dev apps/NetCoreTestApp/Program.cs
@@ -41,7 +41,7 @@ namespace NetCoreTestApp
         private readonly static string ClientIdForPublicApp = "0615b6ca-88d4-4884-8729-b178178f7c27";
         private readonly static string ClientIdForConfidentialApp = "<enter id>";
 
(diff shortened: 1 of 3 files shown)


msal/tests/dev apps/NetCoreTestApp/Program.cs
```diff
@@ -41,15 +41,20 @@ public class Program
         private readonly static string ClientIdForPublicApp = "0615b6ca-88d4-4884-8729-b178178f7c27";
         private readonly static string ClientIdForConfidentialApp = "<enter id>";
 
-        private readonly static string Username = ""; // used for WIA and U/P, can be empty
+        private readonly static string Username = "bogavril@microsoft.com"; // used for WIA and U/P, cannot be empty on .net core
```
henrik-me: >"bogavril@microsoft.com" [](start = 50, length = 24)

perhaps you do not want to commit this one

henrik-me (approved):
:shipit:

msal/tests/dev apps/NetCoreTestApp/Program.cs
bgavrilMS: Yes, good point :)

PR #1309, opened 2018-10-08
Updating broker redirect uri error messages to provide more detail.

trwalke:
The broker redirect URI can be difficult to understand on the android platform. This PR adds a link to https://aka.ms/adal-net-broker-redirect-uri-android to the broker redirect uri error messages on android.


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Exceptions/AdalError.cs
```diff
@@ -294,6 +294,11 @@ public static class AdalError
         /// <summary>
         /// Broker redirectUri has incorrect format
         /// </summary>
-        public const string BrokerRedirectUriIncorrectFormat = "broker_redirect_uri_incorrect_format ";
+        public const string BrokerRedirectUriIncorrect = "broker_redirect_uri_incorrect";
```
henrik-me: >BrokerRedirectUriIncorrect [](start = 28, length = 26)

You can't change this name... even if the text is off. you can change the content however and make another const with the same name... it will potentially break upgrades #Closed

henrik-me (changes requested):
:clock1:

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
```diff
@@ -417,9 +417,13 @@ private Bundle GetBrokerOptions(AuthenticationRequest request)
                 request.Resource);
             string computedRedirectUri = GetRedirectUriForBroker();
 
-            if (!string.IsNullOrEmpty(request.RedirectUri) && !string.Equals(computedRedirectUri, request.RedirectUri, StringComparison.OrdinalIgnoreCase))
+            if (string.IsNullOrWhiteSpace(request.RedirectUri))
```
jennyf19: >string.IsNullOrWhiteSpace(request.RedirectUri) [](start = 16, length = 46)

@trwalke and I synced off-line. I don't think we need this check here. We already check for a null or empty redirectUri in the ValidateRedirectUri(), so is it necessary to check again here? Seems like we should fail fast and not get this far, so maybe failing when creating the broker payload is a better option if we don't already fail sooner? Then we won't need this line, or the new exception, and can just use line 424 w/the new improved error message. thoughts?

trwalke: I agree, I ran a test and a null or empty redirect uri is caught way before this check #Closed

jennyf19: cool...so i would do the fix as we discussed.

---
In reply to: [223531269](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1309#discussion_r223531269) [](ancestors = 223531269)

jennyf19 (approved):
:shipit:

henrik-me (approved):
:shipit:

PR #1310, opened 2018-10-09
Refactor http to reduce static singleton dependencies

MarkZuber:


henrik-me:
Seems like the PR has changes from unrelated PR's as well.

core/src/Platforms/iOS/iOSPlatformProxy.cs
```diff
@@ -72,5 +83,60 @@ public string GetDeviceModel()
...
+        public ITokenCacheAccessor CreateTokenCacheAccessor()
+        {
+            return new iOSTokenCacheAccessor();
+        }
+
+        /// <inheritdoc />
+        public ICryptographyManager CreateCryptographyManager()
+        {
+            return new iOSCryptographyManager();
+        }
```
henrik-me: thinking... should this be cached? We considering the same for some of the other objects. however perhaps this doesn't pay off as those are relatively cheap objects to create.  #Closed

henrik-me:
Looks great! Added a few points... most of them probably just for a quick chat.  #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Platform/IHttpClientFactory.cs
```diff
@@ -29,8 +29,8 @@
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
 {
-    interface IHttpClientFactory
-    {
-        IHttpClient Create(string uri, RequestContext requestContext);
-    }
+    //interface IHttpClientFactory
+    //{
+    //    IHttpClient Create(string uri, RequestContext requestContext);
```
henrik-me: remove completely? #Closed

core/src/PlatformProxyFactory.cs
```diff
@@ -1,66 +1,80 @@
...
+        }
+
+        private static bool IsMsal()
+        {
+#if IS_ADAL
+            return false;
+#elif IS_MSAL
+            return true;
+#else
+            return true;  // TODO: should we default to MSAL or should we default to throw exception?
```
henrik-me: >throw exception? [](start = 85, length = 16)

IMO: Throw. should not ever happen #Closed

core/src/IPlatformProxy.cs
```diff
@@ -32,5 +34,15 @@ internal interface IPlatformProxy
...
+
+        void ValidateRedirectUri(Uri redirectUri, RequestContext requestContext);
+        string GetRedirectUriAsString(Uri redirectUri, RequestContext requestContext);
+        string GetDefaultRedirectUri(string correlationId);
+
+        string GetProductName();
+
+        ILegacyCachePersistance CreateLegacyCachePersistence();
+        ITokenCacheAccessor CreateTokenCacheAccessor();
+        ICryptographyManager CreateCryptographyManager();
```
henrik-me: I made most of my comments on the "Create" pattern in the implementation class. Adding here that perhaps we should follow the lazy pattern used in the platform proxy class to create the platform information objects instead of newing  "helper"objects every time? Perhaps the gain is neglectable. 
My thinking is not for the client side scenarios but rather server side scenarios. #Closed

henrik-me:
>                 return version[1];

Outside the scope of the refactoring,,, but considering introducing caching here... using well, a static given the class is static. 
perhaps it would be worth caching this value? no need for every telemetry entry to re-calculate this value.
thoughts?

Aside: perhaps this should really be moved to the AssemblyUtils class? 


[mzuber] -- I'm going to make a note in my refactoring onenote to look at cleaning up this class since I agree it's weird for us to recalculate this stuff constantly.  but i'm going to push it out of this PR and do it later. #Pending

---
Refers to: core/src/MsalIdHelper.cs:111 in e980b31. [](commit_id = e980b3170756ce718e96b831d8b95d315e197342, deletion_comment = False)

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
```diff
@@ -33,8 +33,9 @@
...
 using Microsoft.Identity.Core;
 using Microsoft.Identity.Core.Telemetry;
-using System.Threading;
-
+using System.Threading;
+using Microsoft.Identity.Core.Http;
+
 namespace Microsoft.Identity.Client
 {
     /// <summary>
```
henrik-me, line 41: >/// <summary> [](start = 4, length = 13)

was thinking if we can do inherit docs for both PBA and CCA? #Pending

msal/src/Microsoft.Identity.Client/Internal/ModuleInitializer.cs
```diff
@@ -33,29 +33,29 @@ namespace Microsoft.Identity.Client.Internal
...
-    /// The CLR defines a module initializer, however this is not implemented in C# and to 
+    /// The CLR defines a module initializer, however this is not implemented in C# and to
     /// use this it would require IL weaving, which does not seem to work on all target frameworks.
     /// Instead, call <see cref="EnsureModuleInitialized"/> from static ctors of public entry points.
     /// </remarks>
     internal class ModuleInitializer
     {
-        private static bool isInitialized = false;
-        private static readonly object lockObj = new object();
+        private static bool _isInitialized = false;
```
henrik-me: >_isInitialized [](start = 28, length = 14)

should this be volatile ? #Closed

msal/src/Microsoft.Identity.Client/Internal/MsalLogger.cs
```diff
@@ -137,7 +137,7 @@ private void Log(LogLevel msalLogLevel, string messageWithPii, string messageScr
                 ? string.Empty
                 : " - " + CorrelationId;
 
-            var msalIdParameters = MsalIdHelper.GetMsalIdParameters(new PlatformInformation());
+            var msalIdParameters = MsalIdHelper.GetMsalIdParameters();
```
henrik-me, line 140: >msalIdParameters [](start = 16, length = 16)

can this be "cached" for the lifetime of logger? perhaps even injected at creation time (or static ctr) #Closed

msal/src/Microsoft.Identity.Client/Internal/PlatformInformationBase.cs
```diff
@@ -30,12 +30,7 @@
...
-    {
-        public override string GetAssemblyFileVersionAttribute()
-        {
-            return
-                typeof(MsalIdHelper).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>().Version;
-        }
-    }
+    //internal abstract class PlatformInformationBase : CorePlatformInformationBase
+    //{
+    //}
```
henrik-me: remove completely ? #Closed

henrik-me:
some of these change reminds me .... I feel we are missing micro/macro performance tests which can quickly help assess whether we will improve especially service performance... which is overall my biggest concern for the library when it comes to performance and memory usage #Pending

msal/tests/Test.MSAL.NET.Unit/ConfidentialClientApplicationTests.cs
```diff
@@ -295,392 +314,514 @@ private ConfidentialClientApplication CreateConfidentialClient(ClientCredential
...
+                Assert.AreEqual("offline_access openid profile r1/scope1 r1/scope2", qp["scope"]);
+                Assert.AreEqual(TestConstants.ClientId, qp["client_id"]);
+                Assert.AreEqual("code", qp["response_type"]);
+                Assert.AreEqual(TestConstants.RedirectUri, qp["redirect_uri"]);
+                Assert.AreEqual(TestConstants.DisplayableId, qp["login_hint"]);
+                Assert.AreEqual(UIBehavior.SelectAccount.PromptValue, qp["prompt"]);
+                Assert.AreEqual("MSAL.Desktop", qp["x-client-sku"]);
+                Assert.IsFalse(string.IsNullOrEmpty(qp["x-client-ver"]));
+                Assert.IsFalse(string.IsNullOrEmpty(qp["x-client-cpu"]));
+                Assert.IsFalse(string.IsNullOrEmpty(qp["x-client-os"]));
```
henrik-me, line 412: seems like most of the validation stuff is the same in more places. ... seems duplicate and makes me think that some refactorings are needed and perhaps the tests are not testing different things... #Closed

core/tests/Test.Microsoft.Identity.Core.Unit/Mocks/MockHttpManager.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <inheritdoc />
+        public void Dispose()
+        {
+            // This ensures we only check the mock queue on dispose when we're not in the middle of an
+            // exception flow.  Otherwise, any early assertion will cause this to likely fail
+            // even though it's not the root cause.
+            if (Marshal.GetExceptionCode() == 0)
+            {
+                Assert.AreEqual(0, _httpMessageHandlerQueue.Count, "All mocks should have been consumed");
```
henrik-me, line 57: > Assert.AreEqual(0, _httpMessageHandlerQueue.Count, "All mocks should have been consumed"); [](start = 15, length = 91)

interesting approach, 

nit as it is easy to attach a debugger...: thought perhaps list the handlers which are still in the queue... to make debugging easier? #Closed

henrik-me:
> <wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">

Is this the netcore ruleset? 
Should we have the same for the one with samples? .... would have liked the same for vs settings... as I do not currently use resharper.

[mzuber] -- this is very close to the netcore ruleset but i didn't copy it from there (we definitely can do that instead).  VS doesn't do _quite_ enough yet to get code formatting consistently so I'm doing this to get some of the major cleanup points done automatically.  and we can tweak the ruleset and see what we can do in plain vanilla VS so we can get more consistent here. #Pending

---
Refers to: LibsNoSamples.sln.DotSettings:1 in e980b31. [](commit_id = e980b3170756ce718e96b831d8b95d315e197342, deletion_comment = False)

henrik-me (changes requested):
:clock1:

core/src/PlatformProxyFactory.cs
```diff
@@ -1,66 +1,80 @@
...
-            new NetDesktopPlatformProxy()
+            new NetDesktopPlatformProxy(IsMsal())
 #endif
         );
 
+        private PlatformProxyFactory()
+        {
+        }
+
+        private static bool IsMsal()
```
bgavrilMS, line 61: I think PlatformProxy does too much now. My initial refactoring idea was to break up the old "PlatformInformation" into a PlatformProxy that focuses only on OS and .net flavour (read the upn from the OS etc.) and a LibraryProxy that would handle things like "get version".  #Pending

core/src/Platforms/iOS/iOSPlatformProxy.cs
```diff
@@ -72,5 +83,60 @@ public string GetDeviceModel()
         {
             return UIDevice.CurrentDevice.Model;
         }
+
+        /// <inheritdoc />
+        public void ValidateRedirectUri(Uri redirectUri, RequestContext requestContext)
+        {
+            if (redirectUri == null)
```
bgavrilMS: string.IsNullOrEmpty

core/src/Platforms/iOS/iOSPlatformProxy.cs
MarkZuber: yes, I've changed these to be class initialized properties so they're only constructed once. #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Platform/IHttpClientFactory.cs
MarkZuber: fixed #Closed

core/src/PlatformProxyFactory.cs
MarkZuber: fixed #Closed

core/src/IPlatformProxy.cs
MarkZuber: I've fixed this.  please check updated code once I push. #Closed

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
MarkZuber: Yes, I want to make another documentation cleanup pass.  Added notes to my backlog. #Pending

msal/src/Microsoft.Identity.Client/Internal/ModuleInitializer.cs
MarkZuber: yes.  fixed. #Closed

msal/src/Microsoft.Identity.Client/Internal/MsalLogger.cs
MarkZuber: yes, it can be.  backlog of notes for future refactor. #Pending

msal/src/Microsoft.Identity.Client/Internal/PlatformInformationBase.cs
MarkZuber: fixed #Closed

msal/tests/Test.MSAL.NET.Unit/ConfidentialClientApplicationTests.cs
MarkZuber: yes, i had similar feelings on copy/paste and duplication across many of the tests, this being a prime example.  test cleanup and organization is also in my backlog. #Pending

core/tests/Test.Microsoft.Identity.Core.Unit/Mocks/MockHttpManager.cs
MarkZuber: Given that the mocks are added in order in the tests (so it's easy to find the last one or two if they're missed), and the handlers don't "tostring" very cleanly i don't think this is worth the effort here.  but let's see as we develop and if we can make improvements here we certainly will. #Closed

core/src/PlatformProxyFactory.cs
MarkZuber: I'd like to proceed with the current refactor since all of the code being moved in here is paltform specific.  Personally, I like the idea of having some root object (like platform proxy) that gives is the central core for how we do platform disambiguation.  But if we want to break this down into smaller parts hanging off platformproxy or something else, let's look into that.   #Pending

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/NonInteractive/AcquireTokenIWAHandler.cs
```diff
@@ -59,24 +60,24 @@ public AcquireTokenIWAHandler(RequestData requestData, IntegratedWindowsAuthInpu
                 SupportADFS = true;
             }
 
-            this.iwaInput = iwaInput;
+            this._iwaInput = iwaInput;
```
jennyf19, line 63: >this [](start = 12, length = 4)

why do you still need `this.` here?

henrik-me (approved):
:shipit:

PR #1311, opened 2018-10-09
remove duplicate code to webviewBase

jennyf19:
Tested w/iOS simulator w/locked screen and exiting during auth w/and w/out MFA w/system browser and embedded webview - adal and msal (respectively)

### DIFF
--- a/core/src/Platforms/iOS/WebviewBase.cs
+++ b/core/src/Platforms/iOS/WebviewBase.cs
@@ -88,5 +88,12 @@ namespace Microsoft.Identity.Core.UI
                 taskId = UIApplication.BackgroundTaskInvalid;
             }
         }
+
+        //Hiding NSObject.Dispose() with new to implement IDisposable interface
+        public new void Dispose()
+        {
+            didEnterBackgroundNotification.Dispose();
+            willEnterForegroundNotification.Dispose();
+        }
     }
 }
--- a/core/src/Platforms/iOS/WebviewBase.cs
+++ b/core/src/Platforms/iOS/WebviewBase.cs
@@ -42,6 +42,10 @@ namespace Microsoft.Identity.Core.UI
         protected SFAuthenticationSession sfAuthenticationSession;
         protected ASWebAuthenticationSession asWebAuthenticationSession;
 
+        protected nint taskId = UIApplication.BackgroundTaskInvalid;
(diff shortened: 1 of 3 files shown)


core/src/Platforms/iOS/WebviewBase.cs
```diff
@@ -42,6 +42,10 @@ internal abstract class WebviewBase : NSObject, IWebUI, ISFSafariViewControllerD
         protected SFAuthenticationSession sfAuthenticationSession;
         protected ASWebAuthenticationSession asWebAuthenticationSession;
 
+        protected nint taskId = UIApplication.BackgroundTaskInvalid;
+        protected NSObject didEnterBackgroundNotification;
+        protected NSObject willEnterForegroundNotification;
```
henrik-me, line 47: Seems like you are no longer disposing these guys. It should probably be ensured that happens. #Closed

henrik-me (changes requested):
:clock1:

core/src/Platforms/iOS/WebviewBase.cs
```diff
@@ -60,5 +64,36 @@ public static bool ContinueAuthentication(string url)
...
+        {
+            if (taskId != UIApplication.BackgroundTaskInvalid)
+            {
+                UIApplication.SharedApplication.EndBackgroundTask(taskId);
+                taskId = UIApplication.BackgroundTaskInvalid;
+            }
+        }
+
+        //Hiding NSObject.Dispose() with new to implement IDisposable interface
+        public new void Dispose()
```
henrik-me: >new [](start = 15, length = 3)

the usual dispose pattern... I believe has an overload which takes a bool which can be overridden. You should probably use that one, otherwise you are not guaranteed the dispose handler is called at the expected time/if at all (unless the one calling does so on the actual type)

henrik-me:
nice refactoring! Pls. ensure you validate that dispose is called at the expected time. Also pls. validate if you can use the std. dispose pattern instead of hiding the dispose of the base object.

henrik-me (approved):
:shipit:

core/src/Platforms/iOS/WebviewBase.cs
jennyf19: sounds good. thanks for the feedback.

PR #1315, opened 2018-10-11
Fix [631] Device Code Flow logs too many errors while polling

MarkZuber:


core/src/Exceptions/CoreException.cs
```diff
@@ -0,0 +1,85 @@
...
+
+using System;
+using System.Globalization;
+
+namespace Microsoft.Identity.Core.Exceptions
+{
+    /// <summary>
+    ///     Abstract base exception, never thrown by itself.
+    /// </summary>
+    public abstract class CoreException : Exception
```
jmprieur: do we want to expose the Core namespace? so far we have avoided doing that?
This probably means that customers won't be able to have ADAL.NET and MSAL.NET in the same application. We know that some customers do that (@bgavrilMS can elaborate)

jmprieur (commented):
I expresed my worry about exposing something public from the core namespace.

core/src/Exceptions/CoreException.cs
MarkZuber: Customers should never see this exception directly since it's abstract and they'll still directly interact with MsalException or AdalException.  But OAuth2Client is in Core and we need to get to the ErrorCode value in order to fix this issue, so I had to get a common exception type in the base that we can check for.  Interested in advice/guidance to move this into a separate namespace or elsewhere since we can't make this an internal type because it is derived by public types.

henrik-me: This would though be exposed in the API.... I believe we need to remove this one. I suppose it will actually also be a breaking change in the API.

---
In reply to: [224602950](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1315#discussion_r224602950) [](ancestors = 224602950)

henrik-me (changes requested):
:clock1:

core/src/OAuth2/OAuth2Client.cs
```diff
@@ -166,7 +167,26 @@ public static void CreateErrorResponse(HttpResponse response, RequestContext req
...
+                {
+                    requestContext.Logger.InfoPii(serviceEx);
+                    shouldLogError = false;
+                }
+            }
+
+            if (shouldLogError)
+            {
+                requestContext.Logger.ErrorPii(serviceEx);
+            }
```
henrik-me, line 189: my suggestion would be to limit the fix to this.. using the existing exceptions. The ExceptionFactory might be the way to filter on the type. Not sure that the change for ADAL is possible as it will be breaking customers.  #Closed

core/src/Exceptions/CoreException.cs
bgavrilMS: Too bad exceptions don't play nice with interfaces...

core/src/OAuth2/OAuth2Client.cs
```diff
@@ -166,7 +167,26 @@ public static void CreateErrorResponse(HttpResponse response, RequestContext req
...
 
-            requestContext.Logger.ErrorPii(serviceEx);
+            // For device code flow, AuthorizationPending can occur a lot while waiting
+            // for the user to auth via browser and this causes a lot of error noise in the logs.
+            // So suppress this particular case to an Info so we still see the data but don't 
+            // log it as an error since it's expected behavior while waiting for the user.
+            bool shouldLogError = true;
+            if (serviceEx is CoreException coreException)
+            {
+                if (string.Compare(coreException.ErrorCode, OAuth2Error.AuthorizationPending,
```
bgavrilMS: The exceptionCode seems to be local to this method, so not sure we really need to add all this logic (i.e. exposing a core exception is problematic for the reasons JM and Henrik mentioned, otherwise I agree that is the right solution for exposing the error code). 

Alternatively, we could add another method in the factory, something like ExceptionFactory.Instance.GetErrorCode where we would let each factory implementation pull out the error code.

msal/src/Microsoft.Identity.Client/Logger.cs
```diff
@@ -117,8 +117,6 @@ public sealed class Logger
         /// and a second time with the <c>containsPii</c> parameter equals to <c>true</c> and the message might contain PII. 
         /// In some cases (when the message does not contain PII), the message will be the same.
         /// <para/>
-        /// The property can only be set once and it will throw an ArgumentException if called twice.
-        /// For details see https://aka.ms/msal-net-logging
```
henrik-me: >For details see https://aka.ms/msal-net-logging [](start = 12, length = 47)

why remove the aka.ms link? #Closed

msal/src/Microsoft.Identity.Client/Logger.cs
```diff
@@ -127,17 +125,17 @@ public static LogCallback LogCallback
             {
                 lock (LockObj)
                 {
-                    if (_logCallback != null)
-                    {
-                        throw new ArgumentException("MSAL logging callback can only be set once per process and" +
-                                                   " should never change once set.");
-                    }
```
henrik-me, line 134: I wonder what the original idea to have this was... I suppose removal is then handled by pasing null #Closed

msal/tests/Test.MSAL.NET.Unit/RequestsTests/DeviceCodeRequestTests.cs
```diff
@@ -95,7 +97,7 @@ private HttpResponseMessage CreateDeviceCodeResponseSuccessMessage()
         [TestCategory("DeviceCodeRequestTests")]
         public void TestDeviceCodeAuthSuccess()
         {
-            var parameters = CreateAuthenticationParametersAndSetupMocks(out HashSet<string> expectedScopes);
+            var parameters = CreateAuthenticationParametersAndSetupMocks(1, out HashSet<string> expectedScopes);
```
henrik-me: >1 [](start = 73, length = 1)

this magic 1 here should imo have had a named const to make it more readable. (sorry for the late comment) #Closed

henrik-me:
main issue addressed thus signing off. still a minor ask... actually blocking again until we have a decision on the logger.

henrik-me (approved):
:shipit:

msal/src/Microsoft.Identity.Client/Logger.cs
jennyf19: @henrik-me  maybe this? https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/174 -> "Make sure to add the check for failing if developer tries to set callback twice" #Closed

henrik-me: That seems to be the reason... I don't get the reason. However we should naturally ensure that nothing breaks even if the logger is set to null. That is imo we should curretly consider to throw an exception if someone tries to set it to null.... perhaps.

---
In reply to: [224878188](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1315#discussion_r224878188) [](ancestors = 224878188)

henrik-me (changes requested):
:clock1:

msal/src/Microsoft.Identity.Client/Logger.cs
MarkZuber: i put it back. #Closed

msal/src/Microsoft.Identity.Client/Logger.cs
MarkZuber: The original bug (thanks jenny for the link) was basically "once its set, don't let someone else set it".  for testing, we need to be able to clear it.  so i can either leave it as it is (and then the original issue is present that someone could hijack the log callback, but technically that's a bug in your own app and we should technically allow multiple callbacks for that case) or I can make it so you can only set it back to null (which still allows a hijack {set null, set to my callback}).  

I'd prefer to leave it how it is even though it returns the ability to hijack since we should be able to safely clear the callback. #Closed

msal/tests/Test.MSAL.NET.Unit/RequestsTests/DeviceCodeRequestTests.cs
MarkZuber: fixed #Closed

msal/src/Microsoft.Identity.Client/Logger.cs
henrik-me: I kind of agree. 

---
In reply to: [224880996](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1315#discussion_r224880996) [](ancestors = 224880996)

henrik-me (approved):
:shipit:

PR #1317, opened 2018-10-12
allow teamid to be accessible

jennyf19:
Fixes MSAL.NET #[626](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/626): PublicClientApplication constructor throws exception when iPhone screen is locked

According to olga, the TeamId can be accessible always as there is no private info in it. This is to access TeamId when the phone is locked.

### DIFF
--- a/core/src/Platforms/iOS/TokenCacheAccessor.cs
+++ b/core/src/Platforms/iOS/TokenCacheAccessor.cs
@@ -86,7 +86,7 @@ namespace Microsoft.Identity.Core
             {
                 Service = "",
                 Account = TeamIdKey,
-                Accessible = SecAccessible.Always
+                Accessible = _defaultAccessiblityPolicy
             };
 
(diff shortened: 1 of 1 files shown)


core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -85,7 +85,8 @@ private string GetTeamId()
             var queryRecord = new SecRecord(SecKind.GenericPassword)
             {
                 Service = "",
-                Account = TeamIdKey
+                Account = TeamIdKey,
+                Accessible = SecAccessible.Always
```
henrik-me: >Always [](start = 43, length = 6)

I would suggest we use: _defaultAccessiblityPolicy, as that is what we use for everything else. not sure we need more than that. However Olga knows best!

henrik-me (approved):
:shipit:

PR #1318, opened 2018-10-12
Jennyf/ios entitlements error

jennyf19:
checking for missing entitlements while saving to keychain and throwing useful error message #supportability

### DIFF
--- a/core/src/Exceptions/CoreErrorMessages.cs
+++ b/core/src/Exceptions/CoreErrorMessages.cs
@@ -97,5 +97,9 @@ namespace Microsoft.Identity.Core
            "The application cannot access the iOS keychain for the application publisher (the TeamId is null). " +
            "This is needed to enable Single Sign On between applications of the same publisher. " +
            "This is an iOS configuration issue. See https://aka.ms/msal-net-enable-keychain-access for more details on enabling keychain access.";
+
+        public const string MissingEntitlements =
+            "The application is missing the necessary entitlements. As a result, there was a failure to save to the iOS keychain." +
+            " See https://aka.ms/msal-net-enable-keychain-groups for more details on enabling keychain groups and entitlements.";
     }
 }
--- a/core/src/Exceptions/CoreErrorMessages.cs
+++ b/core/src/Exceptions/CoreErrorMessages.cs
@@ -99,7 +99,8 @@ namespace Microsoft.Identity.Core
            "This is an iOS configuration issue. See https://aka.ms/msal-net-enable-keychain-access for more details on enabling keychain access.";
 
         public const string MissingEntitlements =
-            "The application is missing the necessary entitlements. As a result, there was a failure to save to the iOS keychain." +
-            " See https://aka.ms/msal-net-enable-keychain-groups for more details on enabling keychain groups and entitlements.";
+            "The application does not have keychain access groups enabled in the Entitlements.plist. " +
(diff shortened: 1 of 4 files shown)


core/src/Exceptions/CoreErrorMessages.cs
```diff
@@ -97,5 +97,9 @@ internal class CoreErrorMessages
            "The application cannot access the iOS keychain for the application publisher (the TeamId is null). " +
            "This is needed to enable Single Sign On between applications of the same publisher. " +
            "This is an iOS configuration issue. See https://aka.ms/msal-net-enable-keychain-access for more details on enabling keychain access.";
+
+        public const string MissingEntitlements =
+            "The application is missing the necessary entitlements. As a result, there was a failure to save to the iOS keychain." +
```
jmprieur: Maybe "The application is missing the necessary entitlements.plist?" ? #Closed

jmprieur (approved):
LGTM

jennyf19:
Forgot to mention I tested w/the iOS Xforms app (simulator) in msal.  #Closed

core/src/Exceptions/CoreErrorMessages.cs
jennyf19: @jmprieur Reading it again, I decided to make the message more specific. thx for the feedback. #Closed

henrik-me:
Perfect. JM said he will test on a device

---
In reply to: [429211089](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1318#issuecomment-429211089) [](ancestors = 429211089)

henrik-me (approved):
:shipit:

jennyf19:
[MSAL issue](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/633)

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -284,6 +284,13 @@ private SecStatusCode Save(string account, string service, string generic, int t
                 secStatusCode = SecKeyChain.Add(recordToSave);
             }
 
+            if(secStatusCode == SecStatusCode.MissingEntitlement)
+            {
+                throw CoreExceptionFactory.Instance.GetClientException(
```
DaRosenberg, line 289: It would be nice if the resulting exception message contained the keychain group that the accessor is attempting to save to, like:

`"The application does not have the keychain access group '{0}' enabled in the Entitlements.plist."`

henrik-me: Makes sense to also add it here. The developer has specified it in their own code though, thus they should know? I suppose for the case where the developer haven't set it, the default value will be shown here.

---
In reply to: [224734206](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1318#discussion_r224734206) [](ancestors = 224734206)

DaRosenberg: @henrik-me Not necessarily - the TokenCacheAccessor might be using its default value for the keychain group, in which case it's super helpful to know what it's trying to do.

DaRosenberg: Haha, I think I replied before you edited your comment. :)

PR #1321, opened 2018-10-12
Fix FxCop error CA1305 complaining about globalization string

bgavrilMS:


core/src/Helpers/StringWriterWithEncoding.cs
```diff
@@ -25,17 +25,20 @@
...
 namespace Microsoft.Identity.Core.Helpers
 {
     internal class StringWriterWithEncoding : StringWriter
-    {
-        public StringWriterWithEncoding(Encoding encoding)
-        {
+    {
+
+        public StringWriterWithEncoding(Encoding encoding) 
+            : base(CultureInfo.InvariantCulture)
```
bgavrilMS, line 38: Derived class calls base parameterless constructor implicitly, which violates CA1305. Setting the default culture works around it.

jennyf19 (approved):
:shipit: thanks @bgavrilMS

Issue #1277, opened 2018-09-27
Cross-thread operation exception when setting ownerWindow in PlatformParameters 


rhythmnewt:
**Which Version of ADAL are you using ?**
ADAL 3.18.0 and later

**Which platform has the issue?**
net462

**What authentication flow has the issue?**
* Desktop 
    * [ X] Interactive
 
Other? - please describe;

**Repro**
Create new WinForms project targeting .NET 462 and add the following code to handle Load event.
```csharp
   private async void Form1_Load(object sender, EventArgs e)
        {
            WindowState = FormWindowState.Normal;

            //fill in as necessary for correct auth
            const string resource = "https://www.myresource.com/";
            const string clientId = "client_id_guid";
            const string redirectUri = "https://myredirecturi";
            const string authority = "https://login.windows.net/myauthority/oauth2/token";

            var context = new AuthenticationContext(authority, true, TokenCache.DefaultShared);

            var result = await context.AcquireTokenAsync(resource, clientId,
                new Uri(redirectUri),
                new PlatformParameters(PromptBehavior.SelectAccount, this));
        }
```

**Expected behavior**
Windows form is shown in the background with the browser login window in the foreground. The browser login window is the child of the windows form.

**Actual behavior**
Exception is thrown: 
System.InvalidOperationException: 'Cross-thread operation not valid: Control 'Form1' accessed from a thread other than the thread it was created on.'

**Possible Solution**

**Additional context/ Logs / Screenshots**
[WindowsFormsApp1.zip](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/files/2425029/WindowsFormsApp1.zip)

If you downgrade the library to 3.17.3 there is no issue with the cross-threading.

rhythmnewt:
Bueller?

MarkZuber:
@rhythmnewt 

Still investigating.  But I have a short term workaround that should hopefully get you unblocked.  

When you construct the PlatformParameters, send in "null" instead of "this" for the owner window.  The owner window is only used for trying to center the browser around the owner app window.  If it's null, then the browser will be centered on the current screen.  That should be the only difference in functionality and I'll get back to this thread once I've gotten deeper on the issue with ownerWindow.

MarkZuber:
Pull request for fix is out:  https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1328

henrik-me:
[...]

PR #1323, opened 2018-10-12
Fix for [1277] Cross-thread operation exception when setting ownerWindow in PlatformParameters 

MarkZuber:
This PR is still a hack but I want to start talking about it with the team.  The concept itself works, but the way we're passing the SynchronizationContext around is totally wrong.  We probably want to plumb it through the same way as the ownerWindow object.

Our sample adal app has been working because it's actually a console app and so the STA scheduler creates the first and only STA thread and there isn't a UI thread / winforms object to interact with.  If we have a winforms app (such as the example provided in #1277) then because of the async methods added in restructuring between 3.17.x and 3.18, we're on an MTA thread.  We create a _new_ STA thread with the STA Scheduler and that isn't the same thread that the UI is running on, so we get the error as shown in the bug.

The proposed fix is to use SynchronizationContext which should work from both WinForms and WPF applications (both implement it in their main UI thread) and then we can call SynchronizationContext.Send() with the delegate method doing the browser loop.  This is the equivalent of calling Control.Invoke() without having a Control object.  In the case of our sample console app, we can create a new SynchronizationContext in the root and use that to execute with but we still need the STA thread to be created because our sample doesn't create a default STA thread since it's a console app.

We could look into deciding not to support raw console applications and require winforms/wpf which would then bring in the assurance we have an STAThread at root and then we could remove the STAScheduler altogether.  The STAScheduler is only needed to cover the console app case in our sample.

### DIFF
--- a/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/PlatformParameters.cs
+++ b/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/PlatformParameters.cs
@@ -25,6 +25,7 @@
 //
 //------------------------------------------------------------------------------
 
+using System.Threading;
 using Microsoft.Identity.Core.UI;
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory
@@ -34,6 +35,8 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     /// </summary>
     public class PlatformParameters : IPlatformParameters
     {
+        public static SynchronizationContext _syncContext = null;
+
         /// <summary>
         /// 
         /// </summary>
@@ -49,6 +52,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
         /// <param name="ownerWindow"></param>
         public PlatformParameters(PromptBehavior promptBehavior, object ownerWindow)
         {
+            _syncContext = SynchronizationContext.Current;
             this.PromptBehavior = promptBehavior;
             this.OwnerWindow = ownerWindow;
         }
--- a/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/PlatformParameters.cs
+++ b/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/PlatformParameters.cs
@@ -52,7 +52,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
         /// <param name="ownerWindow"></param>
         public PlatformParameters(PromptBehavior promptBehavior, object ownerWindow)
         {
-            _syncContext = SynchronizationContext.Current;
+            _syncContext = SynchronizationContext.Current ?? new SynchronizationContext();
             this.PromptBehavior = promptBehavior;
             this.OwnerWindow = ownerWindow;
         }
--- a/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/WebUI.cs
+++ b/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/WebUI.cs
@@ -48,42 +48,13 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
         {
             AuthorizationResult authorizationResult = null;
 
-            var sendAuthorizeRequest = new Action(
-                delegate
+            var other = PlatformParameters._syncContext;
+            other.Send(state =>
                 {
-                    authorizationResult = this.Authenticate(authorizationUri, redirectUri);
-                });
-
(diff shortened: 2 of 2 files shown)


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/PlatformParameters.cs
```diff
@@ -49,6 +52,7 @@ public PlatformParameters(PromptBehavior promptBehavior):this(promptBehavior, nu
         /// <param name="ownerWindow"></param>
         public PlatformParameters(PromptBehavior promptBehavior, object ownerWindow)
         {
+            _syncContext = SynchronizationContext.Current ?? new SynchronizationContext();
```
henrik-me, line 55: >_syncContext [](start = 12, length = 12)

Does this enforce the ui thread context? For other libraries we have put the synhronization responsiblities on the caller, however naturally we have to open the  web ui on the UI thread, that should be done where the web view is created though (assuming the controls are not created on the ui thread), Is suppose it makes sense to pass the sync context around ... in my book we would have been able to use the OwnerWindow to get the UI thread...

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/WebUI.cs
```diff
@@ -48,11 +48,16 @@ public async Task<AuthorizationResult> AcquireAuthorizationAsync(Uri authorizati
...
             AuthorizationResult authorizationResult = null;
 
-            var sendAuthorizeRequest = new Action(
-                delegate
-                {
-                    authorizationResult = this.Authenticate(authorizationUri, redirectUri);
-                });
+            var sendAuthorizeRequest = new Action(() =>
+            {
+                var other = PlatformParameters._syncContext;
```
henrik-me, line 53: >_syncContext [](start = 47, length = 12)

alternatively it should be done by invoking the call on the ownerwindow thread?

henrik-me:
Makes sense, thanks for the detailed explanation.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/PlatformParameters.cs
```diff
@@ -34,6 +35,8 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
     /// </summary>
     public class PlatformParameters : IPlatformParameters
     {
+        public static SynchronizationContext _syncContext = null;
```
bgavrilMS, line 38: Do you want to expose a public getter instead of a member?

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/WebUI.cs
bgavrilMS: @henrik-me - we expect the ownerwindow to be a Handle (intptr handle), see https://github.com:443/AzureAD/azure-activedirectory-library-for-dotnet.git/blob/master/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/WindowsFormsWebAuthenticationDialogBase.cs#L94

Also, it seems the "ownerwindow" is only used to center the dialog, and in most our samples it is null. So we can expect most people will not have set it.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/PlatformParameters.cs
bgavrilMS: So this is assuming that the PlatformParameters will be created on the UI thread. Which makes sense if you are passing a window object anyway. However, ownerWindow is only used to find another window and to center the browser UI onto it. Most people will pass in null (and so do our samples). So, it is quite possible that we users will create PlatformParamters on a non-UI thread.

I just wonder if we should be more explicit to our users, perhaps through the PlatformParamters comments (which are missing completely) or better yet via another constructor that takes in a syncronization context ? 

CCing @jmprieur for documentation implications..

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/PlatformParameters.cs
bgavrilMS: *internal 

---
In reply to: [225124992](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1323#discussion_r225124992) [](ancestors = 225124992)

MarkZuber:
New fix is here:  https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1328 

closing this out as we've taken a different approach to fix this.

PR #1324, opened 2018-10-14
- fix for accessible setting for TeamId

jennyf19:
- change teamIdKey default value

I talked w/Olga on Friday, and these two fixes should address the issues I was having Friday w/iOS simulator. The teamID access needs to be always. 

@bgavrilMS we will need to test this before running a release build. I have tested and it works fine on simulator. 

cc: @jmprieur

### DIFF
--- a/core/src/Platforms/iOS/TokenCacheAccessor.cs
+++ b/core/src/Platforms/iOS/TokenCacheAccessor.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Identity.Core
         private const SecAccessible _defaultAccessiblityPolicy = SecAccessible.AfterFirstUnlockThisDeviceOnly;
 
         private readonly string DefaultKeychainGroup = "com.microsoft.adalcache";
-        private readonly string TeamIdKey = "teamIDHint";
+        private readonly string TeamIdKey = ".NetTeamIDHint";
 
         private string keychainGroup;
         private RequestContext _requestContext;
@@ -87,7 +87,7 @@ namespace Microsoft.Identity.Core
             {
                 Service = "",
                 Account = TeamIdKey,
-                Accessible = _defaultAccessiblityPolicy
+                Accessible = SecAccessible.Always
             };
 
             SecRecord match = SecKeyChain.QueryAsRecord(queryRecord, out SecStatusCode resultCode);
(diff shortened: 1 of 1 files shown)


jmprieur:
thanks @jennyf19 
Is this paragraph still relevant: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Xamarin-ios-specificities#enable-keychain-access ?

core/src/Platforms/iOS/TokenCacheAccessor.cs
```diff
@@ -59,7 +59,7 @@ enum CredentialAttrType
         private const SecAccessible _defaultAccessiblityPolicy = SecAccessible.AfterFirstUnlockThisDeviceOnly;
 
         private readonly string DefaultKeychainGroup = "com.microsoft.adalcache";
-        private readonly string TeamIdKey = "teamIDHint";
+        private readonly string TeamIdKey = ".NetTeamIDHint";
```
bgavrilMS: 1. Is this a breaking change for iOS cache wise?
2. Use const instead of readonly. 
3. Could you add a comment in the code on what this constant controls ?  #Resolved

bgavrilMS (approved):
Approved with comments

core/src/Platforms/iOS/TokenCacheAccessor.cs
henrik-me: I'm still not sure why we really need to do this. I get that it may be failing on the test device you are using, however I assume it's possible to reset the device or remove the record from the device?
(The default access policy hasn't shipped yet)

---
In reply to: [225133657](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1324#discussion_r225133657) [](ancestors = 225133657)

henrik-me (changes requested):
:clock1:

core/src/Platforms/iOS/TokenCacheAccessor.cs
jennyf19: 1. no. Not a breaking change because we will write a new entry if the old one is not found. The reason this broke earlier is because we have to use SecAccessible.Always for TeamId.
2. Yes. Fixed
3. Yes. 

---
In reply to: [225223957](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1324#discussion_r225223957) [](ancestors = 225223957,225133657)

jennyf19: @henrikm the default value is `AfterFirstUnlockThisDeviceOnly`, which is set when you don't specify anything (which is what we were doing previously). Now that we want to set the accessibility to something less restrictive (only for TeamId), we need to change the Team identifier. 

---
In reply to: [225223957](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1324#discussion_r225223957) [](ancestors = 225223957,225133657)

henrik-me (approved):
:shipit:

PR #1326, opened 2018-10-15
Fixing issue where the redirect URI is checked incorrectly when perfo…

trwalke:
Fixing issue where the redirect URI is checked incorrectly when perfo…rming the silent authentication broker flow on android.

This was testing with the "Interactive deviceAuth for MAM devices through Azure Authenticator" and "Silent deviceAuth for MAM devices through Azure Authenticator" manual tests

### DIFF
--- a/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
+++ b/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
@@ -452,13 +452,13 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
 
         private void ValidateBrokerRedirectURI(AuthenticationRequest request)
         {
-            string computedRedirectUri = GetRedirectUriForBroker();
-
             //During the silent broker flow, the redirect URI will be null.
             if (string.IsNullOrEmpty(request.RedirectUri))
             {
                 return;
             }
+
+            string computedRedirectUri = GetRedirectUriForBroker();
 
             if (!string.Equals(computedRedirectUri, request.RedirectUri, StringComparison.OrdinalIgnoreCase))
             {
--- a/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
+++ b/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
@@ -417,7 +417,9 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
                 request.Resource);
             string computedRedirectUri = GetRedirectUriForBroker();
 
-            if (!string.Equals(computedRedirectUri, request.RedirectUri, StringComparison.OrdinalIgnoreCase))
+            //Do not remove the null or empty check from the following if statement. During the silent broker flow, there is an expectation that the
+            //redirect URI will be null. Thus, disabling this check in that scenario.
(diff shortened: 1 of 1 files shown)


jennyf19 (approved):
:shipit:

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
```diff
@@ -417,7 +417,9 @@ private Bundle GetBrokerOptions(AuthenticationRequest request)
                 request.Resource);
             string computedRedirectUri = GetRedirectUriForBroker();
 
-            if (!string.Equals(computedRedirectUri, request.RedirectUri, StringComparison.OrdinalIgnoreCase))
+            //Do not remove the null or empty check from the following if statement. During the silent broker flow, there is an expectation that the
+            //redirect URI will be null. Thus, disabling this check in that scenario.
+            if (!string.IsNullOrEmpty(request.RedirectUri) && !string.Equals(computedRedirectUri, request.RedirectUri, StringComparison.OrdinalIgnoreCase))
             {
```
henrik-me: this is confusing .... good to instruct devs not to remove the check. Consider making the text even simpler and possibly create a helper method which is named elegantly. 

What if the redirect uri is null, is there a bug out to fix that? is it really expected? #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
```diff
@@ -454,6 +450,22 @@ private Bundle GetBrokerOptions(AuthenticationRequest request)
             return brokerOptions;
         }
 
+        private void ValidateBrokerRedirectURI(AuthenticationRequest request)
+        {
+            string computedRedirectUri = GetRedirectUriForBroker();
+
+            //During the silent broker flow, the redirect URI will be null.
```
henrik-me, line 455: >//During the silent broker flow, the redirect URI will be null. [](start = 12, length = 63)

I assume you also have to ensure the computedUri should not be null? Which it likely isn't ... not sure if anything should be done here, but raising it as a question.

trwalke: @henrik-me im actually not sure what should happen if the computed redirect URI is null. If it is null, that could indicate that our app does not have to ability to get the redirect URI from the application. In that case, we would have no idea if the the user's redirect URI is correct. 

So the question I have is, should we stop the user's authentication request if ADAL cannot check the redirect URI?

My opinion is we should allow the request to continue because the user's redirect URI may very well be valid.

jennyf19: The computed redirectUri will return string.Empty if it's unable to determine the package signature. 

---
In reply to: [225360513](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1326#discussion_r225360513) [](ancestors = 225360513)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/BrokerProxy.cs
```diff
@@ -454,6 +450,22 @@ private Bundle GetBrokerOptions(AuthenticationRequest request)
...
+            if (string.IsNullOrEmpty(request.RedirectUri))
+            {
```
jennyf19, line 457: if you're just checking for a null redirectUri (for silent flows), then check before even going into this method. then you're not calculating the computedRedirectUri too and wasting cycles.

PR #1328, opened 2018-10-16
Fix [1277] for cross-thread exception using ownerWindow

MarkZuber:


henrik-me (approved):
:shipit:

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/WindowsFormsWebAuthenticationDialogBase.cs
```diff
@@ -351,11 +355,31 @@ protected virtual void OnAuthenticate()
...
 
+        /// <summary>
+        /// Some calls need to be made on the UI thread and this is the central place to check if we have an owner
+        /// window and if so, ensure we invoke on that proper thread.
+        /// </summary>
+        /// <param name="action"></param>
+        protected void InvokeHandlingOwnerWindow(Action action)
+        {
+            // We only support WindowsForms (since our dialog is winforms based)
+            if (ownerWindow != null && ownerWindow is Control winFormsControl)
```
jennyf19, line 366: >winFormsControl [](start = 62, length = 15)

cool

jennyf19 (approved):

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/WindowsFormsWebAuthenticationDialogBase.cs
```diff
@@ -351,11 +355,31 @@ protected virtual void OnAuthenticate()
...
         }
 
+        /// <summary>
+        /// Some calls need to be made on the UI thread and this is the central place to check if we have an owner
+        /// window and if so, ensure we invoke on that proper thread.
+        /// </summary>
+        /// <param name="action"></param>
+        protected void InvokeHandlingOwnerWindow(Action action)
+        {
+            // We only support WindowsForms (since our dialog is winforms based)
```
bgavrilMS, line 365: Will this work for a WPF application? i.e. is the ownerWindow a Control object?

MarkZuber: No, but our browser is only WinForms and the owner window can only be an IWin32Window or we fail.  We should investigate an option for embedding Edge which they now have as a WPF and WinForms plugin.

PR #1332, opened 2018-10-18
Refactor requestbase to reduce complexity

MarkZuber:
I have 2 u/p integration tests failing with "invalid credentials" but I think it may be something on my machine.  But i'd love to get feedback from @bgavrilMS and @jmprieur overnight so I can take review feedback when I get in on Thursday.

### DIFF
--- a/adal/tests/Test.ADAL.NET.Unit/DeviceCodeFlowTests.cs
+++ b/adal/tests/Test.ADAL.NET.Unit/DeviceCodeFlowTests.cs
@@ -135,14 +135,14 @@ namespace Test.ADAL.NET.Unit
 
             TokenCache cache = new TokenCache();
             AuthenticationContext ctx = new AuthenticationContext(TestConstants.DefaultAdfsAuthorityTenant, false, cache);
-            AuthenticationResult result = await ctx.AcquireTokenByDeviceCodeAsync(dcr);
+            AuthenticationResult result = await ctx.AcquireTokenByDeviceCodeAsync(dcr).ConfigureAwait(false);
             Assert.IsNotNull(result);
             Assert.AreEqual("some-access-token", result.AccessToken);
             Assert.AreEqual(AdalHttpMessageHandlerFactory.MockHandlersCount(), 0);
         }
 
         [TestMethod]
-        public async Task AdfsPostMethodTest()
+        public async Task AdfsPostMethodTestAsync()
         {
             AdalHttpMessageHandlerFactory.InitializeMockProvider();
             MockHttpMessageHandler mockMessageHandler = new MockHttpMessageHandler()
@@ -155,7 +155,7 @@ namespace Test.ADAL.NET.Unit
             AdalHttpMessageHandlerFactory.AddMockHandler(mockMessageHandler);
 
             AuthenticationContext context = new AuthenticationContext(TestConstants.DefaultAdfsAuthorityTenant, false);
-            DeviceCodeResult dcr = await context.AcquireDeviceCodeAsync(TestConstants.DefaultResource, TestConstants.DefaultClientId);
+            DeviceCodeResult dcr = await context.AcquireDeviceCodeAsync(TestConstants.DefaultResource, TestConstants.DefaultClientId).ConfigureAwait(false);
 
             Assert.IsNotNull(dcr);
             Assert.AreEqual(dcr.UserCode, "some-user-code");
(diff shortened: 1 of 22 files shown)


henrik-me (approved):
:shipit:

msal/src/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -317,9 +317,12 @@ internal Authority GetAuthority(IAccount account)
             }
 
             var handler = new SilentRequest(
-                    HttpManager, CryptographyManager, CreateRequestParameters(authority, scopes, account, UserTokenCache),
-                forceRefresh)
-            { ApiId = apiId };
```
jmprieur, line 322: the intiailization of ApiId is now missing?

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
```diff
@@ -318,7 +318,7 @@ async Task<AuthenticationResult> IConfidentialClientApplicationWithCertificate.A
             requestParameters.ExtraQueryParameters = extraQueryParameters;
 
             var handler =
-                new InteractiveRequest(HttpManager, CryptographyManager, requestParameters, null, loginHint, UIBehavior.SelectAccount, null);
+                new InteractiveRequest(HttpManager, CryptographyManager, requestParameters, ApiEvent.ApiIds.None, null, loginHint, UIBehavior.SelectAccount, null);
```
bgavrilMS, line 321: I wonder why ApiId is null / none in this case. Is it because we do not / should not have telemetry for confidential client?

adal/tests/Test.ADAL.NET.Unit/DeviceCodeFlowTests.cs
```diff
@@ -135,14 +135,14 @@ public async Task AdfsPositiveTestAsync()
...
-            AuthenticationResult result = await ctx.AcquireTokenByDeviceCodeAsync(dcr);
+            AuthenticationResult result = await ctx.AcquireTokenByDeviceCodeAsync(dcr).ConfigureAwait(false);
             Assert.IsNotNull(result);
             Assert.AreEqual("some-access-token", result.AccessToken);
             Assert.AreEqual(AdalHttpMessageHandlerFactory.MockHandlersCount(), 0);
         }
 
         [TestMethod]
-        public async Task AdfsPostMethodTest()
+        public async Task AdfsPostMethodTestAsync()
```
bgavrilMS: I fixed these just now because the GitHub build status was red.

msal/src/Microsoft.Identity.Client/ClientApplicationBase.cs
bgavrilMS: It was moved in ctor, since it is pretty much required for all flows.

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
jmprieur: no, we should have telemetry for confidential clients as well.

adal/tests/Test.ADAL.NET.Unit/DeviceCodeFlowTests.cs
jmprieur: what did you fix @bgavrilMS : I have not seen any other commit than @MarkZuber's

bgavrilMS: @jmprieur - the changes in this file were fixing a build break. This is because we now enforce static analysis on unit test projects as well.

bgavrilMS (approved):

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
MarkZuber: Only because it wasn't there before.  I kept it this way for the first round PR to explicitly raise this issue to ensure we realize the gaps in our telemetry.  I'll see if we already have a code for this and fix any of the null/none cases.

MarkZuber: Just looked.  We don't have an API ID for GetAuthorizationRequestUrlAsync() and those are the two cases in the public API where API ID is null.  I'm going to file a separate issue for that so I don't hold up this PR since it's not regressing existing functionality.

PR #1333, opened 2018-10-18
Make error codes public in MSAL

bgavrilMS:


msal/src/Microsoft.Identity.Client/Exceptions/MsalError.cs
```diff
@@ -172,5 +172,10 @@ internal static class MsalError
         /// An error response was returned by the OAuth2 server and it could not be parsed
         /// </summary>
         public const string NonParsableOAuthError = "non_parsable_oauth_error";
+
+        /// <summary>
+        /// Device code expired before contacting the server
```
jmprieur: What happens?
In the context of Device code flow (See https://aka.ms/msal-net-device-code-flow), this error happens when the device code expired before the user signed-in on another device (this is usually after 15 mins)

Mitigation:
None: inform the user that they took too long to sign-in at the provided URL and enter the provided code.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Exceptions/AdalError.cs
```diff
@@ -295,5 +295,26 @@ public static class AdalError
...
         /// </summary>
         public const string BrokerRedirectUriIncorrectFormat = "broker_redirect_uri_incorrect_format";
+
+        /// <summary>
+        /// Cannot fetch the upn from the OS, possibly due to missing permissions
+        /// </summary>
+        public const string CannotAccessUserInformationOrUserNotDomainJoined = "user_information_access_failed";
+
+        /// <summary>
+        /// Error code that denotes that this library is still polling to see if the user
```
jmprieur: This is for information (not an error per se).

henrik-me: imo. it still makes sense. Good addition to the text. perhaps it should be a remark section? and it could say Informational ?

---
In reply to: [226308298](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1333#discussion_r226308298) [](ancestors = 226308298)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/uap/WebUI.cs
```diff
@@ -38,6 +38,8 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
 {
     internal class WebUI : IWebUI
     {
+        private const string MsAppScheme = "ms-app";
```
henrik-me, line 41: >private const string MsAppScheme = "ms-app"; [](start = 8, length = 44)

why ? It seems to be in the constant class?

henrik-me (approved):
:shipit:

PR #1334, opened 2018-10-18
remove static http, it's now owned at clientapplicationbase

MarkZuber:
Now that HttpManager is fully plumbed through, we can remove the static HttpClient.  now, we have an HttpClient created per ClientApplicationBase/AuthenticationContext (via construction of a single HttpManager) that's not process wide static.

### DIFF
--- a/core/src/Http/HttpClientFactory.cs
+++ b/core/src/Http/HttpClientFactory.cs
@@ -37,12 +37,11 @@ namespace Microsoft.Identity.Core.Http
 
     internal class HttpClientFactory : IHttpClientFactory
     {
-        // as per guidelines HttpClient should be a singleton instance in an application.
-        private static HttpClient _client;
-        private static readonly object LockObj = new object();
+        // The HttpClient is a singleton per ClientApplication so that we don't have a process wide singleton.
+        private readonly HttpClient _httpClient;
         public const long MaxResponseContentBufferSizeInBytes = 1024*1024;
 
-        private static HttpClient CreateHttpClient()
+        public HttpClientFactory()
         {
             var httpClient = new HttpClient(new HttpClientHandler() { UseDefaultCredentials = true })
             {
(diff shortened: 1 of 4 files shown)


core/src/Http/HttpClientFactory.cs
```diff
@@ -37,12 +37,11 @@ internal interface IHttpClientFactory
 
     internal class HttpClientFactory : IHttpClientFactory
     {
-        // as per guidelines HttpClient should be a singleton instance in an application.
-        private static HttpClient _client;
-        private static readonly object LockObj = new object();
+        // The HttpClient is a singleton per ClientApplication so that we don't have a process wide singleton.
```
henrik-me, line 40: > The HttpClient is a singleton per ClientApplication [](start = 10, length = 52)

I'm surprised that no prod code needed an update #Closed

core/src/Http/HttpClientFactory.cs
```diff
@@ -52,28 +51,12 @@ private static HttpClient CreateHttpClient()
...
+            _httpClient = httpClient;
         }
 
-        public HttpClient GetHttpClient()
-        {
-            return GetHttpClientStatic();
-        }
-
-        public static HttpClient GetHttpClientStatic()
+        public HttpClient GetHttpClient()
```
henrik-me: >GetHttpClient() [](start = 26, length = 15)

property instead?

henrik-me (changes requested):
:clock1:

core/src/Http/HttpClientFactory.cs
henrik-me: discussed offline. 

---
In reply to: [226472465](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1334#discussion_r226472465) [](ancestors = 226472465)

core/src/Http/HttpClientFactory.cs
henrik-me: approving PR, however IMO it would be cleaner if this was changed to a property.

---
In reply to: [226472521](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1334#discussion_r226472521) [](ancestors = 226472521)

henrik-me (approved):
:shipit:

bgavrilMS (approved):

core/src/Http/HttpClientFactory.cs
MarkZuber: fixed

jennyf19 (approved):

PR #1337, opened 2018-10-19
Add support for platform specific unit tests

bgavrilMS:
This PR is very large because files were moved around etc. I will point out some areas of interest, otherwise I would not bother reviewing this way. Instead, please download the branch, have a look at the test projects and solution structures, paying special attention to the csproj files. 

A few notes: 

1. We will have one test assembly per platform. Currently there are net45 and netcore test assemblies
2. Platform tests share code (e.g. mocks, helpers) through a netstandard assembly (one for adal, one for msal)
3. I removed the core unit tests and moved them to MSAL. ADAL now includes a copy of the tests, much in the same
way that we share code from core.dll into adal.dll and msal.dll. I did this for 2 reasons: 

a. By putting code into shared assemblies, I started having the problem that Petro is having in his PR, i.e. type conflicts. 
For example the MockHttpRequest is not of type IHttpRequest because the types are from different assemblies. 

b. This structure is closer to the code that we ship, i.e. we do not ship core.dll. We can remove core.dll in a 
similar manner, by putting all the files into MSAL and "referencing" them from ADAL.

4. As a coroloary to #3, we now run core unit tests twice - once part of ADAL, once part of MSAL. Not a big deal 
since they run in a few seconds.


MarkZuber:
Questions (requires some setup so bear with me)…

* msal/tests/Test.MSAL.NET.Unit.net45 directory has a Test.MSAL.NET.Unit.net45.csproj
* Many of the tests in this directory are valid for netstandard/ios/android/etc.

Are you planning on having a Test.MSAL.NET.Unit.ios.csproj in the same directory so that it picks up the same source?  If so, that's weird since the directory has .net45 in the name

Is that directory eventually only for the .NET Framework specific code and we'll move the standard code out somewhere else for testing that's "cross platform"?  

Or are we only going to test the cross plat stuff once within NET45?  (I don't recommend this since I think we should run tests on each of our platforms fully and I believe that's what this PR is starting)

Along these lines:
Do we want to split up the <targetframeworks> stuff in the msal/adal directories into separate csproj files as well?  For example, if I want to work on this code on a mac or a Linux box, I can't because .NET Framework (net45 target) isn't supported on them so I have to manually edit the csproj to update the targets.  It would be cool to (eventually) be able to load just the ios stuff and code on a mac, or just the .net core stuff and code on a Linux box (or even mac or windows if i'm just targeting netcore scenarios or just don't want to wait to build ios/android every time until i'm ready to test).

</longwinded>

bgavrilMS:
Nice, thank you checking this out in such detail! 

The unit test projects are created against a single framework, i.e. the net45 one will test the net45 version of MSAL, the netcore one will test the netcore version of MSAL etc. I agree that we could have a unit test project that targets multiple frameworks, matching MSAL but there are a few problems with this approach: 

a) Visual Studio has very poor support for multi-framework unit tests. It just runs the under the first framework. So you could break an iOS test and not realize it because VS runs only for net45. Note that our CI build will run the tests correctly though, because it uses the command line. 
b) Existing tests are running against net45. Making the tests run under all platforms would not be trivial - missing features on some platforms etc. We'd also have to recreate the structure of the product assemblies with all the flags etc. 
c) If we have a block of product code that does not contain conditional compilation, do we gain anything by testing it under 5 platforms? I'd argue that .net ensures this.
d) In the future we could add a unit test project that runs for all platforms and move some tests there. I see this as a separate task. Today we have the net45 tests and a single netcore test. 


As for splitting ADAL / MSAL in directories - I think the initial multi-plat version of the libraries was like this and it wasn't scaling well. Maybe @henrinkm knows more. 

---
In reply to: [431450306](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1337#issuecomment-431450306) [](ancestors = 431450306)

PR #1347, opened 2018-10-24
Jennyf/custom tabs

jennyf19:
## Devices and browsers tested

| | Browser*     |  Result  | 
| ------------- |:-------------:|:-----:|
| Huawei/One+ | Edge* | Pass
| Huawei/One+ | Firefox* | Pass
| Huawei/One+ | Brave* | Pass
| One+ | Ecosia* | Pass
| One+ | Kiwi* | Pass
| Huawei/One+ | Opera | Pass
| Huawei | OperaMini | Pass
| Huawei/One+ | InBrowser | Pass
| One+ | Maxthon | Pass
| Huawei/One+ | DuckDuckGo | User cancelled auth
| Huawei/One+ | UC Browser | User cancelled auth
| One+ | Dolphin | User cancelled auth
| One+ | CM browser | User cancelled auth
| Hauwei/One+ | none installed | AndroidActivityNotFound ex
---
> Supports Custom Tabs*

**Tested helper method** w/no browsers, w/Chrome disabled, w/non-custom tabs browsers & w/custom tabs browsers

**TODO** Update wiki with the new browser information. I have a possible aka.ms link  that I will use to link to the wiki. Was going to wait until after the PR is merged.

If you want to test w/any of the devices, lkm.


core/src/Exceptions/CoreErrorMessages.cs
```diff
@@ -51,7 +51,8 @@ internal class CoreErrorMessages
         public const string UnsupportedAuthorityValidation =
             "Authority validation is not supported for this type of authority";
 
-        public const string AuthenticationCanceled = "User canceled authentication";
+        public const string AuthenticationCanceled = "User canceled authentication. On an Android device, this could be due to the system browser." +
+            " See https://aka.ms/msal-net-system-browsers for more information.";
```
jennyf19, line 56: On a few of the browsers that do not support custom tabs, after entering creds, we do not get redirected back to the app, and the user needs to leave the app and then go back to it. when that happens, a User Canceled Auth exception is returned, so I thought to add a bit to the error message to handle that case. Not sure where else to put it, as the app is basically frozen, but wanted to keep the message short, as it's android specific. thoughts? #Closed

core/src/Platforms/Android/SystemWebview/AuthenticationActivity.cs
```diff
@@ -119,39 +117,36 @@ protected override void OnResume()
...
                     StartActivity(browserIntent);
                 }
                 catch (ActivityNotFoundException ex)
                 {
-                    const string chromeDisabledMessage = "Chrome is disabled on the device, cannot proceed with authentication ";
                     throw CoreExceptionFactory.Instance.GetClientException(
-                        CoreErrorCodes.ChromeDisabledError, 
-                        chromeDisabledMessage + ex.InnerException);
+                           CoreErrorCodes.AndroidActivityNotFound,
+                           CoreErrorMessages.AndroidActivityNotFound, ex);
```
jennyf19, line 140: If the user does not have any browsers installed OR chrome is disabled on the device, we will throw here.  #Closed

msal/src/Microsoft.Identity.Client/UIParent.cs
```diff
@@ -116,20 +118,24 @@ public static bool IsSystemWebviewAvailable()
...
-                    packageManager.GetPackageInfo(_chromePackages[i], PackageInfoFlags.Activities);
+                    customTabPackages.Add(packageManager.GetPackageInfo(_customTabPackages[i], PackageInfoFlags.Activities).ToString());
                     counter++;
                 }
                 catch (PackageManager.NameNotFoundException)
                 {
                 }
             }
-            if (counter > 0 && applicationInfo.Enabled == true)
+            if (counter == 1 && !string.IsNullOrEmpty(customTabPackages[0]) && !applicationInfo.Enabled)
```
jennyf19: If you guys have a better suggestion for this, let me know. :)
Basically, Chrome can not be uninstalled (well, it's really complicated), so most users will just disable it, but the package will still show up, but with `applicationInfo.Enabled` false, so we need to check that `com.android.chrome` is disabled, otherwise, we'll return a count of 1, continue w/system browser and it will fail.  #Closed

jennyf19: Background info if you want it:
Native android is making the decision for the dev to switch to embedded webview if no browsers are available. Due to security issues (ex. [Google auth](https://auth0.com/blog/google-blocks-oauth-requests-from-embedded-browsers/) does not accept embedded webview), we (.net) decided to let the dev choose instead of the library, which is what this helper function is for. If the dev doesn't care if embedded or system browser is used, they can run this method to check if the device can handle system browser, and if not, launch embedded, but this is the dev devision. The experience is invisible to the user. #Closed

core/src/Exceptions/CoreErrorMessages.cs
```diff
@@ -51,7 +51,8 @@ internal class CoreErrorMessages
         public const string UnsupportedAuthorityValidation =
             "Authority validation is not supported for this type of authority";
 
-        public const string AuthenticationCanceled = "User canceled authentication";
+        public const string AuthenticationCanceled = "User canceled authentication. On an Android device, this could be due to the system browser." +
```
henrik-me: >due to the system browser. [](start = 120, length = 26)

suggestion: due to the lack of capabilities such as custom tabs for the system browser. #Closed

core/src/Exceptions/CoreErrorMessages.cs
henrik-me: Good enough as long as the aka.ms link is there and it shows details on capabilities the browser needs to have and some examples of browsers you have validated works and browsers you have seen fail. (I suggest simply re-using the table you have here)

---
In reply to: [227610125](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227610125) [](ancestors = 227610125)

msal/src/Microsoft.Identity.Client/UIParent.cs
```diff
@@ -116,20 +118,24 @@ public static bool IsSystemWebviewAvailable()
...
+            {
+                return false;
+            }
+            else if (counter > 0)
             {
                 return true;
             }
```
henrik-me, line 125: see comment above #Closed

msal/src/Microsoft.Identity.Client/Exceptions/MsalClientException.cs
```diff
@@ -154,28 +154,17 @@ public class MsalClientException : MsalException
...
-        /// Xamarin.Android specific. This error indicates that chrome, or a browser implementing chrome tabs, is not installed on the device. 
-        /// The library sdk uses chrome custom tab for authorize request if applicable or falls back to chrome browser.
-        /// <para>Mitigation</para>If you really need to use the System web browser (for instance to get SSO with the browser), notify the end 
-        /// user that chrome or a browser implementing chrome custom tabs needs to be installed on the device. 
-        /// Otherwise you can also use <see cref="UIParent.IsSystemWebviewAvailable"/> to check if a required browser is available on the device
-        /// and require the library to use the embedded web view if it is not by setting the boolean to <c>true</c> in the following
-        /// constructor: <see cref="UIParent.UIParent(Android.App.Activity, bool)"/>
-        /// <para>For more details</para> See https://aka.ms/msal-net-uses-web-browser
-        /// </summary>
-        public const string ChromeNotInstalledError = "chrome_not_installed";
```
henrik-me: >ChromeNotInstalledError [](start = 28, length = 23)

if someone is using these.... perhaps those should stay or alternatively we should have information saying those are deprecated? then remove them later? (same for the one below) #Resolved

msal/src/Microsoft.Identity.Client/UIParent.cs
henrik-me: I suppose this is valid also if the counter is larger than 1 ? Thus it has to do this validation for the browser used

---
In reply to: [227610888](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227610888) [](ancestors = 227610888)

henrik-me (changes requested):
:clock1:

msal/src/Microsoft.Identity.Client/Exceptions/MsalClientException.cs
jmprieur: Yes, that's a breaking change in the public API. we cannot remove it. you might want to add and [Obsolete] attribute with the error property set to true. In case people use it today, they will know they need to use `AndroidActivityNotFound` instead. #Closed

jmprieur (changes requested):
Suggesting to add Obsolete attributes instead of removing error strings from the public API

core/src/Platforms/Android/SystemWebview/AuthenticationActivity.cs
```diff
@@ -119,39 +117,36 @@ protected override void OnResume()
...
-                        chromeNotInstalledMessage);
-                }
-
-                Intent browserIntent = new Intent(Intent.ActionView);
-                browserIntent.SetData(Uri.Parse(_requestUrl));
-                browserIntent.SetPackage(chromePackage);
+                Intent browserIntent = new Intent(Intent.ActionView, Uri.Parse(_requestUrl));
                 browserIntent.AddCategory(Intent.CategoryBrowsable);
+
+                CoreLoggerBase.Default.Info(
```
bgavrilMS: >Info [](start = 39, length = 4)

Should this be a warning? #Resolved

core/src/Platforms/Android/SystemWebview/AuthenticationActivity.cs
```diff
@@ -212,54 +207,26 @@ private string GetChromePackageWithCustomTabSupport(Context context)
             }
 
             Intent customTabServiceIntent = new Intent(_customTabsServiceAction);
-            IList<ResolveInfo> resolveInfoList = context.PackageManager.QueryIntentServices(
-                    customTabServiceIntent, 0);
+
+            IList<ResolveInfo> resolveInfoListWithCustomTabs = context.PackageManager.QueryIntentServices(
```
bgavrilMS: >IList [](start = 12, length = 5)

Prefer IEnumerable as it is readonly by default. IList allows you to change the collection, which you do not need / want. Your code also becomes easier to read (i.e. if I see an IList somehwere, I wonder "why do we need to change this collection?").. #Closed

core/src/Platforms/Android/SystemWebview/AuthenticationActivity.cs
```diff
@@ -212,54 +207,26 @@ private string GetChromePackageWithCustomTabSupport(Context context)
...
+                    customTabServiceIntent, PackageInfoFlags.MatchAll);
 
             // queryIntentServices could return null or an empty list if no matching service existed.
-            if (resolveInfoList == null || resolveInfoList.Count == 0)
+            if (resolveInfoListWithCustomTabs == null || resolveInfoListWithCustomTabs.Count == 0)
```
bgavrilMS: >Count [](start = 87, length = 5)

nit: collection.Any() instead of Count #Closed

msal/src/Microsoft.Identity.Client/UIParent.cs
```diff
@@ -116,20 +118,24 @@ public static bool IsSystemWebviewAvailable()
 
             int counter = 0;
 
-            ApplicationInfo applicationInfo = Application.Context.PackageManager.GetApplicationInfo(_chromePackages[0], 0);
-
-            for (int i = 0; i < _chromePackages.Length; i++)
+            ApplicationInfo applicationInfo = Application.Context.PackageManager.GetApplicationInfo(_customTabPackages[0], 0);
```
bgavrilMS: >0 [](start = 119, length = 1)

if chrome has special meaning, I would pull it out in a separate constant. You can still keep it in _customTabPacakges. #Closed

msal/src/Microsoft.Identity.Client/UIParent.cs
bgavrilMS: You should add this line about Chrome as a comment in the code.

---
In reply to: [227610888](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227610888) [](ancestors = 227610888)

bgavrilMS: >counter [](start = 16, length = 7)

I think you can simplify this logic if you pull out the chrome stuff separtely, maybe not have any counter variable at all. #Closed

bgavrilMS (approved):
:shipit:

core/src/Exceptions/CoreErrorMessages.cs
jennyf19: I was going to add the table and maybe something else (a cat pic or two). 

---
In reply to: [227648339](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227648339) [](ancestors = 227648339,227610125)

msal/src/Microsoft.Identity.Client/UIParent.cs
jennyf19: I think i understand your question. If the counter is larger than 1, it doesn't matter because it means there is a package w/custom tabs available. so even if chrome is disabled, there is something else, and that package will be used (not chrome). does that answer your question?

---
In reply to: [227649549](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227649549) [](ancestors = 227649549,227610888)

core/src/Platforms/Android/SystemWebview/AuthenticationActivity.cs
jennyf19: great suggestion. fixed.

---
In reply to: [227735960](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227735960) [](ancestors = 227735960)

core/src/Platforms/Android/SystemWebview/AuthenticationActivity.cs
jennyf19: done.

---
In reply to: [227736307](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227736307) [](ancestors = 227736307)

msal/src/Microsoft.Identity.Client/UIParent.cs
jennyf19: i tried this. let me know what you think. 

---
In reply to: [227737484](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227737484) [](ancestors = 227737484)

msal/src/Microsoft.Identity.Client/Exceptions/MsalClientException.cs
jennyf19: I added them back and added obsolete. 

---
In reply to: [227689910](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227689910) [](ancestors = 227689910)

jennyf19: added them back and added obsolete

---
In reply to: [227649302](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227649302) [](ancestors = 227649302)

core/src/Platforms/Android/SystemWebview/AuthenticationActivity.cs
jennyf19: done.

---
In reply to: [227734902](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227734902) [](ancestors = 227734902)

bgavrilMS (approved):

jmprieur (approved):

msal/src/Microsoft.Identity.Client/UIParent.cs
```diff
@@ -73,8 +74,11 @@ public UIParent(bool useEmbeddedWebview) : this()
 #if ANDROID
         private Activity Activity { get; set; }
 
-        private static readonly string[] _chromePackages =
-        {"com.android.chrome", "com.chrome.beta", "com.chrome.dev"};
+        private static readonly string ChromePackage = "com.android.chrome";
+
+        private static readonly string[] _customTabPackages =
+        { ChromePackage, "com.chrome.beta", "com.chrome.dev", "com.microsoft.emmx", "org.mozilla.firefox",
+        "com.ecosia.android", "com.kiwibrowser.browser", "com.brave.browser"};
```
henrik-me: can't we ask the browser itself if it supports custom tabs? Do we really need to maintain this list out selves? I would not like to be required to ship an update to the library to support a new browser #Resolved

jennyf19: i'll look into it.

---
In reply to: [227919485](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227919485) [](ancestors = 227919485)

henrik-me: great. I'll approve this PR and you can investigate and add that particular change in a later PR. Thanks!

---
In reply to: [227920771](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227920771) [](ancestors = 227920771,227919485)

henrik-me (approved):
:shipit:

msal/src/Microsoft.Identity.Client/UIParent.cs
jennyf19: ok...pushed up some changes. lmk what you think.

---
In reply to: [227919485](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227919485) [](ancestors = 227919485)

msal/src/Microsoft.Identity.Client/UIParent.cs
```diff
@@ -138,6 +128,37 @@ public static bool IsSystemWebviewAvailable()
                 return false;
             }
         }
+
+        private static bool IsBrowserWithCustomTabSupportAvailable()
```
henrik-me, line 132: >IsBrowserWithCustomTabSupportAvailable [](start = 28, length = 38)

Did you validate this for all the cases you mention in your PR? #Closed

msal/src/Microsoft.Identity.Client/UIParent.cs
henrik-me: Looks much better to me :)

---
In reply to: [227935640](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227935640) [](ancestors = 227935640,227919485)

msal/src/Microsoft.Identity.Client/UIParent.cs
```diff
@@ -112,24 +114,12 @@ public UIParent(Activity activity, bool useEmbeddedWebview) : this(activity)
         /// </example>
         public static bool IsSystemWebviewAvailable()
```
henrik-me, line 115: >IsSystemWebviewAvailable [](start = 27, length = 24)

I believe your testing also found that even without custom tabs there are some browsers where getting the token still works? E.g Opera. I suppose we should still allow those through? #Closed

msal/src/Microsoft.Identity.Client/UIParent.cs
jennyf19: not all...i did: chrome disabled, chrome enabled, chrome disabled w/firefox, chrome disabled w/firefox and edge

---
In reply to: [227939830](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227939830) [](ancestors = 227939830)

henrik-me: ok. 

---
In reply to: [227941563](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227941563) [](ancestors = 227941563,227939830)

msal/src/Microsoft.Identity.Client/UIParent.cs
jennyf19: but then we are back to list. we know for sure the ones w/custom tabs works. not sure why some other browsers do work and others do not. Up for discussing this further. 

---
In reply to: [227940590](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227940590) [](ancestors = 227940590)

jennyf19: we could also only allow browsers w/custom tabs support, which covers most of the popular android ones, except for a Opera and Dolphin (which doesn't work). 

---
In reply to: [227940590](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1347#discussion_r227940590) [](ancestors = 227940590)

henrik-me (approved):
:shipit:

PR #1350, opened 2018-10-24
cache format tests + format  changes 

SomkaPe:
Cache format tests

While writing cache format test some cache format inconsistencies with current version of unified schema were found:
1. Timestamps should be stored as strings in Json
2. It was decided to not store client_info(which is optional field) in all cache entities  - just in account
3. Credential Type field should be Case sensitive
4. Keys of cache entities should be in lower case (including iOS key chain key)
5. Scopes should not be stored in sorted form, instead should be  stored as is (as returned from STS)
6. Change "preferred_username not in id_token" to "Missing from the token response"
7. Changing iOS key component "Generic" for account cache entity - store "username" instead of "local_account_id" (requested by Olga)

Others:

1. Persisting given_name and family_name optional claims in account cache entity
2. Fixing issue of client_info misuse:
  2.1. AccounId is derived from ClientInfo (Not opposite way - ClientInfo can not be recreated from accountId, it can contains more fields )
  2.2. Some code was written based on assumption that client info is always persisted with cache entities, which is wrong because it is optional field, so some library can not persist it.


core/src/Cache/MsalCacheCommon.cs
```diff
@@ -38,9 +38,9 @@ internal class MsalCacheCommon
         public const string ScopesDelimiter = " ";
         public const string CacheKeyDelimiter = "-";
 
-        public const string IdToken = "idtoken";
-        public const string AccessToken = "accesstoken";
-        public const string RefreshToken = "refreshtoken";
+        public const string IdToken = "IdToken";
+        public const string AccessToken = "AccessToken";
+        public const string RefreshToken = "RefreshToken";
```
henrik-me, line 43: not sure I understand why you are CamelCasing these. Though I see that the methods lowercases everything. #Pending

core/src/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -60,8 +60,8 @@ internal MsalAccessTokenCacheItem
             NormalizedScopes = scopes;
             TenantId = tenantId;
             Secret = secret;
-            ExpiresOnUnixTimestamp = CoreHelpers.DateTimeToUnixTimestamp(accessTokenExpiresOn);
-            CachedAt = CoreHelpers.CurrDateTimeInUnixTimestamp();
+            ExpiresOnUnixTimestamp = CoreHelpers.DateTimeToUnixTimestamp(accessTokenExpiresOn).ToString();
```
henrik-me: Would it make sense to make the DateTimeToUnixTimestamp return as string or perhaps to add another helper e.g. DateTimeToUnixTimestampAsString ensuring people do not have to rememeber to add .ToString() ? #Closed

core/src/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -45,7 +45,7 @@ internal MsalAccessTokenCacheItem()
         internal MsalAccessTokenCacheItem
             (string environment, string clientId, MsalTokenResponse response, string tenantId) :
 
-            this(environment, clientId, response.TokenType, ScopeHelper.ConvertStringToLowercaseSortedSet(response.Scope).AsSingleString(),
+            this(environment, clientId, response.TokenType, response.Scope,
```
henrik-me, line 48: >response.Scope [](start = 60, length = 14)

Does this handle sorting and normalizing? I believe that is a requirement we wanted to ensure happened for the cache? #Pending

msal/tests/dev apps/DesktopTestApp/MsalUserAccessTokenControl.cs
```diff
@@ -35,7 +35,7 @@ public MsalUserAccessTokenControl()
         private void expireAccessTokenButton_Click(object sender, System.EventArgs e)
         {
             expiresOnLabel.Text = DateTimeOffset.UtcNow.ToString();
-            _item.ExpiresOnUnixTimestamp = CoreHelpers.DateTimeToUnixTimestamp(DateTimeOffset.UtcNow);
+            _item.ExpiresOnUnixTimestamp = CoreHelpers.DateTimeToUnixTimestamp(DateTimeOffset.UtcNow).ToString();
```
henrik-me: >.ToString(); [](start = 101, length = 12)

made a comment elsewhere on this

msal/tests/dev apps/XForms/XForms/AccessTokenCacheItemDetails.xaml.cs
```diff
@@ -52,7 +53,7 @@ public partial class AccessTokenCacheItemDetails : ContentPage
             expiresOnLabel.Text = msalAccessTokenCacheItem.ExpiresOn.ToString();
             scopesLabel.Text = msalAccessTokenCacheItem.NormalizedScopes;
 
-            cachedAtLabel.Text = CoreHelpers.UnixTimestampToDateTime(msalAccessTokenCacheItem.CachedAt).ToString();
+            cachedAtLabel.Text = CoreHelpers.UnixTimestampToDateTime(Convert.ToInt64(msalAccessTokenCacheItem.CachedAt)).ToString();
```
henrik-me: >Convert.ToInt64( [](start = 69, length = 16)

can this be moved to the MsalAccessTokenCacheItem class?  #Closed

henrik-me:
Thanks a lot for looking into this Petro! #Closed

henrik-me:
> {

for my education, are these test data the same across all libraries?

---
Refers to: msal/tests/Test.MSAL.NET.Unit/Resources/AADTestData.txt:1 in cc222be. [](commit_id = cc222beb1099b5f150b5083a1ce40d1a6197603b, deletion_comment = False)

henrik-me (changes requested):
:clock1:

core/src/IdToken.cs
```diff
@@ -72,6 +74,12 @@ internal class IdToken
         [DataMember(Name = IdTokenClaim.HomeObjectId, IsRequired = false)]
         public string HomeObjectId { get; set; }
 
+        [DataMember(Name = IdTokenClaim.GivenName)]
```
jennyf19, line 77: this one doesn't need IsRequired = false?

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -56,7 +56,7 @@ namespace Microsoft.Identity.Client
     public sealed class TokenCache
 #pragma warning restore CS1574 // XML comment has cref attribute that could not be resolved
     {
-        internal const string NullPreferredUsernameDisplayLabel = "preferred_username not in id_token";
+        internal const string NullPreferredUsernameDisplayLabel = "Missing from the token response";
```
jennyf19, line 59: >Missing from the token response [](start = 67, length = 31)

much better wording #Closed

core/src/Cache/MsalCacheCommon.cs
SomkaPe: According to current schema design doc https://identitydivision.visualstudio.com/DevEx/DevEx%20Team/_git/AuthLibrariesApiReview?path=%2FUnifiedSchema%2FSchema.md&version=GBoldalton%2Funified_schema_transfer
credential_type is case sensitive, but all keys are lowercased

core/src/Cache/Items/MsalAccessTokenCacheItem.cs
SomkaPe: sure, will change  #Closed

core/src/Cache/Items/MsalAccessTokenCacheItem.cs
SomkaPe: according to schema scopes should be stored as returned from sts

msal/tests/dev apps/XForms/XForms/AccessTokenCacheItemDetails.xaml.cs
SomkaPe: yes, will change #Closed

core/src/IdToken.cs
SomkaPe: it is optional, which means that other sdk might not persist it #ByDesign

msal/src/Microsoft.Identity.Client/TokenCache.cs
SomkaPe: ) #Closed

core/src/IdToken.cs
jennyf19: I see. thanks.

---
In reply to: [227982217](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1350#discussion_r227982217) [](ancestors = 227982217)

SomkaPe:
> > {
> 
> for my education, are these test data the same across all libraries?
> 
> Refers to: msal/tests/Test.MSAL.NET.Unit/Resources/AADTestData.txt:1 in [cc222be](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/commit/cc222beb1099b5f150b5083a1ce40d1a6197603b). [](commit_id = [cc222be](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/commit/cc222beb1099b5f150b5083a1ce40d1a6197603b), deletion_comment = False)

Yes Test cases can be found at VSTS at AuthLibrariesApiReview /UnifiedSchema/testcases #Closed

core/src/Cache/MsalCacheCommon.cs
henrik-me: ok... not sure though why the values here are CamelCased (if it's being lower cased later anyway).

---
In reply to: [227980008](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1350#discussion_r227980008) [](ancestors = 227980008)

core/src/Cache/Items/MsalAccessTokenCacheItem.cs
henrik-me: great, that was not how I understood it. Are we sure the server is sorting etc.?

---
In reply to: [227981415](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1350#discussion_r227981415) [](ancestors = 227981415)

core/src/Cache/Keys/MsalAccountCacheKey.cs
```diff
@@ -76,17 +76,17 @@ public string GetiOSAccountKey()
...
             stringBuilder.Append(_environment);
 
-            return stringBuilder.ToString();
+            return stringBuilder.ToString().ToLower();
         }
 
         public string GetiOSServiceKey()
         {
-            return _tenantId ?? "";
+            return (_tenantId ?? "").ToLower();
```
henrik-me, line 84: >     return (_tenantId ?? "").ToLower(); [](start = 7, length = 40)

I'm speculating, however wouldn't this righ backwards compatibility? #Closed

core/src/Cache/MsalCacheCommon.cs
SomkaPe: same values stored in cache entity in Camel Case but in corresponding key in lower case (it was agreed to have lower cased keys for all cache entities)

core/src/Cache/Items/MsalAccessTokenCacheItem.cs
SomkaPe: no, it does not , we should do it during cache lookup logic

core/src/Cache/Keys/MsalAccountCacheKey.cs
SomkaPe: we do not access cache by individual key , we do it by cache entity type - like read all access tokens,
so this is not a problem #Closed

msal/tests/Test.MSAL.NET.Unit.net45/UnifiedCacheFormatTests.cs
```diff
@@ -0,0 +1,308 @@
+﻿using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Internal;
```
jennyf19, line 29: Missing the header info

core/src/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -78,12 +78,12 @@ internal MsalAccessTokenCacheItem
...
         [DataMember(Name = "cached_at", IsRequired = true)]
-        internal long CachedAt { get; set; }
+        internal string CachedAt { get; set; }
 
         [DataMember(Name = "expires_on", IsRequired = true)]
-        internal long ExpiresOnUnixTimestamp { get; set; }
+        internal string ExpiresOnUnixTimestamp { get; set; }
 
-        [DataMember(Name = "user_assertion_hash")]
+        [DataMember(Name = "user_assertion_hash", EmitDefaultValue = false)]
```
jennyf19, line 86: >EmitDefaultValue = false [](start = 50, length = 24)

what's this for?

core/src/Cache/Items/MsalCacheItemBase.cs
```diff
@@ -41,7 +41,6 @@ internal abstract class MsalCacheItemBase
         [DataMember(Name = "environment", IsRequired = true)]
         internal string Environment { get; set; }
 
-        [DataMember(Name = "client_info")]
```
jennyf19, line 44: remove this as a datamember because it's not required...correct? #Closed

msal/tests/Test.MSAL.NET.Unit.net45/UnifiedCacheFormatTests.cs
```diff
@@ -0,0 +1,308 @@
...
+                var propName = prop.Key;
+                var expectedPropValue = prop.Value;
+                var actualPropValue = actualPayload.GetValue(propName);
+
+                Assert.AreEqual(expectedPropValue, actualPropValue);
+            }
+        }
+    }
+}
+#endif
```
jennyf19, line 335: this is pretty cool.  What about ADFS scenarios?

core/src/Cache/Items/MsalAccountCacheItem.cs
```diff
@@ -90,9 +94,19 @@ internal MsalAccountCacheItem(string environment, MsalTokenResponse response, st
...
+            set
+            {
+                RawClientInfo = value;
+            }
+        }
+
         internal MsalAccountCacheKey GetKey()
         {
-            return new MsalAccountCacheKey(Environment, TenantId, HomeAccountId, LocalAccountId);
+            return new MsalAccountCacheKey(Environment, TenantId, HomeAccountId, PreferredUsername);
```
jennyf19, line 109: >PreferredUsername [](start = 81, length = 17)

Why is this changing to PreferredUsername? Will that impact existing cache (pre-next release)?

core/src/Cache/Keys/MsalAccountCacheKey.cs
```diff
@@ -39,9 +39,9 @@ internal class MsalAccountCacheKey
         private readonly string _environment;
         private readonly string _homeAccountId;
         private readonly string _tenantId;
-        private readonly string _localAccountId;
+        private readonly string _username;
```
jennyf19, line 42: >_username [](start = 32, length = 9)

why the switch to _username? #Closed

msal/tests/Test.MSAL.NET.Unit.net45/UnifiedCacheFormatTests.cs
SomkaPe: sure, will add

core/src/Cache/Items/MsalAccessTokenCacheItem.cs
SomkaPe: is used to determine whether to serialize the default value for a field or property being serialized

core/src/Cache/Items/MsalCacheItemBase.cs
SomkaPe: it was decided to persist it only with account object #Closed

msal/tests/Test.MSAL.NET.Unit.net45/UnifiedCacheFormatTests.cs
SomkaPe: can be also included quite easily, main challenge is to create test data for adfs -> sts response and expectations , it probably should be done in separate PR related to enabling ADFS in msal

core/src/Cache/Items/MsalAccountCacheItem.cs
SomkaPe: this change was requested by iOS team, purpose is to optimize cache lookups, 
it will not affect .NET Msal lookup because it does not perform lookup on individual keys

core/src/Cache/Keys/MsalAccountCacheKey.cs
SomkaPe: this field is used only for persistence on iOS - as value for "Generic", iOS team requested this change   
 to optimize cache lookup, from .Net side we need to change it to be compatible with iOS Msal #Closed

msal/tests/dev apps/XForms/XForms/CachePage.xaml.cs
```diff
@@ -111,7 +111,7 @@ public void OnExpire(object sender, EventArgs e)
             var tokenCache = App.MsalPublicClient.UserTokenCache;
 
             // set access token as expired
-            accessTokenCacheItem.ExpiresOnUnixTimestamp = GetCurrentTimestamp();
+            accessTokenCacheItem.ExpiresOnUnixTimestamp = GetCurrentTimestamp().ToString();
```
henrik-me: >GetCurrentTimestamp [](start = 58, length = 19)

nit: consider making the method return a string

core/src/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -61,7 +61,7 @@ internal MsalAccessTokenCacheItem
             TenantId = tenantId;
             Secret = secret;
             ExpiresOnUnixTimestamp = CoreHelpers.DateTimeToUnixTimestamp(accessTokenExpiresOn);
-            CachedAt = CoreHelpers.CurrDateTimeInUnixTimestamp();
+            CachedAt = CoreHelpers.CurrDateTimeInUnixTimestamp().ToString();
```
henrik-me: >CurrDateTimeInUnixTimestamp(). [](start = 35, length = 30)

is this used anywhere else where it's supposed to return a long or can the method be updated to return a string?

henrik-me (approved):
:shipit:

msal/tests/dev apps/XForms/XForms/CachePage.xaml.cs
SomkaPe: changed

core/src/Cache/Items/MsalAccessTokenCacheItem.cs
SomkaPe: changed

PR #1352, opened 2018-10-25
Sagonzal/updatetelemetrypart1

sangonzal:
Waiting to hear back from Aria team regarding issues with ingesting events for certain platforms, so split PR into two:

1) This PR: getting MSAL telemetry on par with other platforms
2) Coming soon: adding telemetry to all dev apps and uploading events to aria


core/src/IPlatformProxy.cs
```diff
@@ -35,6 +35,24 @@ internal interface IPlatformProxy
 
         Task<bool> IsUserLocalAsync(RequestContext requestContext);
 
+        /// <summary>
+        /// Returns the name of the calling assembly
+        /// </summary>
+        /// <returns></returns>
+        string GetApplicationName();
```
bgavrilMS: I would name them as GetCallingAseemblyName and GetCallingAssemblyVersion to differentiate better with the GetLibraryName etc. #Closed

core/src/OAuth2/OAuth2Client.cs
```diff
@@ -105,6 +105,28 @@ internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, R
 
                 httpEvent.HttpResponseStatus = (int)response.StatusCode;
                 httpEvent.UserAgent = response.UserAgent;
+                
+                httpEvent.HttpMethod = method.Method.ToString();
+
+                if(response.HeadersAsDictionary.ContainsKey("x-ms-request-id") &&
```
bgavrilMS: HeadersAsDictioary copies the headers into a dictionary. Not an very expensive operation, but since it's on the critical path, could you store this dictionary as a local method variable and reuse it? #Closed

core/src/OAuth2/OAuth2Client.cs
```diff
@@ -105,6 +105,28 @@ internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, R
...
+                if(response.HeadersAsDictionary.ContainsKey("x-ms-clitelem") && 
+                    response.HeadersAsDictionary["x-ms-clitelem"] != null)
+                {
+                    XMsTelemetryInfo xMsTeleminfo = XMsTelemInfoHelper.parseXMsTelemHeader(response.HeadersAsDictionary["x-ms-clitelem"], requestContext);
+                    if (xMsTeleminfo != null)
+                    {
+                        httpEvent.TokenAge = xMsTeleminfo.TokenAge;
+                        httpEvent.SpeInfo = xMsTeleminfo.SpeInfo;
+                        httpEvent.ServerErrorCode = xMsTeleminfo.ServerErrorCode;
+                        httpEvent.ServerErrorCode = xMsTeleminfo.ServerSubErrorCode;
```
bgavrilMS: ServerSubErrorCode? #Closed

core/src/Platforms/iOS/iOSPlatformProxy.cs
```diff
@@ -120,6 +122,20 @@ public string GetProductName()
         {
             return _isMsal ? "MSAL.Xamarin.iOS" : "PCL.iOS";
         }
+        public string GetApplicationName()
+        {
+            return (NSString)NSBundle.MainBundle.InfoDictionary["CFBundleName"].ToString();
```
bgavrilMS: Are all of these app name / app version strings guaranteed to have a value / not throw a null ref? Do you want to have a fallback ? #Closed

core/src/Platforms/net45/NetDesktopPlatformProxy.cs
```diff
@@ -204,6 +207,22 @@ public string GetProductName()
...
+
+        public string GetApplicationVersion()
+        {
+            return Assembly.GetEntryAssembly().GetName().Version.ToString();
+        }
+
+        public string GetDeviceId()
+        {
+            return NetworkInterface.GetAllNetworkInterfaces().Where(nic => nic.OperationalStatus == OperationalStatus.Up)
+                .Select(nic => nic.GetPhysicalAddress().ToString()).FirstOrDefault();
```
bgavrilMS: Isn't this PII (in which case you might want to hash it...) ? And as above, is this guaranteed to return a value? Do you want to have a fallback? #Closed

core/src/Platforms/netcore/NetCorePlatformProxy.cs
```diff
@@ -108,6 +109,21 @@ public string GetProductName()
...
+        }
+
+        public string GetApplicationVersion()
+        {
+            return Assembly.GetEntryAssembly().GetName().Version.ToString();
+        }
+
+        public string GetDeviceId()
+        {
+            return null;
```
bgavrilMS: Do you want to use Enviroment.MachineName or a hash of that ? #Closed

core/src/Platforms/Android/AndroidPlatformProxy.cs
```diff
@@ -134,6 +134,20 @@ public string GetProductName()
         {
             return _isMsal ? "MSAL.Xamarin.Android" : "PCL.Android";
         }
+        public string GetApplicationName()
```
bgavrilMS: Please add some comments to these properties (here or in the interface) that these contain PII and should not be used elsewhere in the code expect for the purposes of telemetry, which implies the user's consent. I'd even add PII in the method names. #Closed

core/src/Telemetry/DefaultEvent.cs
```diff
@@ -25,18 +25,35 @@
...
     {
-        public DefaultEvent(string clientId) : base(EventNamePrefix + "default_event")
+        public DefaultEvent(string clientId, ConcurrentDictionary<string, int> eventCount) : base((string) (EventNamePrefix + "default_event"))
         {
             var platformProxy = PlatformProxyFactory.GetPlatformProxy();
             this[EventNamePrefix + "client_id"] = clientId;
             this[EventNamePrefix + "sdk_platform"] = platformProxy.GetProductName()?.ToLowerInvariant();
             this[EventNamePrefix + "sdk_version"] = MsalIdHelper.GetMsalVersion();
+            this[EventNamePrefix + "application_name"] = PlatformProxyFactory.GetPlatformProxy().GetApplicationName()?.ToLowerInvariant();
+            this[EventNamePrefix + "application_version"] = PlatformProxyFactory.GetPlatformProxy().GetApplicationVersion()?.ToLowerInvariant();
```
bgavrilMS: Aren't app name and app version also Pii ? #Closed

core/src/Telemetry/DefaultEvent.cs
```diff
@@ -25,18 +25,35 @@
...
+            this[EventNamePrefix + "device_id"] = HashPersonalIdentifier(PlatformProxyFactory.GetPlatformProxy().GetDeviceId()?.ToLowerInvariant());
+            this[EventNamePrefix + "ui_event_count"] = SetEventCount(EventNamePrefix + "ui_event", eventCount);
+            this[EventNamePrefix + "http_event_count"] = SetEventCount(EventNamePrefix + "http_event", eventCount);
+            this[EventNamePrefix + "cache_event_count"] = SetEventCount(EventNamePrefix + "cache_event", eventCount);
         }
+
+        private string SetEventCount(string eventName, ConcurrentDictionary<string, int> eventCount)
```
bgavrilMS: Method name should be "GetEventCount" because you are not chaning anything., #Closed

core/src/Telemetry/DefaultEvent.cs
```diff
@@ -25,18 +25,35 @@
...
+        {
+            if (!eventCount.ContainsKey(eventName))
+            {
+                return string.Empty;
```
bgavrilMS: If even isn't in the count dictionary, don't you want to return 0 (well, the string "0" anyway) ? #Closed

core/src/Telemetry/DefaultEvent.cs
```diff
@@ -25,18 +25,35 @@
...
-using Microsoft.Identity.Core;
+using System.Collections.Concurrent;
+using System.Globalization;
 
 namespace Microsoft.Identity.Core.Telemetry
 {
     internal class DefaultEvent : EventBase
     {
-        public DefaultEvent(string clientId) : base(EventNamePrefix + "default_event")
+        public DefaultEvent(string clientId, ConcurrentDictionary<string, int> eventCount) : base((string) (EventNamePrefix + "default_event"))
```
bgavrilMS: You are passing in a concurrent dictionary yet you are using it like a normal vanilla dictionary. ConcurrentDictionary needs to be used with TryGet / AddOrUpdate etc. type of methods to ensure concurrency. AFAIK using normal dictionary methods on a ConcurrentDictionary defeats the concurrency insurances.

This being said, all C# collections are thread safe when it comes to reading. So if at the time of reading you're done with writing, you can use IReadonlyDictionary

https://stackoverflow.com/questions/39086520/make-dictionary-read-only-in-c-sharp

core/src/Telemetry/XMsTelemetryInfo.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System.Globalization;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Identity.Core.Telemetry
+{
+    internal class XMsTelemetryInfo
```
bgavrilMS: Please add comments to the class and public methods / properties where they are not self explanatory (XMSTelemetryInfo is not, SpeInfo is not, Version is quite overloaded with meaning)

core/src/Telemetry/XMsTelemetryInfo.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        public string Version { get; set; }
+        public string ServerErrorCode { get; set; }
+        public string ServerSubErrorCode { get; set; }
+        public string TokenAge { get; set; }
+        public string SpeInfo { get; set; }
+    }
+
+    internal static class XMsTelemInfoHelper
```
bgavrilMS: It's a Parser, not a helper... #Closed

core/src/Telemetry/XMsTelemetryInfo.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            if (string.IsNullOrEmpty(headerValue))
+            {
+                return null;
+            }
+
+            string[] headerSegments = headerValue.Split(',');
+            if (headerSegments.Length == 0)
+            {
+                requestContext.Logger.Warning("Malformed x-ms-clitelem header");
```
bgavrilMS: You can add the actual header to make the log more useful.

core/src/Telemetry/XMsTelemetryInfo.cs
```diff
@@ -0,0 +1,69 @@
...
+                return null;
+            }
+            string headerVersion = headerSegments[0];
+
+            XMsTelemetryInfo xMsTelemetryInfo = new XMsTelemetryInfo();
+            xMsTelemetryInfo.Version = headerVersion;
+
+            if (headerVersion.Equals("1"))
```
bgavrilMS: Please use string.Equals("A", "b", StringComparison.OrdinalIgnoreCase), otherwise we'll get a build error from one of the analyzers.

core/src/Telemetry/XMsTelemetryInfo.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                // Verify the expected format "<version>, <error_code>, <sub_error_code>, <token_age>, <ring>"
+                Regex headerFormat = new Regex(@"^[1-9]+\.?[0-9|\\.]*,[0-9|\\.]*,[0-9|\\.]*,[^,]*[0-9\\.]*,[^,]*$");
```
bgavrilMS: This method is a bit too long, please pull the actual parsing logic to a private method...

msal/src/Microsoft.Identity.Client/Telemetry.cs
```diff
@@ -187,7 +193,7 @@ internal void Flush(string requestId)
 
             if (eventsToFlush.Count > 0)
             {
-                eventsToFlush.Insert(0, new DefaultEvent(ClientId));
+                eventsToFlush.Insert(0, new DefaultEvent(ClientId, eventCountToFlush));
```
bgavrilMS: eventCountToFlush could be null because the TryRemove can return false #Closed

msal/src/Microsoft.Identity.Client/Telemetry.cs
```diff
@@ -201,11 +207,36 @@ private List<EventBase> CollateOrphanedEvents(String requestId)
...
                     orphanedEvents.Add(orphan);
                 }
             }
             return orphanedEvents;
+        }
+
+        private void IncrementEventCount(string requestId, EventBase eventToIncrement)
+        {
+            string eventName;
+            if (eventToIncrement[EventBase.EventNameKey].Substring(0,10) == "msal.token")
```
bgavrilMS: Are you sure the eventToIncrement has that key? You should probably check first to avoid a null ref. #Closed

msal/tests/Test.MSAL.NET.Unit.net45/TelemetryTests.cs
```diff
@@ -411,5 +432,53 @@ public void AuthorityNotInTrustedHostList_AuthorityIsSetAsNullValueTest()
             }
             Assert.IsTrue(myReceiver.EventsReceived.Count > 0);
         }
+
+        [TestMethod]
+        [TestCategory("TelemetryInternalAPI")]
+        public void TelemetryEventCountsAreCorrectTest()
```
bgavrilMS: Since you've written some concurrent code, have a think about a test that tries to make sure your code works in multi threaded env. For example, you could create 100 tasks that all create events, start them all up, wait for all for all of them to finish and do your counts. #Closed

core/src/Telemetry/XMsTelemetryInfo.cs
bgavrilMS: You should pull these strings into constants as they will help you test./

core/src/Telemetry/XMsTelemetryInfo.cs
```diff
@@ -0,0 +1,69 @@
...
+        public string Version { get; set; }
+        public string ServerErrorCode { get; set; }
+        public string ServerSubErrorCode { get; set; }
+        public string TokenAge { get; set; }
+        public string SpeInfo { get; set; }
+    }
+
+    internal static class XMsTelemInfoHelper
+    {
+        internal static XMsTelemetryInfo parseXMsTelemHeader(string headerValue, RequestContext requestContext)
```
bgavrilMS: This is not tested and it should be very easy to test. Just pass in some headerValues and ensure that parsing is ok or that each of the failure conditions is met. You can use a mocked logger to verify the errors. #Closed

bgavrilMS (changes requested):

core/src/IPlatformProxy.cs
sangonzal: Done #Closed

core/src/OAuth2/OAuth2Client.cs
sangonzal: done #Closed

core/src/Platforms/iOS/iOSPlatformProxy.cs
sangonzal: You're right, added fallbacks to app name, version, and device id #Closed

core/src/Platforms/net45/NetDesktopPlatformProxy.cs
sangonzal: Device Id is PII. It's being hashed before it's set in DefaultEvent. We are doing all hashing in the telemetry classes, so I put it there for consistency. I can bring it here if that makes more sense though. #Closed

core/src/Platforms/netcore/NetCorePlatformProxy.cs
sangonzal: done #Closed

core/src/Telemetry/DefaultEvent.cs
sangonzal: I thought they were not. They are not hashed in MATS telemetry data. @henrik-me thoughts? #Closed

core/src/Telemetry/DefaultEvent.cs
sangonzal: done #Closed

core/src/Telemetry/XMsTelemetryInfo.cs
sangonzal: done  #Closed

core/src/Telemetry/XMsTelemetryInfo.cs
sangonzal: done

core/src/Telemetry/XMsTelemetryInfo.cs
sangonzal: done #Closed

core/src/Telemetry/XMsTelemetryInfo.cs
sangonzal: done

msal/src/Microsoft.Identity.Client/Telemetry.cs
sangonzal: eventToIncrement will always have that key, it is a requirement for instantiation of EventBase #Closed

core/src/OAuth2/OAuth2Client.cs
sangonzal: done #Closed

msal/src/Microsoft.Identity.Client/Telemetry.cs
sangonzal: done #Closed

core/src/Telemetry/DefaultEvent.cs
sangonzal: done #Closed

core/src/Telemetry/XMsTelemetryInfo.cs
sangonzal: done #Closed

core/src/Platforms/net45/NetDesktopPlatformProxy.cs
jennyf19: I think you could make a comment in the code here, so it's clear the hashing is happening elsewhere.

---
In reply to: [228231893](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1352#discussion_r228231893) [](ancestors = 228231893)

core/src/Telemetry/DefaultEvent.cs
jennyf19: I don't think app name and version are considered Pii. I have a link to the chart, i'll post it in teams.

---
In reply to: [228236523](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1352#discussion_r228236523) [](ancestors = 228236523)

core/src/Telemetry/HttpEvent.cs
```diff
@@ -77,5 +83,34 @@ public string OauthErrorCode
...
+        public string RequestIdHeader
+        {
+            set { this[RequestIdHeaderKey] = value; }
+        }
+        public string TokenAge
+        {
+            set { this[TokenAgeKey] = value; }
+        }
+
+        public string SpeInfo
```
jennyf19, line 114: >SpeInfo [](start = 22, length = 7)

What's SpeInfo?

jennyf19:
> using System.Globalization;

Remember to add the header information. just copy/paste from another class.

---
Refers to: core/src/Telemetry/XMsTelemetryInfo.cs:1 in 983892a. [](commit_id = 983892ab7a0400ca7c03b559a82c9c8261d008fc, deletion_comment = False)

core/src/Telemetry/DefaultEvent.cs
sangonzal: Good point. ConcurrentDictionary does not implement IReadOnlyDictionary until .NET framework 4.6.2. so made it IDictionary instead #Closed

core/src/Platforms/net45/NetDesktopPlatformProxy.cs
henrik-me: also, where exactly is that other place?

---
In reply to: [228346528](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1352#discussion_r228346528) [](ancestors = 228346528,228231893)

henrik-me: agree there should be a fallback

---
In reply to: [228073704](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1352#discussion_r228073704) [](ancestors = 228073704)

core/src/Platforms/netcore/NetCorePlatformProxy.cs
henrik-me: pls. ensure to hash it...

---
In reply to: [228235975](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1352#discussion_r228235975) [](ancestors = 228235975)

core/src/Telemetry/DefaultEvent.cs
henrik-me: IMO, it depends... and we will not necessarily know unless we let the user specify these. IMO, we should consider hashing these or we should have a flag allowing for these to be sent as is.
to determine if these values are of GDPR/PII interest it requires examples. E.g. grabbing name and version from a running assembly could reveal an internal code name... IMO, these values should be set by the developer and we should document how we get the values if not special.

---
In reply to: [228347497](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1352#discussion_r228347497) [](ancestors = 228347497,228236523)

core/src/Telemetry/DefaultEvent.cs
```diff
@@ -25,18 +25,35 @@
...
     internal class DefaultEvent : EventBase
     {
-        public DefaultEvent(string clientId) : base(EventNamePrefix + "default_event")
+        public DefaultEvent(string clientId, IDictionary<string, int> eventCount) : base((string) (EventNamePrefix + "default_event"))
         {
             var platformProxy = PlatformProxyFactory.GetPlatformProxy();
             this[EventNamePrefix + "client_id"] = clientId;
             this[EventNamePrefix + "sdk_platform"] = platformProxy.GetProductName()?.ToLowerInvariant();
             this[EventNamePrefix + "sdk_version"] = MsalIdHelper.GetMsalVersion();
+            this[EventNamePrefix + "application_name"] = PlatformProxyFactory.GetPlatformProxy().GetCallingAssemblyName()?.ToLowerInvariant();
```
henrik-me: >PlatformProxyFactory.GetPlatformProxy() [](start = 57, length = 39)

pls. use the platformProxy variable from above, instead of calling the factory. Same below as well. #Closed

core/src/Telemetry/DefaultEvent.cs
```diff
@@ -25,18 +25,35 @@
...
 {
     internal class DefaultEvent : EventBase
     {
-        public DefaultEvent(string clientId) : base(EventNamePrefix + "default_event")
+        public DefaultEvent(string clientId, IDictionary<string, int> eventCount) : base((string) (EventNamePrefix + "default_event"))
         {
             var platformProxy = PlatformProxyFactory.GetPlatformProxy();
             this[EventNamePrefix + "client_id"] = clientId;
             this[EventNamePrefix + "sdk_platform"] = platformProxy.GetProductName()?.ToLowerInvariant();
             this[EventNamePrefix + "sdk_version"] = MsalIdHelper.GetMsalVersion();
```
henrik-me, line 40: >MsalIdHelper [](start = 52, length = 12)

isn't this available in the platformProxy as well? #Closed

core/src/Telemetry/Event.cs
```diff
@@ -87,5 +87,12 @@ public void Stop()
             string scrubbedPath = String.Join("/", pieces);
             return uri.Scheme + "://" + uri.Authority + scrubbedPath;
         }
+
+        public static string HashPersonalIdentifier(string valueToHash)
```
henrik-me: >static [](start = 15, length = 6)

consider not making this static

core/src/Telemetry/HttpEvent.cs
henrik-me: I'm curious about this one as well :)

---
In reply to: [228347941](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1352#discussion_r228347941) [](ancestors = 228347941)

core/src/Telemetry/XmsCliTelemInfo.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// </summary>
+        public string TokenAge { get; set; }
+
+        /// <summary>
+        /// Bundle id for spe_ring info.
+        /// </summary>
+        public string SpeInfo { get; set; }
+    }
+
+    internal static class XmsCliTelemInfoParser
```
henrik-me: >internal static class XmsCliTelemInfoParser [](start = 4, length = 43)

Separate file

core/src/Telemetry/XmsCliTelemInfo.cs
```diff
@@ -0,0 +1,123 @@
...
+    {
+        internal static XmsCliTelemInfo parseXMsTelemHeader(string headerValue, RequestContext requestContext)
```
henrik-me: >p [](start = 40, length = 1)

nit: Captitalize

core/src/Telemetry/XmsCliTelemInfo.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                requestContext.Logger.Warning(
+                    string.Format(CultureInfo.InvariantCulture,
+                    TelemetryError.XmsUnrecognizedHeaderVersion, headerVersion));
+                return null;
+            }
+            return xMsTelemetryInfo;
+        }
+
+        private static MatchCollection parseHeader(string headerValue)
```
henrik-me: >p [](start = 39, length = 1)

same... CamelCase'ing

core/src/Telemetry/XmsCliTelemInfo.cs
```diff
@@ -0,0 +1,123 @@
...
+                    string.Format(CultureInfo.InvariantCulture,
+                    TelemetryError.XmsTelemMalformed, headerValue));
+                return null;
+            }
+            string headerVersion = headerSegments[0];
+
+            XmsCliTelemInfo xMsTelemetryInfo = new XmsCliTelemInfo();
+            xMsTelemetryInfo.Version = headerVersion;
+
+            if (string.Equals(headerVersion, "1"))
```
henrik-me: >"1" [](start = 45, length = 3)

const?

henrik-me: >if (string.Equals(headerVersion, "1")) [](start = 12, length = 38)

"Code readability" : consider inverting, that allows you to exit early and reduce nesting.

core/src/Telemetry/XmsCliTelemInfo.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                MatchCollection formatMatcher = parseHeader(headerValue);
```
henrik-me: this method should have a more meaningful name, I have no idea what I should look for here (thus assessing the < 1 is difficult) unless I go to the method and examine what it does.

core/src/Telemetry/XmsCliTelemInfo.cs
```diff
@@ -0,0 +1,123 @@
...
+                    requestContext.Logger.Warning(
+                            string.Format(CultureInfo.InvariantCulture,
+                            TelemetryError.XmsTelemMalformed, headerValue));
+                    return null;
+                }
+
+                int ErrorCodeIndex = 1;
+                int SubErrorCodeIndex = 2;
+                int TokenAgeIndex = 3;
+                int SpeInfoIndex = 4;
```
henrik-me: Seems like these should simply be class consts?

core/src/Telemetry/XmsCliTelemInfo.cs
henrik-me: >static [](start = 13, length = 6)

while the class holds no state, consider if this should not be static. 
Context: statics are evil and suddenly state may be applied to the class.

msal/src/Microsoft.Identity.Client/Telemetry.cs
```diff
@@ -201,11 +216,36 @@ private List<EventBase> CollateOrphanedEvents(String requestId)
...
```
henrik-me: >msal.token" [](start = 77, length = 11)

const

msal/src/Microsoft.Identity.Client/Telemetry.cs
```diff
@@ -201,11 +216,36 @@ private List<EventBase> CollateOrphanedEvents(String requestId)
...
+            {
+                eventName = "msal.cache_event";
```
henrik-me: >"msal.cache_event"; [](start = 28, length = 19)

const #Closed

core/src/Platforms/net45/NetDesktopPlatformProxy.cs
sangonzal: Hashed in DefaultEvent #Closed

core/src/Platforms/netcore/NetCorePlatformProxy.cs
sangonzal: Hashed in DefaultEvent #Closed

core/src/Telemetry/XMsTelemetryInfo.cs
sangonzal: Done

core/src/Telemetry/HttpEvent.cs
sangonzal: SpeInfo is the server's version of deployment group/ring

core/src/Telemetry/DefaultEvent.cs
sangonzal: done #Closed

core/src/Platforms/net45/NetDesktopPlatformProxy.cs
```diff
@@ -204,6 +207,23 @@ public string GetProductName()
             return _isMsal ? "MSAL.Desktop" : "PCL.Desktop";
         }
 
+        public string GetCallingAssemblyName()
+        {
+            return Assembly.GetEntryAssembly()?.GetName()?.Name?.ToString();
```
bgavrilMS: Did you test on an asp.net and asp.net core website to see what this gives? I see that there are quite a few cases where this will be null. Would it be a good idea to fallback to Process.GetCurrentProcess().ProcessName when EntryAssembly returns null?

core/src/Telemetry/DefaultEvent.cs
sangonzal: It is not. Only msal name is available (GetProductName) #Closed

msal/tests/Test.MSAL.NET.Unit.net45/TelemetryTests.cs
```diff
@@ -411,5 +439,77 @@ public void AuthorityNotInTrustedHostList_AuthorityIsSetAsNullValueTest()
...
+                }
+            }
+            //Assert
+            foreach(Dictionary<string, string> telemetryEvent in myReceiver.EventsReceived)
+            {
+                if(telemetryEvent[EventBase.EventNameKey] == "msal.default_event")
+                {
+                    Assert.AreEqual("2", telemetryEvent["msal.http_event_count"]);
+                    Assert.AreEqual("2", telemetryEvent["msal.cache_event_count"]);
+                    Assert.AreEqual("0", telemetryEvent["msal.ui_event_count"]);
```
henrik-me: imo, these texts should be consts in the library #Closed

core/src/Telemetry/Event.cs
sangonzal: done #Closed

henrik-me (changes requested):
:clock1:

core/src/Telemetry/XmsCliTelemInfo.cs
sangonzal: done

core/src/Telemetry/XmsCliTelemInfo.cs
sangonzal: done

core/src/Telemetry/XmsCliTelemInfo.cs
sangonzal: done

core/src/IPlatformProxy.cs
```diff
@@ -35,6 +35,24 @@ internal interface IPlatformProxy
 
         Task<bool> IsUserLocalAsync(RequestContext requestContext);
 
+        /// <summary>
+        /// Returns the name of the calling assembly
+        /// </summary>
+        /// <returns></returns>
+        string GetCallingAssemblyName();
```
bgavrilMS: Sorry I didn't realize this is an interface for all platforms. iOS and Android don't really have assemblies, so perhaps calling these as GetCallingAppName / GetCallingAppVersion would be better. #Closed

core/src/Telemetry/XmsCliTelemInfo.cs
sangonzal: done

core/src/Telemetry/XmsCliTelemInfo.cs
sangonzal: done

core/src/Telemetry/XmsCliTelemInfo.cs
sangonzal: done

msal/src/Microsoft.Identity.Client/Telemetry.cs
sangonzal: done #Closed

msal/src/Microsoft.Identity.Client/Telemetry.cs
sangonzal: done #Closed

msal/tests/Test.MSAL.NET.Unit.net45/TelemetryTests.cs
sangonzal: done #Closed

core/src/IPlatformProxy.cs
sangonzal: No prob, switched them back.  #Closed

msal/tests/Test.MSAL.NET.Unit.net45/TelemetryTests.cs
sangonzal: done #Closed

core/src/Telemetry/XmsCliTelemInfo.cs
sangonzal: done

core/src/Telemetry/XmsCliTelemInfo.cs
sangonzal: done

core/src/Microsoft.Identity.Core.csproj
```diff
@@ -99,6 +99,10 @@ in ADAL and MSAL and to ease unit testing. No projects should take a dependency
     <Compile Include="Platforms\iOS\**\*.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Remove="Telemetry\XmsCliTelemInfoParser.cs" />
```
MarkZuber: You can thank VS, but you should delete this itemgroup  chunk here.

core/src/Platforms/Android/AndroidPlatformProxy.cs
```diff
@@ -134,6 +134,25 @@ public string GetProductName()
         {
             return _isMsal ? "MSAL.Xamarin.Android" : "PCL.Android";
         }
+        public string GetCallingApplicationName()
+        {
+            return Android.App.Application.Context.ApplicationInfo?.LoadLabel(Android.App.Application.Context.PackageManager);
+
```
MarkZuber: nit: remove extra blank line here and 140 and 146 and 154

core/src/Platforms/iOS/iOSPlatformProxy.cs
```diff
@@ -120,6 +122,20 @@ public string GetProductName()
         {
             return _isMsal ? "MSAL.Xamarin.iOS" : "PCL.iOS";
         }
+        public string GetCallingApplicationName()
+        {
+            return (NSString)NSBundle.MainBundle?.InfoDictionary?["CFBundleName"];
```
MarkZuber, line 137: If these are empty, how will that impact our telemetry data?  Will we get it keyed off of "null" or will it be dropped?  If it will be dropped if null, i'd suggest having a default value for these like "unknown_calling_application" and "unknown_device_id" so taht we can bucketize the counts of telemetry that are failing with unknown/uncalculatable values. #Closed

core/src/Platforms/net45/NetDesktopPlatformProxy.cs
```diff
@@ -138,14 +141,14 @@ public string GetEnvironmentVariable(string variable)
...
         {
-            if (IsWindows)
-            {
-                return WindowsNativeMethods.GetProcessorArchitecture();
-            }
-            else
-            {
-                return null;
-            }
+            if (IsWindows)
```
MarkZuber: nit: collapse to ternary operator here (like GetOperatingSystem() below)

core/src/Platforms/net45/NetDesktopPlatformProxy.cs
```diff
@@ -204,6 +207,23 @@ public string GetProductName()
...
+        public string GetCallingApplicationVersion()
+        {
+            return Assembly.GetEntryAssembly()?.GetName()?.Version?.ToString();
+        }
+
+        public string GetDeviceId()
+        {
+            return  NetworkInterface.GetAllNetworkInterfaces().Where(nic => nic.OperationalStatus == OperationalStatus.Up)
+                .Select(nic => nic.GetPhysicalAddress()?.ToString()).FirstOrDefault();
+
```
MarkZuber: nit: remove extra blank line

core/src/Telemetry/Event.cs
```diff
@@ -87,5 +87,12 @@ public void Stop()
             string scrubbedPath = String.Join("/", pieces);
             return uri.Scheme + "://" + uri.Authority + scrubbedPath;
         }
+
+        public string HashPersonalIdentifier(string valueToHash)
+        {
+            var crypto = PlatformProxyFactory.GetPlatformProxy().CryptographyManager;
+            return crypto.CreateBase64UrlEncodedSha256Hash(valueToHash);
+
```
MarkZuber: nit: remove extra blank line

core/src/Telemetry/HttpEvent.cs
MarkZuber: Please comment/document the meaning of this field.   Links to telemetry/server documentation?

core/src/Telemetry/XmsCliTelemInfoParser.cs
```diff
@@ -0,0 +1,94 @@
...
+            }
+
+            string headerVersion = headerSegments[0];
+            XmsCliTelemInfo xMsTelemetryInfo = new XmsCliTelemInfo();
+            xMsTelemetryInfo.Version = headerVersion;
+
+            if (!string.Equals(headerVersion, ExpectedCliTelemHeaderVersion))
+            {
+                requestContext.Logger.Warning(
+                    string.Format(CultureInfo.InvariantCulture,
```
MarkZuber: This makes an indirect assumption that the string has one {0} parameter in it.  This can easily get broken and not noticed by the compiler.  Suggestion:  make a method for this that takes the one parameter and returns the formatted string? #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -1,332 +1,337 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+using System;
```
MarkZuber: something got messed up with line endings.  can you fix?

core/src/Microsoft.Identity.Core.csproj
sangonzal: done

core/src/Platforms/Android/AndroidPlatformProxy.cs
sangonzal: done

core/src/Platforms/net45/NetDesktopPlatformProxy.cs
sangonzal: done

MarkZuber (changes requested):
🕐

core/src/Platforms/net45/NetDesktopPlatformProxy.cs
sangonzal: done

core/src/Telemetry/Event.cs
sangonzal: done

core/src/Platforms/iOS/iOSPlatformProxy.cs
sangonzal: In case of these values being null, we would still get a telemetry batch with these fields - they would just be empty. #Closed

core/src/Telemetry/XmsCliTelemInfoParser.cs
sangonzal: Fixed #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
sangonzal: done

core/src/Telemetry/DefaultEvent.cs
```diff
@@ -25,18 +25,35 @@
...
+            this[EventNamePrefix + "application_name"] = HashPersonalIdentifier(platformProxy.GetCallingApplicationName()?.ToLowerInvariant());
+            this[EventNamePrefix + "application_version"] = HashPersonalIdentifier(platformProxy.GetCallingApplicationVersion()?.ToLowerInvariant());
+            this[EventNamePrefix + "device_id"] = HashPersonalIdentifier(platformProxy.GetDeviceId()?.ToLowerInvariant());
+            this[EventNamePrefix + "ui_event_count"] = GetEventCount(EventNamePrefix + "ui_event", eventCount);
+            this[EventNamePrefix + "http_event_count"] = GetEventCount(EventNamePrefix + "http_event", eventCount);
+            this[EventNamePrefix + "cache_event_count"] = GetEventCount(EventNamePrefix + "cache_event", eventCount);
```
henrik-me, line 46: >EventNamePrefix + "cache_event_count" [](start = 17, length = 37)

seems like some of these are specified in the HttpEvent class as const's. That would be beneficial.

core/src/Telemetry/TelemetryConstants.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Microsoft.Identity.Core.Telemetry
+{
+    internal class TelemetryError
+    {
+        public const string XmsCliTelemMalformed = "Malformed x-ms-clitelem header: '{0}'";
+        public const string XmsUnrecognizedHeaderVersion = "Header version '{0}' unrecognized";
+    }
+
+    internal class TelemetryEventProperties
```
henrik-me, line 36: should have been different classes

henrik-me:
I left a few comments open. Pls. follow up on those (e.g. create a work item to track that). Otherwise looks good imo.

henrik-me (approved):
:shipit:

core/src/Telemetry/DefaultEvent.cs
sangonzal: None of these are specified in HttpEvent. There should be no overlap in fields for different types of events. On a separate note, it might make sense to add these all in a constants class. Adding to a work item to track.

core/src/Telemetry/TelemetryConstants.cs
sangonzal: different files? I was following convention set in the other constant classes. Will add also add item to track.

PR #1357, opened 2018-10-26
Prepare tests for other platforms

bgavrilMS:
This just adds the precompilation flags for UWP and for most of Android and iOS tests. Getting a test runner for the tests will be a separate PR.

### DIFF
--- a/msal/tests/Test.MSAL.NET.Unit.net45/CacheTests/TokenCacheTests.cs
+++ b/msal/tests/Test.MSAL.NET.Unit.net45/CacheTests/TokenCacheTests.cs
@@ -85,19 +85,7 @@ namespace Test.MSAL.NET.Unit.CacheTests
             _cache.Clear();
         }
 
-        [TestMethod]
-        [TestCategory("TokenCacheTests")]
-        public void CanDeserializeTokenCacheInNet462()
-        {
-            var previousLogLevel = Logger.Level;
-            // Setting LogLevel.Verbose causes certain static dependencies to load
-            Logger.Level = LogLevel.Verbose;
-            var tokenCache = new TokenCache();
-            tokenCache.Deserialize(null);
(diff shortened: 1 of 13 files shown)


msal/tests/Test.MSAL.NET.Unit.net45/CacheTests/TokenCacheTests.cs
```diff
@@ -85,19 +85,7 @@ public void TestCleanup()
             _cache.Clear();
         }
 
-        [TestMethod]
```
bgavrilMS, line 88: moved lower down.

msal/tests/Test.MSAL.NET.Unit.net45/CryptographyTests.cs
```diff
@@ -36,6 +36,7 @@
 
 namespace Test.MSAL.NET.Unit
 {
+#if !WINDOWS_APP // not available on UWP
```
MarkZuber, line 39: suggestion:  can we setup additional defines somewhere globally (like in the msal csproj?) where we define a feature #define (e.g. MSAL_IS_USERNAME_PASSWORD_AVAILABLE) and we can use these here?  We already decide the feature directories to include within the csproj and it would be great if we could centrally manage those with a define and drive the unit tests against them.  I just see it being hard to get all of the unit tests' #defines maintained properly as I see the #if !ios && !android && !windows_app in this PR.

MarkZuber (approved):

msal/tests/Test.MSAL.NET.Unit.net45/CryptographyTests.cs
bgavrilMS: Yes, very good idea. In fact this is how a lot of other multi platform libraries are doing it - they do conditional compilation based on feature flags not on platform flags. I'll raise an issue.

PR #1358, opened 2018-10-26
Fix Integration test by using correct tenant id

bgavrilMS:


core/tests/Test.Microsoft.Identity.SideBySide/UnifiedCacheTests.cs.cs
```diff
@@ -33,52 +33,53 @@
...
-using msal::Microsoft.Identity.Client;
 
 namespace Test.MSAL.NET.Integration
 {
     [TestClass]
     public class UnifiedCacheTests
     {
         public const string ClientId = "0615b6ca-88d4-4884-8729-b178178f7c27";
-        public const string Authority = "https://login.microsoftonline.com/d34a1bb7-3481-4d5f-8b94-f3cc27bf8eac/";
+        //public const string AuthorityTemplate = "https://login.microsoftonline.com/d34a1bb7-3481-4d5f-8b94-f3cc27bf8eac/";
```
MarkZuber: dead code?  if we need to keep the guid around for testing/notes, can you add a comment around dead code?

MarkZuber (approved):

core/tests/Test.Microsoft.Identity.SideBySide/UnifiedCacheTests.cs.cs
bgavrilMS: I don't think we do, fixed by deleting code.

PR #1360, opened 2018-10-26
fix issue with sharing in memory cache for different instances of TokenCache

SomkaPe:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/656


henrik-me (approved):
:shipit:

PR #1361, opened 2018-10-28
Remove dependencies on statics in internal code for Telemetry

MarkZuber:
The static Telemetry instance is pervasive throughout the code.  This PR cleans up the majority of the code to depend on the ITelemetryManager for interfacing with telemetry and only within ClientApplicatonBase do we bind to the existing public Telemetry static in default initialization.  The goal being that we can eventually make a breaking public API contract change to allow developers to register the ITelemetryReceiver with a PublicClientApp / ConfidentialClientApp instead of having a process wide singleton.


msal/src/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -70,35 +69,42 @@ static ClientApplicationBase()
...
+        ///  <param name="telemetryManager"></param>
         internal ClientApplicationBase(string clientId, string authority, string redirectUri,
-            bool validateAuthority, IHttpManager httpManager)
+            bool validateAuthority, IHttpManager httpManager, ITelemetryManager telemetryManager)
         {
             HttpManager = httpManager ?? new HttpManager();
             CryptographyManager = PlatformProxyFactory.GetPlatformProxy().CryptographyManager;
             WsTrustWebRequestManager = new WsTrustWebRequestManager(HttpManager);
-            TelemetryManager = new TelemetryManager(Telemetry.GetInstance());  // TODO: pass this in?
+            TelemetryManager = telemetryManager ?? new TelemetryManager(Telemetry.GetInstance());  // TODO: pass this in?
```
bgavrilMS: Is this todo actual?

core/src/Telemetry/TelemetryManager.cs
```diff
@@ -0,0 +1,243 @@
...
+        public ITelemetryReceiver TelemetryReceiver
+        {
+            get
+            {
+                lock (_lockObj)
+                {
+                    return _telemetryReceiver;
+                }
+            }
+            set
```
bgavrilMS, line 77: I haven't unpacked the code to see where the setter is used, but if we can replace the setter with just building a new object, then TelemetryManager would become immutable and there would be no need for a lock.

core/src/Telemetry/TelemetryManager.cs
```diff
@@ -0,0 +1,243 @@
...
+        internal readonly ConcurrentDictionary<string, List<EventBase>> CompletedEvents = new ConcurrentDictionary<string, List<EventBase>>();
+
+        internal void StartEvent(string requestId, EventBase eventToStart)
+        {
+            if (!HasReceiver() || string.IsNullOrWhiteSpace(requestId))
+            {
+                return;
+            }
+
+            EventsInProgress[new Tuple<string, string>(requestId, eventToStart[EventBase.EventNameKey])] = eventToStart;
```
bgavrilMS: Would it be a lot of work to replace this tuple with a class or a struct? These are so difficult to read...

core/src/Telemetry/TelemetryManager.cs
```diff
@@ -0,0 +1,243 @@
...
+        }
+
+        internal void StopEvent(string requestId, EventBase eventToStop)
+        {
+            if (!HasReceiver() || string.IsNullOrWhiteSpace(requestId))
+            {
+                return;
+            }
+
+                var eventKey = new Tuple<string, string>(requestId, eventToStop[EventBase.EventNameKey]);
```
bgavrilMS: Nit: code is misaligned

core/src/Telemetry/iTelemetry.cs
```diff
@@ -25,20 +25,12 @@
...
 
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
-
 namespace Microsoft.Identity.Core.Telemetry
 {
     internal interface ITelemetry
```
bgavrilMS, line 30: The file name does not match the interface name.

core/src/Telemetry/TelemetryManager.cs
```diff
@@ -0,0 +1,243 @@
...
+            {
+                onlySendFailureTelemetry = _telemetryReceiver?.OnlySendFailureTelemetry ?? false;
+            }
+
+            if (onlySendFailureTelemetry)
+            {
+                // iterate over Events, if the ApiEvent was successful, don't dispatch
+                bool shouldRemoveEvents = false;
+
+                foreach (var anEvent in eventsToFlush)
```
bgavrilMS: Nit: possibly easier to read: eventsToFlush.Any(ev => ev is ApiEvent a && a.WasSuccessful)

msal/tests/Test.MSAL.NET.Unit.net45/CoreTests/Telemetry/TelemetryHelperTests.cs
```diff
@@ -35,140 +35,117 @@ namespace Test.Microsoft.Identity.Core.Unit.Telemetry
...
-
-            Assert.AreEqual(1, telem.NumStartEventCalls);
-            Assert.AreEqual(1, telem.NumStopEventCalls);
-
-            Assert.AreEqual(expectedRequestId, telem.LastStartEventRequestId);
-            Assert.AreEqual(expectedRequestId, telem.LastStopEventRequestId);
-
-            Assert.AreEqual(expectedStartEvent, telem.LastEventToStart);
-            Assert.AreEqual(expectedStopEvent, telem.LastEventToStop);
+            // TODO: implement this to check for proper events...
```
bgavrilMS: Dead code?

msal/tests/Test.MSAL.NET.Unit.net45/CoreTests/Telemetry/TelemetryHelperTests.cs
```diff
@@ -35,140 +35,117 @@ namespace Test.Microsoft.Identity.Core.Unit.Telemetry
...
             {
             }
 
-            ValidateResults(_telem, requestId, _startEvent, _stopEvent, false);
+            ValidateResults(_testReceiver, RequestId, ClientId, _startEvent, _stopEvent, false);
         }
 
         [TestMethod]
         [TestCategory("TelemetryHelperTests")]
         public void TestTelemetryHelperWithDifferentStopStartEventsWithFlush()
```
bgavrilMS: Do any of these tests cover finding the orphaned events?

bgavrilMS (commented):
Great improvement, this was one of the last big ugly statics ...

msal/src/Microsoft.Identity.Client/ClientApplicationBase.cs
MarkZuber: Eventually, we need to get rid of the static from the public api.  But you're right, this TODO can't go away until we do that so I'll remove this.

core/src/Telemetry/TelemetryManager.cs
MarkZuber: Our customer developers can set the telemetry receiver at any time, but the TelemetryManager will be constructed and injected into existing ClientApplication instances so we can't just replace the entire TelemetryManager everywhere but we need to update where to route telemetry events to.  

Take a look in ClientApplicationBase line 72, i'm assuming this will eventually become public.  The other option in the public API would be to pass in the TelemetryReceiver as part of ClientApplication construction.  I'd prefer to do this.  But until we can remove the public Telemetry static for setting the receiver whenever someone feels like it (and at a global process level) having this setter (and the appropriate locks) are the only solution I can see.  Once we can break the API a bit for how these get set on a per-PCA basis then we can clean this up.

core/src/Telemetry/TelemetryManager.cs
MarkZuber: I agree.  Let me look into what we can do here...

core/src/Telemetry/TelemetryManager.cs
MarkZuber: fixed

core/src/Telemetry/iTelemetry.cs
MarkZuber: fixed

core/src/Telemetry/TelemetryManager.cs
MarkZuber: I played with various cleanups of this and never got one I liked. This is exactly what I want.  Will fix.

msal/tests/Test.MSAL.NET.Unit.net45/CoreTests/Telemetry/TelemetryHelperTests.cs
MarkZuber: Incomplete code.  I sent out the PR initially to check for passage of unit tests and get initial feedback.  I still need to restore some sanity to these tests before completion...

henrik-me:
generally, like the change.

core/src/Telemetry/TelemetryManager.cs
MarkZuber: Fixed

core/src/Telemetry/TelemetryManager.cs
bgavrilMS: Makes sense, thanks!

bgavrilMS (approved):

PR #1364, opened 2018-10-29
Complete fix for issue 656

bgavrilMS:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/656


core/src/Platforms/Android/AndroidPlatformProxy.cs
```diff
@@ -135,11 +135,21 @@ public string GetProductName()
             return _isMsal ? "MSAL.Xamarin.Android" : "PCL.Android";
         }
 
+        private readonly ILegacyCachePersistence createLegacyCachePersistence = new AndroidLegacyCachePersistence();
```
MarkZuber: Why on android is the value re-used when it's created new on every other platform?

msal/tests/Test.MSAL.NET.Unit.net45/CoreTests/PlatformProxyFactoryTests.cs
```diff
@@ -43,5 +43,26 @@ public void PlatformProxyFactoryCachesTheProxy()
...
+            // Act and Assert
+            Assert.AreNotEqual(
+                factory.CreateLegacyCachePersistence(),
+                factory.CreateLegacyCachePersistence());
+
+            Assert.AreNotEqual(
+                factory.CreateTokenCacheAccessor(),
+                factory.CreateTokenCacheAccessor());
+
+            Assert.AreEqual(
```
MarkZuber: Do you want "AreEqual" here or "AreSame" (which checks for the same reference)

MarkZuber (changes requested):
🕐

msal/tests/Test.MSAL.NET.Unit.net45/CoreTests/PlatformProxyFactoryTests.cs
bgavrilMS: fixed.

core/src/Platforms/Android/AndroidPlatformProxy.cs
bgavrilMS: Good catch, fixed.

MarkZuber (approved):

core/src/Platforms/Android/AndroidPlatformProxy.cs
henrik-me: as far as I can see this is still not fixed (uniform across languages)

and also why would it need to be created new for every call? Name is meaningful, however isn't there a need to retrieve the one created?

---
In reply to: [229044749](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1364#discussion_r229044749) [](ancestors = 229044749)

henrik-me:
See also: [https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1360](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1360)

henrik-me (changes requested):
:clock1:

core/src/Platforms/Android/AndroidPlatformProxy.cs
bgavrilMS: LegacyCachePersistance is not a singleton, a separate instance is created with each TokenCache. Also on some platforms, such as net45, it has state, so reusing the same instance will have consequences.

PR #1365, opened 2018-10-30
not removing account cache entities

SomkaPe:
Taking to account that account cache entity can be shared by different applications, RemoveAccount api should not delete it because it is application specific api , and it can break sso in other applications .
This can happen if cache is shared by different applications and same user log in in both


core/tests/Test.Microsoft.Identity.SideBySide/UnifiedCacheTests.cs
```diff
@@ -540,7 +540,6 @@ private void AssertMsalCacheIsEmpty()
             Assert.IsTrue(msalCache.tokenCacheAccessor.GetAllAccessTokensAsString().Count == 0);
             Assert.IsTrue(msalCache.tokenCacheAccessor.GetAllRefreshTokensAsString().Count == 0);
             Assert.IsTrue(msalCache.tokenCacheAccessor.GetAllIdTokensAsString().Count == 0);
-            Assert.IsTrue(msalCache.tokenCacheAccessor.GetAllAccountsAsString().Count == 0);
```
henrik-me: > Assert.IsTrue(msalCache.tokenCacheAccessor.GetAllAccountsAsString().Count == 0); [](start = 11, length = 81)

nit: If the cache initially comes empty this could still be 0. I see you added tests for this scenario, perhaps it still makes sense to assert that it's 0 or not?

henrik-me (approved):
:shipit:

msal/tests/Test.MSAL.NET.Unit.net45/UnifiedCacheTests.cs
```diff
@@ -131,6 +124,92 @@ public void UnifiedCache_MsalStoresToAndReadRtFromAdalCache()
...
+                Assert.AreEqual(2, tokenCache1.tokenCacheAccessor.GetAllIdTokensAsString().Count);
+
+                // remove account from app
+                app.RemoveAsync(app.GetAccountsAsync().Result.First()).Wait();
+
+                // make sure account removed from app
+                Assert.AreEqual(0, app.GetAccountsAsync().Result.Count());
+
+                // make sure account Not removed from app1
+                Assert.AreEqual(1, app1.GetAccountsAsync().Result.Count());
```
jennyf19, line 210: do you want to check app1 too?
`Assert.AreEqual(0, app1.GetAccountsAsync().Result.Count()); `

jennyf19 (approved):
🐱

core/tests/Test.Microsoft.Identity.SideBySide/UnifiedCacheTests.cs
SomkaPe: right, It is used for different purposes, reverted it back and added AssertNoCredentialsInMsalCache for validation of remove account api

msal/tests/Test.MSAL.NET.Unit.net45/UnifiedCacheTests.cs
SomkaPe: yes, i check remove account for app return 0 account but for app1 1 account - so remove call on app did not affect app1

PR #1366, opened 2018-10-30
remove required parameter for ext_expires_on

jennyf19:
this should not be required...for example, b2c accounts will not have this, and then we get a serialization exception. 
image


MarkZuber (approved):

henrik-me (approved):
:shipit:

PR #1369, opened 2018-10-31
Consolidate test initialization and add more logging to a problematic test

bgavrilMS:


msal/tests/Test.MSAL.NET.Unit.net45/RequestsTests/DeviceCodeRequestTests.cs
```diff
@@ -194,6 +194,14 @@ public void VerifyAuthorizationPendingErrorDoesNotLogError()
             
             Logger.LogCallback = (level, message, pii) =>
             {
+                if (level == LogLevel.Error)
```
bgavrilMS, line 197: This test sometimes fails. I've looked at if a good few hours and I can't understand why. The failing reason is that an error message is logged, which is a valid reason for failing. However, the error message comes from a completely different test (an IWA test), suggesting tests are run in parallel. I managed to repro even when passing --MSTest.MaxCpuCount=1. 

The only other idea I have is that there is an async test where we forgot to add an await and the code keeps on going. Hopefully the stack trace will tell us more.

msal/tests/Test.MSAL.NET.Unit.net45/TestCommon.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        }
+
+        public static void MockInstanceDiscoveryAndOpenIdRequest(MockHttpManager mockHttpManager)
+        {
+            mockHttpManager.AddInstanceDiscoveryMockHandler();
+            mockHttpManager.AddMockHandlerForTenantEndpointDiscovery(MsalTestConstants.AuthorityHomeTenant);
+        }
+
+    }
```
jennyf19, line 39: nice.

jennyf19 (approved):
:shipit:

bgavrilMS:
Merged manually

PR #1379, opened 2018-11-02
Enable only embedded webview (not sure why both are getting creating)…

jennyf19:
… for AppCenter UITesting


henrik-me (approved):
:shipit:

PR #1383, opened 2018-11-03
workaround for b2clogin.com  with (dead) fish

jennyf19:
🐟 🎣 🍥


core/src/Instance/B2CAuthority.cs
```diff
@@ -37,10 +37,15 @@ internal class B2CAuthority : AadAuthority
...
         public const string Prefix = "tfp"; // The http path of B2C authority looks like "/tfp/<your_tenant_name>/..."
         public const string B2CCanonicalAuthorityTemplate = "https://{0}/{1}/{2}/{3}/";
+        public const string MicrosoftOnline = "https://login.microsoftonline.com";
+        public const string OpenIdConfigurationEndpoint = "v2.0/.well-known/openid-configuration";
 
         internal B2CAuthority(string authority, bool validateAuthority) 
             : base(authority, validateAuthority)
         {
+            AuthorityType = AuthorityType.B2C;
+            ValidateAuthority = validateAuthority;
```
henrik-me: >validateAuthority [](start = 32, length = 17)

We should consider setting this to false always for b2c.... #Resolved

henrik-me:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/669
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/632 #Closed

henrik-me:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/632

core/src/Instance/B2CAuthority.cs
```diff
@@ -50,8 +55,15 @@ internal B2CAuthority(string authority, bool validateAuthority)
 
             CanonicalAuthority = string.Format(CultureInfo.InvariantCulture, B2CCanonicalAuthorityTemplate, authorityUri.Authority,
                 pathSegments[0], pathSegments[1], pathSegments[2]);
+        }
 
-            AuthorityType = AuthorityType.B2C;
+        protected override string GetDefaultOpenIdConfigurationEndpoint()
+        {
+            Uri b2cAuthority = new Uri(CanonicalAuthority);
```
henrik-me, line 85: Should be validated with a non b2clogin.com domain/setup #Closed

core/src/Instance/B2CAuthority.cs
```diff
@@ -50,8 +55,15 @@ internal B2CAuthority(string authority, bool validateAuthority)
...
+
+            // Ignoring sovereign clouds
```
henrik-me: >// Ignoring sovereign clouds [](start = 12, length = 28)

to support host we would need to make the call to the tenant discovery end point, which is not supported for b2clogin.com #Closed

henrik-me:
utests are missing/not updated. 33 tests are failing.
To minimize risk we should:
 - override UpdateCanonicalAuthorityAsync in B2CAuthority, ensuring the DoInstanceDiscoveryAndCacheAsync is called with the correct Authority (which can actually do the tenant lookup)
 - revert the change for DoInstanceDiscoveryAndCacheAsync

core/src/Instance/AadAuthority.cs
```diff
@@ -52,6 +52,8 @@ internal class AadAuthority : Authority
             "login.cloudgovapi.us" // Microsoft Azure US Government
         };
 
+        internal static string B2CTrustedHost = "b2clogin.com";
```
henrik-me: > internal static string B2CTrustedHost = "b2clogin.com"; [](start = 7, length = 56)

ideally we would have this in the B2CAuthotity class and then we have a virtual method which we override there which does the B2Clogin addition to the validation. #Resolved

core/src/Instance/AadAuthority.cs
```diff
@@ -118,8 +125,11 @@ protected override string GetDefaultOpenIdConfigurationEndpoint()
 
         internal static bool IsInTrustedHostList(string host)
         {
-            return !string.IsNullOrEmpty(
-                       TrustedHostList.FirstOrDefault(a => string.Compare(host, a, StringComparison.OrdinalIgnoreCase) == 0));
+            bool isInList = 
+                !string.IsNullOrEmpty(
+                    TrustedHostList.FirstOrDefault(a => string.Compare(host, a, StringComparison.OrdinalIgnoreCase) == 0));
+            isInList |= host.EndsWith(B2CTrustedHost);
```
henrik-me: >isInList |= host.EndsWith(B2CTrustedHost); [](start = 12, length = 42)

should be moved to the B2CAuthority... could also make the additional check in the line above, also what about other custom domains... not sure if it will work or not (though this PR is only for working around b2clogin.com). #Resolved

henrik-me (changes requested):
:clock1:

jennyf19:
done. There are a few more utests failing, and i added a few more. have some more work to do here but manual testing seems to be working. might have a few questions for you on Monday.

---
In reply to: [435548715](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1383#issuecomment-435548715) [](ancestors = 435548715)

core/src/Instance/B2CAuthority.cs
```diff
@@ -37,37 +38,85 @@ internal class B2CAuthority : AadAuthority
...
             }
+            
+            var metadata = await AadInstanceDiscovery
+                             .Instance.GetMetadataEntryAsync(
+                                 httpManager,
+                                 telemetryManager,
+                                 new Uri(UpdateCanonicalAuthorityB2C()),
+                                 ValidateAuthority,
+                                 requestContext)
+                             .ConfigureAwait(false);
```
henrik-me: doesn't this fail for the b2clogin.com ?

core/src/Instance/B2CAuthority.cs
```diff
@@ -37,37 +38,85 @@ internal class B2CAuthority : AadAuthority
...
+
+        private string UpdateCanonicalAuthorityB2C()
+        {
+            Uri b2cAuthority = new Uri(CanonicalAuthority);
+
+            string canonicalAuthorityUri = string.Format(CultureInfo.InvariantCulture, MicrosoftOnline + b2cAuthority.AbsolutePath);
+            return canonicalAuthorityUri;
+        }
 
+        internal new static bool IsInTrustedHostList(string host)
```
henrik-me: >internal new static [](start = 8, length = 19)

not a fan of this construct. I suggest adding an overload or enabling passing in an additional delegate to the IsInTrustedHostlist, alternatively given the code is called from here it's possible to simply create a new method. #Resolved

msal/tests/Test.MSAL.NET.Common/Core/Helpers/CoreTestConstants.cs
```diff
@@ -46,6 +46,8 @@ class CoreTestConstants
         public static readonly string PrefCacheAuthorityCommonTenant = "https://" + ProductionPrefCacheEnvironment + "/common/";
         public static readonly string AuthorityOrganizationsTenant = "https://" + ProductionPrefNetworkEnvironment + "/organizations/";
         public const string B2CAuthority = "https://login.microsoftonline.in/tfp/tenant/policy/";
+        public const string B2CLoginAuthority = "https://sometenantid.b2clogin.com/tfp/sometenantid/policy/";
+        public const string B2CInvalidAuthority = "https://sometenantid.b2blogin.com/tfp/sometenantid/policy/";
```
henrik-me: >B2CInvalidAuthority [](start = 28, length = 19)

these 2 authorities looks very similar, not sure why one is Invalid  #Resolved

msal/tests/Test.MSAL.NET.Unit.net45/CoreTests/InstanceTests/B2cAuthorityTests.cs
```diff
@@ -80,29 +85,40 @@ public void NotEnoughPathSegmentsTest()
...
+                    Task.Run(
+                        async () =>
+                        {
+                            await instance.ResolveEndpointsAsync(
+                                httpManager,
+                                new TelemetryManager(),
+                                null,
+                                new RequestContext(null, new TestLogger(Guid.NewGuid(), null))).ConfigureAwait(false);
+                        }).GetAwaiter().GetResult();
+                }
```
henrik-me, line 98: >} [](start = 16, length = 1)

seems like you are expecting an exception. 
1: Please only catch the exception type you are expecting.
2: Please ensure to Assert.Fail, for the "happy" scenario as that is not expected. (the code you cleaned-up had this) #Resolved

henrik-me:
Generally looks great. Thanks a lot for looking into this. Have one minor comment, which might be easier to discuss in person.

---
In reply to: [435641720](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1383#issuecomment-435641720) [](ancestors = 435641720,435548715)

msal/tests/Test.MSAL.NET.Unit.net45/CoreTests/InstanceTests/B2cAuthorityTests.cs
jennyf19: done. we are checking the authority earlier now, so needed to use a different method to get the same result here.

---
In reply to: [230837184](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1383#discussion_r230837184) [](ancestors = 230837184)

msal/tests/Test.MSAL.NET.Common/Core/Helpers/CoreTestConstants.cs
jennyf19: i fixed the name... this one is not in the list of trusted hosts, so if you were to pass ValidateAuthority=true, it would fail. We can discuss this further if you want. PS. this one has b2b instead of b2c.

---
In reply to: [230836424](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1383#discussion_r230836424) [](ancestors = 230836424)

core/src/Instance/B2CAuthority.cs
jennyf19: made a new method instead.

---
In reply to: [230835339](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1383#discussion_r230835339) [](ancestors = 230835339)

core/src/Instance/B2CAuthority.cs
```diff
@@ -37,37 +38,85 @@ internal class B2CAuthority : AadAuthority
     {
         public const string Prefix = "tfp"; // The http path of B2C authority looks like "/tfp/<your_tenant_name>/..."
         public const string B2CCanonicalAuthorityTemplate = "https://{0}/{1}/{2}/{3}/";
+        public const string MicrosoftOnline = "https://login.microsoftonline.com";
+        public const string OpenIdConfigurationEndpoint = "v2.0/.well-known/openid-configuration";
+        internal const string B2CTrustedHost = "b2clogin.com";
```
bgavrilMS: All the other constants are public, keep this one public as well. It doesn't really matter if a member is public or internal if the class is internal, so consistecy trumps. I'd recommend using public inside internal classes.  #Resolved

core/src/Instance/B2CAuthority.cs
```diff
@@ -37,37 +38,85 @@ internal class B2CAuthority : AadAuthority
...
+            ITelemetryManager telemetryManager,
             RequestContext requestContext)
         {
-            if (ValidateAuthority && !IsInTrustedHostList(new Uri(CanonicalAuthority).Host))
+            if (ValidateAuthority && !InTrustedHostList(new Uri(CanonicalAuthority).Host))
             {
                 throw new ArgumentException(CoreErrorMessages.UnsupportedAuthorityValidation);
             }
+            
+            var metadata = await AadInstanceDiscovery
```
bgavrilMS, line 71: Doesn't calling base.UpdateCannonicalAuthorityAsync do the same thing?

core/src/Instance/B2CAuthority.cs
```diff
@@ -37,37 +38,85 @@ internal class B2CAuthority : AadAuthority
...
+                                 telemetryManager,
+                                 new Uri(UpdateCanonicalAuthorityB2C()),
+                                 ValidateAuthority,
+                                 requestContext)
+                             .ConfigureAwait(false);
+
+            CanonicalAuthority = UpdateHost(CanonicalAuthority, metadata.PreferredNetwork);
+        }
+
+        private string UpdateCanonicalAuthorityB2C()
```
bgavrilMS: This method name does not match what it is doing. This is a "query" type of method, i.e. it returns some info without changing any state, but you named it as a "command" method, i.e. it changes state. 

Also, I think this can be static since it only uses some constants from this class, i.e. it doesn't use state. #Resolved

msal/tests/Test.MSAL.NET.Unit.net45/CoreTests/InstanceTests/B2cAuthorityTests.cs
```diff
@@ -80,26 +85,28 @@ public void NotEnoughPathSegmentsTest()
...
+            using (var httpManager = new MockHttpManager())
             {
-                Assert.IsInstanceOfType(exc, typeof(ArgumentException));
-                Assert.AreEqual(CoreErrorMessages.UnsupportedAuthorityValidation, exc.Message);
+                var instance = Authority.CreateAuthority(CoreTestConstants.B2CAuthority, true);
+                Assert.IsNotNull(instance);
+                Assert.AreEqual(instance.AuthorityType, AuthorityType.B2C);
+                try
+                {
+                    Task.Run(
```
bgavrilMS: can't you do instance.UpdateCannonicalAuthorityAsync(...).GetAwaiter().GetResult() instead? i.e. without the Task.Run ? #Resolved

jennyf19: i can. thanks/

---
In reply to: [230921547](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1383#discussion_r230921547) [](ancestors = 230921547)

core/src/Instance/B2CAuthority.cs
jennyf19: we have to change the canonical authority to get the metadata back because this is not supported in b2c yet. 

---
In reply to: [230919572](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1383#discussion_r230919572) [](ancestors = 230919572)

core/src/Instance/B2CAuthority.cs
jennyf19: this was copy/paste, so thanks for picking it up. 

---
In reply to: [230918925](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1383#discussion_r230918925) [](ancestors = 230918925)

core/src/Instance/B2CAuthority.cs
jennyf19: what name would you suggest? it is changing the value of the canonical authority, because we need to replace the host w/microsoftonline.com for the b2clogin.com scenarios.

---
In reply to: [230920739](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1383#discussion_r230920739) [](ancestors = 230920739)

bgavrilMS (approved):

core/src/Instance/B2CAuthority.cs
```diff
@@ -37,37 +38,85 @@ internal class B2CAuthority : AadAuthority
...
+
+            string canonicalAuthorityUri = string.Format(CultureInfo.InvariantCulture, MicrosoftOnline + b2cAuthority.AbsolutePath);
+            return canonicalAuthorityUri;
+        }
+
+        private bool InTrustedHostList(string host)
+        {
+            bool isInList =
+                !string.IsNullOrEmpty(
+                    TrustedHostList.FirstOrDefault(a => string.Compare(host, a, StringComparison.OrdinalIgnoreCase) == 0));
```
henrik-me: consider replacing this with the method which already exists on AadAuthority. Also you should probably combine the two comparisons into one "if" statement to make it conditional if both conditions should be evaluated (it will have a few CPU cycles). #Resolved

core/src/Instance/B2CAuthority.cs
bgavrilMS: GetCannonicalAuthorityB2C

---
In reply to: [230939266](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1383#discussion_r230939266) [](ancestors = 230939266,230920739)

core/src/Instance/B2CAuthority.cs
jennyf19: yes...good feedback. thanks.

---
In reply to: [231014061](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1383#discussion_r231014061) [](ancestors = 231014061)

core/src/Instance/B2CAuthority.cs
jennyf19: i thought of that name too, but wasn't sure, so i'll go w/it since you suggested it too. :)

---
In reply to: [231095905](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1383#discussion_r231095905) [](ancestors = 231095905,230939266,230920739)

henrik-me (approved):
:shipit:

core/src/Instance/B2CAuthority.cs
```diff
@@ -37,37 +38,84 @@ internal class B2CAuthority : AadAuthority
...
+            return authorityUri;
+        }
+
+        protected override async Task<string> GetOpenIdConfigurationEndpointAsync(
+            IHttpManager httpManager,
+            ITelemetryManager telemetryManager,
+            string userPrincipalName,
+            RequestContext requestContext)
+        {
             return await Task.Run(() => GetDefaultOpenIdConfigurationEndpoint()).ConfigureAwait(false);
```
dcarr42, line 116: Task.FromResult maybe better here.

bgavrilMS: Yes, thanks for picking up on that @dallancarr . 

@jennyf19  - Task.FromResult will wrap the result in a Task object, while Task.Run will actually dispatch to a worker thread, fetch the result etc. Same result, but with better perf. 

https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1401

PR #1385, opened 2018-11-05
Switching ui tests to use appids supplied by the lab.

trwalke:
Adding dismiss keyboard to enterText method in test controller.

Changing determineUser to detrmineFieldIds for extra clarity.


core/tests/Test.Microsoft.Identity.LabInfrastructure/LabServiceApi.cs
```diff
@@ -82,7 +82,10 @@ private IUser GetUserFromLab(UserQueryParameters query)
                 throw new LabUserNotFoundException(query, "No lab user with specified parameters exists");
             }
 
-            LabUser user = JsonConvert.DeserializeObject<LabResponse>(result).Users;
+            LabResponse response = JsonConvert.DeserializeObject<LabResponse>(result);
+
+            LabUser user =  response.Users;
+            user.AppId = response.AppId;
```
henrik-me: >user.AppId = response.AppId; [](start = 12, length = 28)

it's a bit odd that the appId is tied to the user. Having a parent object which holds a user and the app id feels more natural to me.  #Closed

core/tests/Test.Microsoft.Identity.Core.UIAutomation/UserInformationFieldIds.cs
```diff
@@ -34,7 +34,7 @@ public class UserInformationFieldIds
         public string PasswordInputId { get; set; }
         public string SignInButtonId { get;  set; }
 
-        public void DetermineUser(IUser user)
+        public void DetermineFieldIds(IUser user)
```
henrik-me: Agree on the field id's, perhaps UpdateFieldIds ? #Closed

adal/tests/Test.ADAL.NET.UIAutomation/ADALMobileTestHelper.cs
```diff
@@ -64,25 +64,25 @@ public void AcquireTokenSilentTestHelper(ITestController controller, UserQueryPa
...
-            controller.DismissKeyboard();
 
             //Acquire token silently
             controller.Tap(CoreUiTestConstants.AcquireTokenSilentID);
 
             CoreMobileTestHelper.VerifyResult(controller);
         }
 
-        public void AcquireTokenInteractiveHelper(ITestController controller, UserQueryParameters userParams)
+        public IUser AcquireTokenInteractiveHelper(ITestController controller, UserQueryParameters userParams)
```
henrik-me: >IUser [](start = 15, length = 5)

Is the user used anywhere? if it is pls. keep it. If not used anywhere, then make it void again. #Closed

henrik-me (changes requested):
:clock1:

adal/tests/Test.ADAL.NET.UIAutomation/ADALMobileTestHelper.cs
bgavrilMS: >AcquireTokenInteractiveHelper [](start = 21, length = 29)

There is an architectural flaw with our tests, which you may or may not want to fix as part of this PR. To fetch a user you need a test controller, and we bundle it with pressing "clear cache" etc. This violates SRP, makes it difficult to debug failing tests etc. The test structure should be more in the lines of: 

1. Get a user from the lab. Optionally fetch its password too. Fail fast if you can't. 
2. Do the test (start pressing buttons etc.) #Closed

henrik-me: Anyone can make this change while updating test code :) #Closed

core/tests/Test.Microsoft.Identity.Core.UIAutomation/UserInformationFieldIds.cs
trwalke: I feel DetermineFieldIds is more clear as we are determining which ids to use rather than updating them #Pending

adal/tests/Test.ADAL.NET.UIAutomation/ADALMobileTestHelper.cs
trwalke: I made the changes. please see latest commit. #Closed

bgavrilMS (approved):

bgavrilMS:
I played a bit with the changes and realized that you are not setting the return uri properly. Let's sync as I'd like to push a commit on top of this to fix a few things.

adal/tests/Test.ADAL.NET.UIAutomation/ADALMobileTestHelper.cs
```diff
@@ -47,85 +47,80 @@ public class ADALMobileTestHelper
...
-            controller.DismissKeyboard();
 
             //Acquire token silently
             controller.Tap(CoreUiTestConstants.AcquireTokenSilentID);
 
             CoreMobileTestHelper.VerifyResult(controller);
         }
 
-        public void AcquireTokenInteractiveHelper(ITestController controller, UserQueryParameters userParams)
+        public IUser AcquireTokenInteractiveHelper(ITestController controller, ILabResponse userData)
```
henrik-me: >userData [](start = 92, length = 8)

rename: labResponse?
- same below #Closed

adal/tests/Test.ADAL.NET.UIAutomation/XamarinDroidADALTests.cs
```diff
@@ -79,7 +79,7 @@ public void AcquireTokenInteractiveTest()
         [Test]
         public void AcquireTokenSilentTest()
         {
-            _adalMobileTestHelper.AcquireTokenSilentTestHelper(xamarinController, LabUserHelper.DefaultUserQuery);
+            _adalMobileTestHelper.AcquireTokenSilentTestHelper(xamarinController, LabUserHelper.GetUser(LabUserHelper.DefaultUserQuery));
```
henrik-me: >LabUserHelper.GetUser [](start = 82, length = 21)

Consider adding a LabUserHelper.GetDefaultUser, this construct seems a bit convoluted and it does many things. 
Note: given that it's likely more returning a labResponse, then the mehtod name should probably be slightly different. 
E.g.: GetLabResponseWithDefaultUser #Closed

adal/tests/Test.ADAL.NET.UIAutomation/XamarinDroidADALTests.cs
```diff
@@ -103,7 +103,7 @@ public void AcquireTokenADFSv4FederatedInteractiveTest()
             user.FederationProvider = FederationProvider.AdfsV4;
             user.IsFederatedUser = true;
```
henrik-me: You could convert this into a method on the helper, E.g. GetLabResponseWithADFSV4User ... or similar (same for other user types) #Closed

core/tests/Test.Microsoft.Identity.Core.UIAutomation/UserInformationFieldIds.cs
henrik-me: ok, not sure determine is the right word though, but I agree better than update. Perhaps SetFieldIdsFromUser or simply SetFieldIds

---
In reply to: [231373212](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1385#discussion_r231373212) [](ancestors = 231373212)

core/tests/Test.Microsoft.Identity.LabInfrastructure/ILabService.cs
```diff
@@ -31,6 +31,6 @@ namespace Test.Microsoft.Identity.LabInfrastructure
 {
     public interface ILabService
     {
-        IUser GetUser(UserQueryParameters query);
+        ILabResponse GetUser(UserQueryParameters query);
```
henrik-me: >GetUser [](start = 21, length = 7)

GetLabResponseFromUser ... or we need to find something better.... it's not just getting the user now. #Closed

core/tests/Test.Microsoft.Identity.LabInfrastructure/LabServiceApi.cs
```diff
@@ -91,17 +93,18 @@ private IUser GetUserFromLab(UserQueryParameters query)
...
         /// Returns a test user account for use in testing.
         /// </summary>
         /// <param name="query">Any and all parameters that the returned user should satisfy.</param>
         /// <returns>Users that match the given query parameters.</returns>
-        public IUser GetUser(UserQueryParameters query)
+        public ILabResponse GetUser(UserQueryParameters query)
         {
-            var user = GetUserFromLab(query) as LabUser;
+            var response = GetUserFromLab(query);
+            var user = response.User as LabUser;
```
henrik-me: > response.User as LabUser [](start = 22, length = 25)

Consider making the response return the type as LabUser instead, any concerns about that? Or rather why isn't it possible to use the Interface, perhaps it needs to be updated? #Closed

trwalke:
Merging bogdans PR in. will address all comments with his changes included #Pending

henrik-me (approved):
:shipit:

PR #1387, opened 2018-11-06
remove process-wide ValidatedAuthoritiesCache and AadInstanceDiscover…

MarkZuber:
…y cache


bgavrilMS:
The code looks good, but I may be lacking context on why this change is required and the performance implications. I'm particularly concerned about the ConfidentialClient flows in WebApp / WebApi scenarios, where the current guidance is to use the user session for token caches - as such, people create a new ConfidentialClient for each user. Or worse, for each acquireToken* call. This means that web servers will perform an extra webrequest for each user / acquireToken call ?

See [https://github.com/Azure-Samples/active-directory-dotnet-admin-restricted-scopes-v2/blob/master/GroupManager/Controllers/UsersController.cs#L65](https://github.com/Azure-Samples/active-directory-dotnet-admin-restricted-scopes-v2/blob/master/GroupManager/Controllers/UsersController.cs#L65)

core/src/MsalIdHelper.cs
```diff
@@ -35,83 +35,103 @@ namespace Microsoft.Identity.Core
...
+                {
+                    parameters[MsalIdParameter.DeviceModel] = deviceInformation;
+                }
+
+                return parameters;
+            });
 
-            var deviceInformation = platformProxy.GetDeviceModel();
-            if (deviceInformation != null)
+        private static readonly Lazy<string> MsalVersion = new Lazy<string>(
```
henrik-me, line 107: >private static readonly Lazy<string> MsalVersion = new Lazy<string>( [](start = 8, length = 68)

:)

core/src/MsalIdHelper.cs
```diff
@@ -35,83 +35,103 @@ namespace Microsoft.Identity.Core
...
+                    StringSplitOptions.None);
+                return version[1];
+            });
+
+        public static IDictionary<string, string> GetMsalIdParameters()
+        {
+            return MsalIdParameters.Value;
         }
 
         public static string GetMsalVersion()
```
henrik-me, line 132: >GetMsalVersion [](start = 29, length = 14)

I suppose putting this in core will cause challenges? Ideally I would have liked to see this one in the PlatformProxy or similar.

core/src/MsalIdHelper.cs
```diff
@@ -35,83 +35,103 @@ namespace Microsoft.Identity.Core
...
+    ///     This class adds additional query parameters or headers to the requests sent to STS. This can help us in
+    ///     collecting statistics and potentially on diagnostics.
     /// </summary>
     internal static class MsalIdHelper
     {
-        public static IDictionary<string, string> GetMsalIdParameters()
-        {
-            var platformProxy = PlatformProxyFactory.GetPlatformProxy();
-            if (platformProxy == null)
+        private static readonly Lazy<IDictionary<string, string>> MsalIdParameters = new Lazy<IDictionary<string, string>>(
```
henrik-me, line 69: >new Lazy<IDictionary<string, string> [](start = 85, length = 36)

:)

core/src/MsalIdHelper.cs
```diff
@@ -35,83 +35,103 @@ namespace Microsoft.Identity.Core
     internal static class MsalIdParameter
     {
         /// <summary>
-        /// MSAL Flavor: .NET or WinRT
+        ///     MSAL Flavor: .NET or WinRT
```
henrik-me, line 38: >      [](start = 11, length = 5)

not sure I understand the new indents we are getting... I'm pretty sure my VS is not setup to do just that #Closed

msal/src/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -105,9 +107,11 @@ internal ITelemetryReceiver TelemetryReceiver
             CryptographyManager = PlatformProxyFactory.GetPlatformProxy().CryptographyManager;
             WsTrustWebRequestManager = new WsTrustWebRequestManager(HttpManager);
             TelemetryManager = telemetryManager ?? new TelemetryManager(Telemetry.GetInstance());
+            ValidatedAuthoritiesCache = new ValidatedAuthoritiesCache();
```
henrik-me: >ValidatedAuthoritiesCache = new ValidatedAuthoritiesCache(); [](start = 12, length = 60)

as discussed and to Bogdan's point, it might be useful to identify a way for the user to enable sharing. 
However we will in that case though also have to ensure we do cache refresh at the appropriate time (if people are keeping e.g. servers running for a very long time with out flushing memory/restaring/etc.)

henrik-me:
The way I see this is. Removal of statics and some additional focus on SRP (espcially for not having the http object where it shouldn't be, or just the S in SOLID). 
Agree there is a side effect of this change regarding the Authority validation cache. I suppose potentially also regarding the openidconnect discovery end point. 
Not sure we should store state between object creations as it's handled today. It could be a developer option, which was on by default (assuming we want to stick to not breaking behavior). Alternatively we can recommend a different approach for web services for re-using the objects (for most IIS processes though they could be persisting in memory between calls by leveraging the session cache.

---
In reply to: [436363256](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1387#issuecomment-436363256) [](ancestors = 436363256)

henrik-me:
Concern + Suggestion: too much stuff which needs to be passed around. Perhaps we can have one object which holds all the stuff which are passed around (an object which holds all our injected objects such as 
            IHttpManager, ICryptographyManager, ITelemetryManager, IValidatedAuthoritiesCache, IAadInstanceDiscovery, AuthenticationRequestParameters, ApiEvent.ApiIds
Perhaps the DependencyInjectionContainer :)

core/src/MsalIdHelper.cs
MarkZuber: There are several things that can move into PlatformProxy but will screw up ADAL or potentially bring in code into ADAL that it doesn't want/need.  I'm holding on those until after "the split(tm)"

MarkZuber:
> The code looks good, but I may be lacking context on why this change is required and the performance implications. I'm particularly concerned about the ConfidentialClient flows in WebApp / WebApi scenarios, where the current guidance is to use the user session for token caches - as such, people create a new ConfidentialClient for each user. Or worse, for each acquireToken* call. This means that web servers will perform an extra webrequest for each user / acquireToken call ?
> 
> See https://github.com/Azure-Samples/active-directory-dotnet-admin-restricted-scopes-v2/blob/master/GroupManager/Controllers/UsersController.cs#L65

I'm going to fix this for now by moving the ConcurrentDictionary instances within AadInstanceDiscovery and ValidatedAuthoritiesCache to be static members.  This allows this PR to continue and decouple the static behavior everywhere else, but maintains existing behavior of process-wide caching of these until we get a better story here.

core/src/MsalIdHelper.cs
MarkZuber: R# #Closed

core/src/Instance/IAadInstanceDiscovery.cs
```diff
@@ -0,0 +1,48 @@
...
+// THE SOFTWARE.
+// 
+// ------------------------------------------------------------------------------
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Core.Instance
+{
+    internal interface IAadInstanceDiscovery
```
henrik-me, line 33: consider using non AAD prefixes here. I would assume that we would need to do something specific for B2C and perhaps even for ADFS. That way we can easily implement a class for each of those authority types.

henrik-me:
@jennyf19 : pls. see the changes done to Authorities.

core/src/Instance/AadInstanceDiscovery.cs
```diff
@@ -84,51 +115,22 @@ public static string BuildInstanceDiscoveryEndpoint(string host)
...
+                                       ? authority.Host
+                                       : AadAuthority.DefaultTrustedHost;
 
             string instanceDiscoveryEndpoint = BuildInstanceDiscoveryEndpoint(discoveryHost);
 
-            InstanceDiscoveryResponse discoveryResponse =
-                await
-                    client.DiscoverAadInstanceAsync(new Uri(instanceDiscoveryEndpoint), requestContext)
-                        .ConfigureAwait(false);
+            var discoveryResponse = await client.DiscoverAadInstanceAsync(new Uri(instanceDiscoveryEndpoint), requestContext)
```
jennyf19, line 132: >var [](start = 12, length = 3)

Is there a reason to use `var` here and not `InstanceDiscoveryResponse`?

msal/src/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -59,8 +59,6 @@ public sealed class TokenCache
         internal const string NullPreferredUsernameDisplayLabel = "Missing from the token response";
         private const string MicrosoftLogin = "login.microsoftonline.com";
```
henrik-me, line 60: >MicrosoftLogin [](start = 29, length = 14)

this is also in B2C authority. perhaps it should be a const on the Authority object instead?

henrik-me (approved):
:shipit:

msal/src/Microsoft.Identity.Client/TokenCache.cs
MarkZuber: I'll look at that when we go to cleanup the caching.

core/src/Instance/AadInstanceDiscovery.cs
MarkZuber: no, other than line width.

PR #1390, opened 2018-11-07
Remove some test interfaces and simplify model

bgavrilMS:


msal/tests/dev apps/XForms/XForms/App.xaml.cs
```diff
@@ -38,12 +38,11 @@ public partial class App : Application
...
         public static PublicClientApplication MsalPublicClient;
         public static UIParent UIParent { get; set; }
-        public const string DefaultClientId = "5a434691-ccb2-4fd1-b97b-b64bcfbc03fc";
+        public const string DefaultClientId = "4b0db8c2-9f26-4417-8bde-3f0e3656f8e0";
         public const string B2cClientId = "750a8822-a6d4-4127-bc0b-efbfacccbc28";
 
-        public const string RedirectUriOnAndroid =
-            "msauth-5a434691-ccb2-4fd1-b97b-b64bcfbc03fc://com.microsoft.identity.client.sample";
         public const string RedirectUriOnIos = "adaliosxformsapp://com.yourcompany.xformsapp";
+        public const string RedirectUriOnAndroid = "urn:ietf:wg:oauth:2.0:oob";
```
jennyf19, line 45: will this work for android?

jennyf19 (approved):

msal/tests/dev apps/XForms/XForms/App.xaml.cs
trwalke: yes. this is currently being used for all of our mobile adal android tests

msal/tests/Test.MSAL.NET.UIAutomation/XamarinDroidMSALTests.cs
```diff
@@ -69,7 +69,8 @@ public void InitializeBeforeTest()
         [Test]
         public void AcquireTokenTest()
         {
-            _msalMobileTestHelper.AcquireTokenInteractiveTestHelper(xamarinController, LabUserHelper.GetUser(LabUserHelper.DefaultUserQuery));
+            var labResponse = LabUserHelper.GetLabUserAndApp(LabUserHelper.DefaultUserQuery);
+            _msalMobileTestHelper.AcquireTokenInteractiveTestHelper(xamarinController, labResponse);
```
trwalke, line 73: can we change this to LabUserHelper.GetLabUserAndApp(LabUserHelper.DefaultUserQuery) like the other functions?

msal/tests/dev apps/XForms/XForms/App.xaml.cs
```diff
@@ -38,12 +38,11 @@ public partial class App : Application
     {
         public static PublicClientApplication MsalPublicClient;
         public static UIParent UIParent { get; set; }
-        public const string DefaultClientId = "5a434691-ccb2-4fd1-b97b-b64bcfbc03fc";
+        public const string DefaultClientId = "4b0db8c2-9f26-4417-8bde-3f0e3656f8e0";
```
trwalke, line 41: Where is this appId from?

core/tests/Test.Microsoft.Identity.LabInfrastructure/LabUserHelper.cs
```diff
@@ -56,7 +56,7 @@ public static UserQueryParameters DefaultUserQuery
             }
         }
 
-        public static ILabResponse GetUser(UserQueryParameters query)
+        public static LabResponse GetLabUserAndApp(UserQueryParameters query)
```
trwalke, line 59: How about GetLabUserData? right now, we are only using the appid from the Lab response but there are many more properties that we may end up using. (like tenantId)

trwalke (changes requested):
Few questions on minor things

msal/tests/Test.MSAL.NET.UIAutomation/XamarinDroidMSALTests.cs
bgavrilMS: When people first try out to run automation, they will get into problems because installing that certificate is a pain and IWA doesn't work because of MFA. I will change it, but it's not a bad idea to separate the 2.

trwalke (approved):

msal/tests/dev apps/XForms/XForms/App.xaml.cs
bgavrilMS: It's the default user query app.

core/tests/Test.Microsoft.Identity.LabInfrastructure/LabUserHelper.cs
bgavrilMS: Sure, will do. The main refactoring was removing some of the interfaces, because they were not really doing much but were complicating things when it came to fetch the password etc.

PR #1391, opened 2018-11-07
Jennyf/ios

jennyf19:
Remove extra callback...causes problems are reported with this [issue](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/672). I verified, as did the customer, that this fix addresses the problem. Does not affect other cancellation operations on iOS. Would be nice to get in next release.


henrik-me (approved):
:shipit:

Issue #1397, opened 2018-11-09
Xamarin iOS support for HttpClient handler


ghost:
**Which Version of ADAL are you using ?**
ADAL 3.19.8
**Which platform has the issue?**
iOS

**What authentication flow has the issue?**
* Desktop / Mobile
    * [X] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Repro**
The question is does the HttpClientFactory for Xamarin iOS use the standard .NET Http Client Handler, or does it use what the client app has set for the http client handler in the Xamarin Project setup?  Specifically if I built a new app and use this nuget package if I have NSUrlSession set as the supported Http Client is the ADAL code using this or is it using the .NET Library.

**Expected behavior**
What ever the project has set as the Http Client that the nuget package uses that.

**Actual behavior**
We believe the http client is the .NET Managed Http Client because we are seeing BSD socket requests instead of standard HTTP traffic when the token is requested.

**Possible Solution**
Honor the settings that the client has or allow the injection of the Http Client Handler so that the developer can override the http client handler with a specific on that implements NSUrlSession/NSUrlConnectionHandler

**Additional context/ Logs / Screenshots**
![screen shot 2018-11-09 at 12 05 09 pm](https://user-images.githubusercontent.com/176927/48277465-c594a000-e418-11e8-9f98-d88b3fce127d.png)

henrik-me:
You might have to look at the proxy settings: https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/wiki/with-proxy
Note: For ADAL v4 the http stack is different and may behave differently, though under the hood it's still using the HttpClient.

henrik-me:
@jmprieur : Seems like a feature ask we have discussed previously regarding the HttpHandler.

ghost:
this is correct - we would like this as a feature.

henrik-me:
@biozal : The library is currently using the std. net HttpHandler as you probably already had found. We will investigate what it will take to offer custom HttpHandlers.

ghost:
[...]

PR #1399, opened 2018-11-09
Allow users to set HttpMessageHandler

bgavrilMS:
This is related to https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1397. This PR is incomplete (needs the same on MSAL and needs unit tests), but I'd like to an input on the approach / public API.

### DIFF
--- a/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
+++ b/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
@@ -26,6 +26,7 @@
 //------------------------------------------------------------------------------
 
 using System;
+using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.Identity.Core;
 using Microsoft.Identity.Core.Cache;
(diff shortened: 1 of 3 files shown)


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -119,6 +120,15 @@ public AuthenticationContext(string authority, bool validateAuthority, TokenCach
             _wsTrustWebRequestManager = new WsTrustWebRequestManager(_httpManager);
         }
 
+        /// <summary>
+        /// Provides a mechanism for adding custom message handlers in the HttpClient used by this library.
+        /// </summary>
+        public void UseCustomHttpMessageHandler(HttpMessageHandler httpMessageHandler)
```
bgavrilMS, line 126: This is the public API change. I don't want to pass it through a ctor, because we already have a ton of ctors, but can do.

MarkZuber: The concern I have with this is technically the user could change this at any time and we will likely have passed the _httpManager around since it's a dependency injection and this effectively changing the graph.  We should add this as an optional to the constructor(s) instead of this or we'll likely need more work to ensure this gets used consistently.

Better API for this (to think about for msal, not here) would be to have a builder or the config object for setting this up.

MarkZuber (changes requested):
see comment about DI/constructors

henrik-me:
Discussed to open up for the HttpClient instead of just the handler. Perhaps also something we would want to do in MSAL instead. The idea that this belongs to the config object makes sense to me... it seems to be about time we get that one prioritized.

henrik-me (changes requested):
:clock1:

bgavrilMS:
Will look at this in MSAL for now

Issue #1400, opened 2018-11-11
AcquireTokenAsync for UWP crashes on null redirectUri parameter


jevansaks:
**Which Version of ADAL are you using ?**
ADAL 4.3.0

**Which platform has the issue?**
UWP

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Repro**

```csharp
  var authResult = await authCtx.AcquireTokenAsync(_resourceId, _clientId, null, platformParameters);
```

**Expected vs actual behavior**
Expected: it works.
Actual: it crashes.

The comments and code make it seem like I'm supposed to be able to pass in null for redirectUri and then have the code eventually call WebAuthenticationBroker.AuthenticateAsync without a redirectUri (implying that my app's ms-app:// URL will be used). 

However, what happens is that the code crashes immediately in AcquireTokenAsync after calling ValidateRedirectUri because it tries to dereference redirectUri.

https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/96 (bullet 3 in the list) says that null is accepted for callbackUri, so it seems like it's expected to work.

**Possible Solution**
What I *think* is wrong is that ValidateRedirectUri is written as if the assignment to redirectUri will affect the caller, but it will not. This line: https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/e47ba877f3da1fbd7e3821e456cf33dcf26906c5/core/src/Platforms/uap/UapPlatformProxy.cs#L169

If I'm right, ValidateRedirectUri should be taking redirectUri by ref instead of by value and that would make the code work as written.

**Additional context/ Logs / Screenshots**
N/A

bgavrilMS:
This should have been fixed and you can try it out with the nightly build https://www.myget.org/feed/aad-clients-nightly/package/nuget/Microsoft.IdentityModel.Clients.ActiveDirectory/4.3.1-internal32001dev

Passing NULL for redirect URI causes the redirectUri to be "http://sso" on UWP, which means that ADAL will try to use Web Broker. 

Closing this, fix will be available in next release - likely a minor release 4.3.1

jevansaks:
Thanks!

[...]

PR #1402, opened 2018-11-11
Fix normalization of redirect uri on UWP

bgavrilMS:
Small regression on redirect uri, reported here https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1400


core/src/Platforms/iOS/iOSPlatformProxy.cs
```diff
@@ -97,13 +97,17 @@ public void ValidateRedirectUri(Uri redirectUri, RequestContext requestContext)
             {
                 if (Constants.DefaultRedirectUri.Equals(redirectUri.AbsoluteUri, StringComparison.OrdinalIgnoreCase))
                 {
-                    // TODO: Need to use CoreExceptionFactory here...?
-                    //throw new MsalException(MsalError.RedirectUriValidationFailed, "Default redirect URI - " + Constants.DefaultRedirectUri +
-                    //                                                               " cannot be used on iOS platform");
-                    throw new InvalidOperationException(
-                        $"Default redirect URI - {Constants.DefaultRedirectUri} cannot be used on iOS platform");
+                    throw CoreExceptionFactory.Instance.GetClientException(
```
bgavrilMS, line 98: @jmprieur - should we add more defaults? 

Currently, the logic across platforms is as follows: 

- on UWP, ADAL will use "http://sso" if the redirect uri is NULL
- on iOS and Android, MSAL will throw if you are trying to pass NULL or "urn:obo:..." standard uri
- everywhere else, throw if redirect uri is NULL

jmprieur: @bgavrilMS :
- I was not aware of http://sso. Would you know if it has a special signification for WAB?  In any case we cannot change it in ADAL.NET as this would be a breaking change. In MSAL.NET I thought this was "urn:obo:..." ?
- on iOS and Android, for MSAL.NET, I'm pretty sure that the default is msal{clientId}://auth?
- for pure client credential flows we don't need a redirect URI (but we do for all the other confidential client flows). Is there something we can do for not checking the redirect URI in client credentials flows (with app password and certificate)?  See [MSAL#426](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/426)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenInteractiveHandler.cs
```diff
@@ -60,18 +60,10 @@ internal class AcquireTokenInteractiveHandler : AcquireTokenHandlerBase
             UserIdentifier userId, string extraQueryParameters, IWebUI webUI, string claims)
             : base(requestData)
         {
-            PlatformProxyFactory.GetPlatformProxy().ValidateRedirectUri(redirectUri, RequestContext);
-            this.redirectUri = redirectUri;
-
-            if (!string.IsNullOrWhiteSpace(this.redirectUri.Fragment))
```
bgavrilMS, line 66: Moved this validation inside ValidateRedirectUri

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenInteractiveHandler.cs
```diff
@@ -60,18 +60,10 @@ internal class AcquireTokenInteractiveHandler : AcquireTokenHandlerBase
...
-            {
-                throw new ArgumentException(AdalErrorMessage.RedirectUriContainsFragment, "redirectUri");
-            }
+            PlatformProxyFactory.GetPlatformProxy().ValidateRedirectUri(redirectUri);
 
             this.authorizationParameters = parameters;
-
-            this.redirectUriRequestParameter = PlatformProxyFactory.GetPlatformProxy().GetRedirectUriAsString(this.redirectUri, RequestContext);
```
bgavrilMS, line 73: The old name is misleading because on UAP it is actually returing an uri from WEB

core/src/Platforms/iOS/iOSPlatformProxy.cs
bgavrilMS: Thank you @jmprieur, I made more changes to reflect this. Indeed, the defaults are exactly as you say. 
On UAP, ADAL uses WEB so the default uri is "http://sso"

core/src/Platforms/netstandard1.3/NetStandard13PlatformProxy.cs
```diff
@@ -1,156 +1,154 @@
...
-        }
-
-        public bool IsDomainJoined()
-        {
-            return false;
-        }
-
-        public string GetEnvironmentVariable(string variable)
-        {
+﻿//----------------------------------------------------------------------
```
bgavrilMS, line 1: My VS keeps telling me that spacing is messed up in some files. I am using the standard VS settings, which are "tabs are 4 spaces, and use space chars, not tab chars". 

In any case, no major changes done to this file in particular

core/src/Platforms/uap/UapPlatformProxy.cs
```diff
@@ -156,34 +157,28 @@ public string GetDeviceModel()
...
-        public void ValidateRedirectUri(Uri redirectUri, RequestContext requestContext)
+        public void ValidateRedirectUri(Uri redirectUri)
         {
-            if (_isMsal)
-            {
-            }
-            else
-            {
-                // FROM ADAL
-                if (redirectUri == null)
```
bgavrilMS, line 167: Moved this logic to GetDefaultRedirectUri

jennyf19 (approved):
nice fixes.
LGTM

PR #1404, opened 2018-11-14
Consolidate redirect URI default on ADAL and MSAL

bgavrilMS:


adal/devApps/AdalUAPTestApp/MainPage.xaml.cs
```diff
@@ -43,7 +43,8 @@ public sealed partial class MainPage : Page
     {
         private const string ClientId = "cd01dc27-9d3c-4812-beda-8229d5d4a8d5";
 
-        private const string ReturnUri = "https://MyDirectorySearcherApp";
+        private readonly Uri ReturnUri = null; // will use the broker (WEB)
+        //private readonly Uri ReturnUri = new Uri("https://MyDirectorySearcherApp");
```
henrik-me: >//private readonly Uri ReturnUri = new Uri("https://MyDirectorySearcherApp"); [](start = 8, length = 77)

shy is this one commented? Shouldn't it be possible to select one or the other in the App?

msal/src/Microsoft.Identity.Client/PublicClientApplication.cs
```diff
@@ -468,6 +468,13 @@ internal IWebUI CreateWebAuthenticationDialog(UIParent parent, UIBehavior behavi
             var requestParams = CreateRequestParameters(authority, scopes, null, UserTokenCache);
             requestParams.ExtraQueryParameters = extraQueryParameters;
 
+#if iOS || ANDROID
+            if (!parent.CoreUIParent.UseEmbeddedWebview)
```
henrik-me: >parent.CoreUIParent.UseEmbeddedWebview) [](start = 17, length = 39)

perhaps this should be available on all platforms? then the #if ios || android could be moved to the RedirectUriHelper validation?

core/src/Exceptions/CoreErrorMessages.cs
```diff
@@ -109,9 +109,11 @@ internal class CoreErrorMessages
         public const string AndroidActivityNotFound = "The Activity cannot be found to launch the given Intent. To ensure authentication, a browser with custom tab support " +
             "is recommended. See https://aka.ms/msal-net-system-browsers for more details on using system browser on Android.";
 
-        public const string DefaultRedirectUriIsInvalid = "Default redirect Uri {0} cannot be used on {1} platform";
+        public const string DefaultRedirectUriIsInvalid = "The OAuth2 redirect uri {0} should not be used with the system browser, because the operating system cannot go back to the app. Consider using the default redirect uri for this platform. See https://aka.ms/msal-client-apps for more details.";
 
         public const string RedirectUriContainsFragment = "'redirectUri' must NOT include a fragment component";
 
+        public const string NoRedirectUri = "No redirectUri was configured. ADAL does not provide any defaults.";
```
bgavrilMS, line 116: @jmprieur - Please see the 2 error messages.

core/src/Http/RedirectUriHelper.cs
```diff
@@ -44,6 +53,19 @@ public static void Validate(Uri redirectUri)
                     CoreErrorMessages.RedirectUriContainsFragment,
                     nameof(redirectUri));
             }
+
+            // Currenlty only MSAL supports the system browser, on Android and iOS
+            if (usesSystemBrowser &&
+                Constants.DefaultRedirectUri.Equals(redirectUri.AbsoluteUri, StringComparison.OrdinalIgnoreCase))
```
henrik-me, line 59: Should this instead be called no redirect uri ?

henrik-me (changes requested):
:clock1:

core/src/Http/RedirectUriHelper.cs
bgavrilMS: What do you mean? The constant ? (this is urn:oob:...)

bgavrilMS: The error message in this case will be: 

The OAuth2 redirect uri {0} should not be used with the system browser, because the operating system cannot go back to the app. Consider using the default redirect uri for this platform. See https://aka.ms/msal-client-apps for more details.";

PR #1408, opened 2018-11-16
Add a public ctor to AuthenticationResult to allow developers to mock auth flows

bgavrilMS:
**Why?**
See customers feedback:
- [MSAL is difficult to test because AuthenticationResult is not mockable](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/682)
- [AcquireToken should return IAuthenticationResult instead of AuthenticationResult](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/48*4)

**What?**
to avoid doing a breaking change, we decided to provide a constructor with all the properties, rather than adding an interface and changing all the AcquireTokenXX methods return types.

### DIFF
--- a/msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
+++ b/msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
@@ -110,7 +110,7 @@ namespace Microsoft.Identity.Client
         /// Client applications accept extended life time tokens only if
         /// the ExtendedLifeTimeEnabled Boolean is set to true on ClientApplicationBase.
         /// </summary>
-        public bool IsExtendedLifeTimeToken { get; internal set; }
+        public bool IsExtendedLifeTimeToken { get; }
 
         /// <summary>
         /// Gets the Unique Id of the account. It can be null. When the <see cref="IdToken"/> is not <c>null</c>, this is its ID, that
--- a/msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
+++ b/msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
@@ -91,7 +91,8 @@ namespace Microsoft.Identity.Client
             this.ExtendedExpiresOn = msalAccessTokenCacheItem.ExtendedExpiresOn;
             this.TenantId = msalIdTokenCacheItem?.IdToken?.TenantId;
             this.IdToken = msalIdTokenCacheItem?.Secret;
-            this.Scopes = msalAccessTokenCacheItem.ScopeSet;            
+            this.Scopes = msalAccessTokenCacheItem.ScopeSet;
+            this.IsExtendedLifeTimeToken = msalAccessTokenCacheItem.IsExtendedLifeTimeToken;
         }
 
         /// <summary>
--- a/msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
+++ b/msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
@@ -46,8 +46,8 @@ namespace Microsoft.Identity.Client
         private const string Oauth2AuthorizationHeader = "Bearer ";
 
         /// <summary>
-        /// Constructor meant for test purposes, to allow mocking of authentication flows. Developers should 
-        /// never new up <see cref="AuthenticationResult"/> in product code.
+        /// Constructor meant to help application developers test their apps. Allows mocking of authentication flows. 
+        /// App developers should never new-up <see cref="AuthenticationResult"/> in product code.
         /// </summary>
         public AuthenticationResult(
             string accessToken, 
--- a/msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
+++ b/msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
@@ -44,8 +44,32 @@ namespace Microsoft.Identity.Client
 #pragma warning restore CS1574 // XML comment has cref attribute that could not be resolved
     {
         private const string Oauth2AuthorizationHeader = "Bearer ";
-        private readonly MsalAccessTokenCacheItem _msalAccessTokenCacheItem;
-        private readonly MsalIdTokenCacheItem _msalIdTokenCacheItem;
+
+        /// <summary>
+        /// Constructor meant for test purposes, to allow mocking of authentication flows. Developers should 
+        /// never new up <see cref="AuthenticationResult"/> in product code.
+        /// </summary>
+        public AuthenticationResult(
+            string accessToken, 
+            bool isExtendedLifeTimeToken, 
+            string uniqueId, 
+            DateTimeOffset expiresOn, 
+            DateTimeOffset extendedExpiresOn, 
+            string tenantId, 
+            IAccount account, 
+            string idToken, 
+            IEnumerable<string> scopes)
+        {
+            AccessToken = accessToken;
+            IsExtendedLifeTimeToken = isExtendedLifeTimeToken;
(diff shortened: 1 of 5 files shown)


msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -43,29 +43,62 @@ namespace Microsoft.Identity.Client
...
-        private readonly MsalAccessTokenCacheItem _msalAccessTokenCacheItem;
-        private readonly MsalIdTokenCacheItem _msalIdTokenCacheItem;
-
+        private const string Oauth2AuthorizationHeader = "Bearer ";
+
+        /// <summary>
+        /// Constructor meant for test purposes, to allow mocking of authentication flows. Developers should 
+        /// never new up <see cref="AuthenticationResult"/> in product code.
+        /// </summary>
+        public AuthenticationResult(
```
bgavrilMS, line 52: I know it's a pretty horrible constructor, but I'd rather keep the object immutable than adding setters

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -43,29 +43,62 @@ namespace Microsoft.Identity.Client
...
 #pragma warning restore CS1574 // XML comment has cref attribute that could not be resolved
     {
-        private const string Oauth2AuthorizationHeader = "Bearer ";
-        private readonly MsalAccessTokenCacheItem _msalAccessTokenCacheItem;
-        private readonly MsalIdTokenCacheItem _msalIdTokenCacheItem;
-
+        private const string Oauth2AuthorizationHeader = "Bearer ";
+
+        /// <summary>
+        /// Constructor meant for test purposes, to allow mocking of authentication flows. Developers should
```
bgavrilMS: @jmprieur - does this comment look ok to you?

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -43,29 +43,62 @@ namespace Microsoft.Identity.Client
...
+            this.Scopes = msalAccessTokenCacheItem.ScopeSet;
+            this.IsExtendedLifeTimeToken = msalAccessTokenCacheItem.IsExtendedLifeTimeToken;
         }
 
         /// <summary>
-        /// Gets the Access Token to use as a bearer token to access the protected web API
+        /// Access Token that can be used as a bearer token to access protected web APIs
         /// </summary>
-        public virtual string AccessToken => _msalAccessTokenCacheItem.Secret;
+        public string AccessToken { get; }
```
bgavrilMS, line 101: I fail to see the point in having virtual properties on a data class. Developers cannot inherit from it anyway, because there is no public constructor.

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -43,29 +43,62 @@ namespace Microsoft.Identity.Client
...
+            this.IdToken = msalIdTokenCacheItem?.Secret;
+            this.Scopes = msalAccessTokenCacheItem.ScopeSet;
+            this.IsExtendedLifeTimeToken = msalAccessTokenCacheItem.IsExtendedLifeTimeToken;
         }
 
         /// <summary>
-        /// Gets the Access Token to use as a bearer token to access the protected web API
+        /// Access Token that can be used as a bearer token to access protected web APIs
         /// </summary>
-        public virtual string AccessToken => _msalAccessTokenCacheItem.Secret;
```
jmprieur, line 68: How does it work with derived classes @bgavrilMS ?

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
jmprieur: For the comment, what about?
Constructor ment to help application developers test their apps, by mocking the authentication flows. App developers should never new-up &lt;see cref="AuthenticationResult"/&gt; in product code.

jmprieur (approved):
The comment looks good to me. 
I'm not competent for the other changes :)

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -43,29 +43,62 @@ namespace Microsoft.Identity.Client
...
             }
+
+            this.AccessToken = msalAccessTokenCacheItem.Secret;
+            this.UniqueId = msalIdTokenCacheItem?.IdToken?.GetUniqueId();
+            this.ExpiresOn = msalAccessTokenCacheItem.ExpiresOn;
+            this.ExtendedExpiresOn = msalAccessTokenCacheItem.ExtendedExpiresOn;
+            this.TenantId = msalIdTokenCacheItem?.IdToken?.TenantId;
+            this.IdToken = msalIdTokenCacheItem?.Secret;
+            this.Scopes = msalAccessTokenCacheItem.ScopeSet;
+            this.IsExtendedLifeTimeToken = msalAccessTokenCacheItem.IsExtendedLifeTimeToken;
```
bgavrilMS, line 95: @MarkZuber - IsExtendedLifeTimeToken was not actually being set from anywhere, but I imagine it should just be compied from the accessTokenCacheItem

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
bgavrilMS: I fail to see the point in having virtual properties on a data class. Developers cannot inherit from it anyway, because there is no public constructor (you get compilation errors). We never derive from it.

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -43,29 +43,62 @@ namespace Microsoft.Identity.Client
...
             {
-                Account = new Account(_msalAccessTokenCacheItem.HomeAccountId,
-                    _msalIdTokenCacheItem?.IdToken?.PreferredUsername, _msalAccessTokenCacheItem.Environment);
+                this.Account = new Account(
+                    msalAccessTokenCacheItem.HomeAccountId,
+                    msalIdTokenCacheItem?.IdToken?.PreferredUsername,
+                    msalAccessTokenCacheItem.Environment);
             }
+
+            this.AccessToken = msalAccessTokenCacheItem.Secret;
```
MarkZuber, line 88: nit: remove "this."

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
MarkZuber: correct.  we built that feature but it's currently not enabled.  i can look at that after your checkin.

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
MarkZuber: For MSAL (since we _are_ still in preview) can we just change this to an interface and break it instead of living with this in GA back-compat?

MarkZuber (changes requested):
🕐

msal/src/Microsoft.Identity.Client/AuthenticationResult.cs
bgavrilMS: If the ctor is too much I can expose a builder. Or add setters to those properties. I don't think interfaces on immutable data objects are useful.

MarkZuber (approved):

bgavrilMS:
Rebased and pushed manually

PR #1409, opened 2018-11-16
Fix webUI initialization to avoid passing in a null requestContext

bgavrilMS:


adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -261,10 +261,24 @@ public async Task<AuthenticationResult> AcquireTokenByDeviceCodeAsync(DeviceCode
             IPlatformParameters parameters,
             UserIdentifier userId, string extraQueryParameters, string claims)
         {
-            return await this.AcquireTokenWithClaimsCommonAsync(resource, new ClientKey(clientId), redirectUri,
-                    parameters,
-                    userId, extraQueryParameters, this.CreateWebAuthenticationDialog((PlatformParameters)parameters), claims)
```
bgavrilMS, line 266: The fix is moving this.CreateWebAuthenticationDialog((PlatformParameters)parameters) to the handler classes, because this class does not have a requestContext object.

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/net45/WebUI.cs
```diff
@@ -39,7 +39,10 @@ internal abstract class WebUI : IWebUI
     {
         protected Uri RequestUri { get; private set; }
 
-        protected RequestContext context;
+        protected RequestContext context {
```
MarkZuber: nit formatting: get; set; on same line

MarkZuber (approved):

PR #1411, opened 2018-11-16
add initial benchmark infrastructure

MarkZuber:
This adds a new msal "test" project which is a console app linked to [BenchmarkDotNet](https://benchmarkdotnet.org/).  

It's currently setup to output github markdown and json files.  I'll look into how we can set this up to report up to VSTS results on the benchmark values so we can look for regressions.  It is not yet wired up to run as part of any tests, but this makes sense to run with Integration and end to end validation.

We can also do memory analysis, GC analysis, framework/runtime comparisons, and stack analysis using ETW for deeper diagnosis and debugging.

This tech is used by the CoreFX and CLR teams as well for their profiling.

### DIFF
--- a/LibsNoSamples.sln
+++ b/LibsNoSamples.sln
@@ -36,7 +36,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test.Microsoft.Identity.Lab
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test.Microsoft.Identity.SideBySide", "core\tests\Test.Microsoft.Identity.SideBySide\Test.Microsoft.Identity.SideBySide.csproj", "{D3033A40-F2C2-4495-BC9F-C3C6D497D1CC}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Msal.Benchmarks", "msal\tests\Msal.Benchmarks\Msal.Benchmarks.csproj", "{D41FB9CF-9474-461D-A6EA-8A27C383446E}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Msal.Benchmarks", "msal\tests\Msal.Benchmarks\Msal.Benchmarks.csproj", "{232B5361-18AE-4EA0-9EED-696F3564909C}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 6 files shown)


LibsNoSamples.sln
```diff
@@ -1,108 +1,115 @@
...
-EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test.ADAL.NET.Unit.net45", "adal\tests\Test.ADAL.NET.Unit.net45\Test.ADAL.NET.Unit.net45.csproj", "{91E9E564-8A44-457B-A97C-E76C9F6275D6}"
-EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test.ADAL.NET.Unit.netcore", "adal\tests\Test.ADAL.NET.Unit.netcore\Test.ADAL.NET.Unit.netcore.csproj", "{FCDDA8D3-F299-426B-9CAD-0824D941ED53}"
-EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test.Microsoft.Identity.LabInfrastructure", "core\tests\Test.Microsoft.Identity.LabInfrastructure\Test.Microsoft.Identity.LabInfrastructure.csproj", "{E22918A8-8FE6-47D2-B65E-34436918C5D7}"
-EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test.Microsoft.Identity.SideBySide", "core\tests\Test.Microsoft.Identity.SideBySide\Test.Microsoft.Identity.SideBySide.csproj", "{D3033A40-F2C2-4495-BC9F-C3C6D497D1CC}"
-EndProject
-Global
```
henrik-me, line 39: I can't see through these changes. diff for me claims everything has changed, which ideally it shouldn't as just a project has been added. Are you also seeing the file update completely?

msal/tests/Msal.Benchmarks/ConfidentialClientAcquireTokenWithCache.cs
```diff
@@ -0,0 +1,63 @@
...
+    [SimpleJob(RunStrategy.Throughput, invocationCount: 1000)]
+    public class ConfidentialClientAcquireTokenWithCache
+    {
+        private static readonly TokenCache Cache = new TokenCache();
+        private static readonly TokenCache Cache2 = new TokenCache();
+
+        private static readonly ConfidentialClientApplication App = new ConfidentialClientApplication(
+            "a40e1db0-b7a2-4e6e-af0e-b4987f73228f",
+            "https://login.microsoftonline.com/botframework.com",
+            "localhost:8040",
```
henrik-me, line 45: >"localhost:8040" [](start = 12, length = 16)

does this port need to be enabled on the running system?

henrik-me:
sounds great to get this setup. I would also be happy with an even simpler solution adding one test cases to the integration tests (though then we will not see this value over time in a nice and elegant way).

msal/tests/Msal.Benchmarks/ConfidentialClientAcquireTokenWithCache.cs
MarkZuber: no

LibsNoSamples.sln
MarkZuber: yes, i see that too.  probably a line ending issue, but also sln files are notoriously bad at diffing.  i'll see if i can regen the sln file add and fix the diff.

MarkZuber: Added it again.  local diff showed just the line changes, github diff shows everything different.  from local diff, i can confirm it's only adding the msal.benchmarks file.

image

sangonzal (approved):
Looks good to me

PR #1412, opened 2018-11-19
Sagonzal/dev apps telemetry

sangonzal:
Instrumenting developer apps to send telemetry data to Aria. 

Two open items: 
1) Xforms.Android events are dropped before getting to the collector. Working with Aria team to resolve this.
2) Need to decide how we want to include the Microsoft.Applications.Events Nuget packages, as they are currently in a VSTS repository only accessible to Microsoft FTEs.


Telemetry/TelemetryServerReceiver.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        private ILogger logger;
+        private string msalEventNameKey;
+        private string ariaTenantId;
+
+        public TelemetryServerReceiver()
+        {
+            EVTStatus status;
+            LogManager.Start(new LogConfiguration());
+            logger = LogManager.GetLogger(ariaTenantId, out status);
```
henrik-me: this seems to never be set? did you intend to use:  ariaTenantId = TelemetryReceiverConstants.AriaTenantId; ? #Closed

Telemetry/TelemetryClientReceiver.cs
```diff
@@ -0,0 +1,94 @@
...
+//
+//------------------------------------------------------------------------------
+
+extern alias Client;
+
+using System;
+using System.Collections.Generic;
+using Client::Microsoft.Applications.Events;
+
+namespace TelemetryReceivers
```
henrik-me: We should probably re-consider the namespace:
perhaps: Microsoft.Identity.Client.Telemetry

henrik-me (changes requested):
:clock1:

Telemetry/TelemetryServerReceiver.cs
sangonzal: Yes. Fixed it  #Closed

Telemetry/TelemetryClientReceiver.cs
sangonzal: Sure, changed it to Microsoft.Identity.Client.Telemetry

henrik-me: I'm not seeing this change

---
In reply to: [234785661](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1412#discussion_r234785661) [](ancestors = 234785661)

sangonzal: Should be there now. After the conversation offline, I also made changes to some of the other class names: changed namespace to Microsoft.Identity.Client.DevAppsTelemetry, and changed the "Receiver" classes to  ClientTelemetryHandler and ServerTelemetryHandler . I don't have a strong opinion either way, so if anybody has suggestions for better names I'm happy to change them.

sangonzal:
Moved to MSAL repo

PR #1413, opened 2018-11-20
dev app fixes required for system browser to work w/b2c

jennyf19:


msal/tests/dev apps/XForms/XForms/App.xaml.cs
```diff
@@ -42,7 +42,7 @@ public partial class App : Application
         public const string B2cClientId = "e3b9ad76-9763-4827-b088-80c7a7888f79";
 
         public static string RedirectUriOnAndroid = Microsoft.Identity.Core.Constants.DefaultRedirectUri; // will not work with embedded browser
-        public const string RedirectUriOnIos = "adaliosxformsapp://com.yourcompany.xformsapp";
+        public static string RedirectUriOnIos = "adaliosxformsapp://com.yourcompany.xformsapp";
```
MarkZuber, line 45: why does const not work here?  const should be preferred, this string variable is now mutable.

MarkZuber: from offline conversation, i now see that you need to be able to modify this.

MarkZuber (approved):

PR #1421, opened 2018-11-21
fix adfs test

jennyf19:


core/tests/Test.Microsoft.Identity.Core.UIAutomation/UserInformationFieldIds.cs
```diff
@@ -32,31 +32,21 @@ namespace Test.Microsoft.Identity.Core.UIAutomation
...
-                        break;
-                }
+                // We use the same IDs for ADFSv3 and ADFSv4
+                PasswordInputId = CoreUiTestConstants.AdfsV4WebPasswordID;
+                SignInButtonId = CoreUiTestConstants.AdfsV4WebSubmitID;
+                return;
             }
-
-            if(user.UserType == UserType.B2C)
+            if (user.UserType == UserType.B2C)
```
henrik-me, line 47: could also consider else if ... however generally I believe returning early is the best option. Consider doing that even for the b2c case (return instead of the else).

henrik-me (approved):
:shipit:

PR #1424, opened 2018-11-22
Initial commit for not showing UI w/edit profile b2c policy

jennyf19:
- clean up unused stuff in b2cAuthority class
- Add UI test for edit profile
- add check for b2c authorityType in RequestBase to by-pass cache issue
- Add UIBehavior None
- Add b2c tests in desktop msal app and msal xamarin apps
- And some resharper suggestions

I have some more work to do, esp. in desktop app, but wanted to get something pushed for @jmprieur before the break.
[MSAL issue for context](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/588)


msal/src/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -238,7 +239,7 @@ protected AuthenticationResult CacheTokenResponseAndCreateAuthenticationResult(M
                 fromServer = ClientInfo.CreateFromJson(msalTokenResponse.ClientInfo);
             }
 
-            if (fromServer!= null && AuthenticationRequestParameters?.Account?.HomeAccountId != null)
+            if (fromServer!= null && AuthenticationRequestParameters?.Account?.HomeAccountId != null && AuthenticationRequestParameters.Authority.AuthorityType != AuthorityType.B2C)
```
jennyf19: `fromServer.UniqueObjectIdentifier` returns `"b9f....b2c_1_profileeditpolicy"`
and 
`AuthRequestParameters.Account.HomeObjectId` is `b9f...b2c_1_sisopolicy"` 
Using fiddler, we get a token back, so I added a check for b2c authorityType to skip this check and not get the `"Returned user identifiers do not match the sent user identifier"` error. 
Will look into this more next week.
 #Resolved

msal/tests/Test.MSAL.NET.UIAutomation/MSALMobileTestHelper.cs
```diff
@@ -309,5 +325,21 @@ public void PerformB2CSignInFlow(ITestController controller, LabUser user, B2CId
...
+            SetB2CInputDataForEditProfileAuthority(controller);
+            
+            controller.Tap(CoreUiTestConstants.AcquirePageID);
+            
+            SetUiBehavior(controller, CoreUiTestConstants.UIBehaviorNone);
+
+            //Acquire token flow
+            controller.Tap(CoreUiTestConstants.AcquireTokenID);
+
+            controller.Tap(CoreUiTestConstants.B2CEditProfileContinueID, XamarinSelector.ByHtmlIdAttribute);
```
jennyf19, line 340: I guessed on the name of the id and got it. lol. 😹  #Resolved

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
bgavrilMS: Does the tenant id match - can we preserve that verification at least? 

In any case, this validation code is getting complex, please move it to a separate private method.  #Resolved

msal/tests/Test.MSAL.NET.UIAutomation/MSALMobileTestHelper.cs
bgavrilMS: :D [insert thug life meme here] #Resolved

msal/tests/dev apps/DesktopTestApp/MainForm.cs
```diff
@@ -498,5 +503,95 @@ private IEnumerable<string> SplitScopeString(string scopes)
...
             return scopes.Split(new[] { " " }, StringSplitOptions.None);
         }
+
+        private async void b2cLogin_Click(object sender, EventArgs e)
+        {
+            using (new UIProgressScope(this))
+            {
+                ClearResultPageInfo();
+                string[] scopes = { "https://msidlabb2c.onmicrosoft.com/msidlabb2capi/read" };
+                string b2cClientId = "e3b9ad76-9763-4827-b088-80c7a7888f79";
```
bgavrilMS: I thought that clientIDs are read from the LabResponse? #Resolved

msal/tests/dev apps/DesktopTestApp/MainForm.cs
```diff
@@ -498,5 +503,95 @@ private IEnumerable<string> SplitScopeString(string scopes)
...
+            }
+        }
+
+        private async void b2cEditProfile_Click(object sender, EventArgs e)
+        {
+            using (new UIProgressScope(this))
+            {
+                ClearResultPageInfo();
+                string[] scopes = { "https://msidlabb2c.onmicrosoft.com/msidlabb2capi/read" };
+                string b2cClientId = "e3b9ad76-9763-4827-b088-80c7a7888f79";
```
bgavrilMS: client id should come from the lab reponse and be configured in the app. #Resolved

msal/src/Microsoft.Identity.Client/UIBehavior.cs
```diff
@@ -61,6 +61,12 @@ struct UIBehavior
         /// </summary>
         public static readonly UIBehavior Consent = new UIBehavior("consent");
 
+        /// <summary>
+        /// For B2C scenarios where each policy is a different authority, to prevent UI from being shown
+        /// to a user who has previously acquired a token for a different policy. It is achieved
+        /// by sending <c>prompt=none</c> to the Azure AD service.
+        /// </summary>
+        public static readonly UIBehavior NoPrompt = new UIBehavior("none");
```
bgavrilMS: What are the implications of using this prompt in non-B2C scenarios? Is it something that we want to restrict users from doing, i.e. IF Authority != B2C and prompt == none THEN throw ex ? #Resolved

bgavrilMS (changes requested):
Mostly questions about new prmpt behavior and client id in tests

msal/src/Microsoft.Identity.Client/UIBehavior.cs
jennyf19: I had the same concern. @jmprieur - your thoughts on this?

---
In reply to: [235682552](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1424#discussion_r235682552) [](ancestors = 235682552)

henrik-me:
>             {

consider switch instead of if's #Resolved

---
Refers to: msal/tests/dev apps/XForms/XForms/SettingsPage.xaml.cs:109 in 506af75. [](commit_id = 506af7511140e852b1b8d2348b133c822965b4b3, deletion_comment = False)

msal/tests/Test.MSAL.NET.UIAutomation/MSALMobileTestHelper.cs
```diff
@@ -309,5 +325,21 @@ public void PerformB2CSignInFlow(ITestController controller, LabUser user, B2CId
...
             CoreMobileTestHelper.VerifyResult(controller);
         }
+
+        public void PerformB2CSignInEditProfileFlow(ITestController controller, B2CIdentityProvider b2CIdentityProvider)
+        {
+            SetB2CInputDataForEditProfileAuthority(controller);
+            
+            controller.Tap(CoreUiTestConstants.AcquirePageID);
+            
+            SetUiBehavior(controller, CoreUiTestConstants.UIBehaviorNone);
```
henrik-me: >SetUiBehavior(controller, CoreUiTestConstants.UIBehaviorNone); [](start = 12, length = 62)

Makes sense to ask JM if this is how he envisions the setup. Also... what if the user is not logged in? Currently thinking the service side has to be part of deciding whether UI is shown or not. Would be great to get JM's perspective, and our friends in B2C (e.g. P and J). #Resolved

msal/tests/Test.MSAL.NET.UIAutomation/XamarinDroidMSALTests.cs
```diff
@@ -134,6 +134,19 @@ public void B2CFacebookProviderWithMicrosoftAuthorityAcquireTokenTest()
...
+        /// <summary>
+        /// B2C acquire token with Facebook provider
+        /// b2clogin.com authority
+        /// call to edit profile authority with
+        ///  UIBehavior none
+        /// </summary>
+        [Test]
+        public void B2CFacebookProviderEditPolicyAcquireTokenTest()
+        {
+            _msalMobileTestHelper.B2CFacebookProviderAcquireTokenSilentTest(xamarinController, LabUserHelper.GetLabResponseWithB2CFacebookProvider(), isB2CLoginAuthority:true);
```
henrik-me: >isB2CLoginAuthority:true [](start = 150, length = 24)

any reason you are specifying the name of the parameter? #Closed

henrik-me (changes requested):
:clock1:

msal/tests/Test.MSAL.NET.UIAutomation/XamarinDroidMSALTests.cs
jennyf19: just to make it clear what the boolean is for, but i can remove the name.

---
In reply to: [236440994](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1424#discussion_r236440994) [](ancestors = 236440994)

msal/tests/Test.MSAL.NET.UIAutomation/MSALMobileTestHelper.cs
jennyf19: if it's used as a silent call, server side will return this:
 `One of two conditions was encountered: 1. The UiBehavior.Never flag was passed, but the constraint could not be honored, because user interaction was required. 2. An error occurred during a silent web authentication that prevented the http authentication flow from completing in a short enough time frame` 

---
In reply to: [236440625](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1424#discussion_r236440625) [](ancestors = 236440625)

henrik-me: Trying to clarify my train of thoughts. 

Yeah, this flow doesn't make sense for the silent flow. My question is if the user is not logged in and the first flow you call is the edit profile flow... I assume regardless of us passing in none, the user should still be met with a prompt dialog... this is why I'm asking for server ack of what we are doing. If they treat none as do not show the auth dialogs, then... it will require the developer to know if the user is logged in or not. 

---
In reply to: [236450550](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1424#discussion_r236450550) [](ancestors = 236450550,236440625)

jennyf19: I see what you're saying. I thought you meant if the dev used prompt=none in another flow to mimic a silent call.
To answer your question, if the user is not signed in and does an edit profile w/prompt=none, we get this back from B2C: 
`- AADB2C90077: User does not have an existing session and request prompt parameter has a value of 'None'.`
if you do edit profile w/select account, you can select the account first (login) and then edit the profile.
so, yes, the dev will need to know if the user has logged in, or they can surface this after the user has logged in, I think that's how our samples are set up. You log-in first and then are given the option to edit profile.

---
In reply to: [236455405](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1424#discussion_r236455405) [](ancestors = 236455405,236450550,236440625)

henrik-me: Great. As expected. It will though be hard as a developer to fully build the experience with-out keeping state separately... but perhaps that is good enough. We should have a wiki page explaining this (and the evo messages at time of writing keeping in mind that is all subject to change). Also would be good to investigate what the b2c service would prefer going forward that we send.

---
In reply to: [236458472](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1424#discussion_r236458472) [](ancestors = 236458472,236455405,236450550,236440625)

jennyf19: sounds good. i'll follow up w/b2c

---
In reply to: [236462577](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1424#discussion_r236462577) [](ancestors = 236462577,236458472,236455405,236450550,236440625)

msal/tests/dev apps/DesktopTestApp/MainForm.cs
jennyf19: i fixed this. had to update to 4.6.2 to be compatible w/the lab. 

---
In reply to: [235680206](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1424#discussion_r235680206) [](ancestors = 235680206)

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -276,17 +258,45 @@ protected AuthenticationResult CacheTokenResponseAndCreateAuthenticationResult(M
...
-                        msalTokenResponse, 
+                        AuthenticationRequestParameters.ClientId,
+                        msalTokenResponse,
                         idToken?.TenantId));
             }
         }
 
+        private void ValidateAccountIdentifiers(ClientInfo fromServer)
+        {
+            if (fromServer == null || AuthenticationRequestParameters?.Account?.HomeAccountId == null ) return;
```
bgavrilMS: Prefer 
{
    return;
} #Resolved

msal/tests/Test.MSAL.NET.UIAutomation/XamarinDroidMSALTests.cs
bgavrilMS: Makes the code easier to read... #Pending

msal/tests/dev apps/DesktopTestApp/MainForm.cs
```diff
@@ -349,6 +356,11 @@ private UIBehavior GetUIBehavior()
                 behavior = UIBehavior.Consent;
             }
 
+            if (noPrompt.Checked)
+            {
+                behavior = UIBehavior.NoPrompt; ;
```
bgavrilMS: double ; #Resolved

msal/tests/dev apps/DesktopTestApp/MainForm.cs
```diff
@@ -63,6 +66,8 @@ public MainForm()
 
             LoadSettings();
             Logger.LogCallback = LogDelegate;
+            LabResponse labResponse = LabUserHelper.GetLabResponseWithB2CLocalAccountProvider();
```
bgavrilMS: We shouldn't make a web request to the lab for B2C everytime we open the app, because it is quite expensive. I believe that the way tests run today is that each test closes and reopens the app. So all our non-B2C tests make this call, which is not needed. 

Instead, you can initialize _b2CClientId = null, then when you click on B2CLogin go query the lab. If I press B2CLogin again, then _b2CClientId is initialized so we won't have to call the lab again. 

This is called lazy initialization, and it's perfectly ok to do it manually like I described. Otherwise, C# has a class for this called Lazy<T> https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1?view=netframework-4.7.2 - we use it in a few places in MSAL.  #Resolved

msal/tests/dev apps/DesktopTestApp/MainForm.cs
```diff
@@ -491,12 +503,89 @@ private void cancelOperationButton_Click(object sender, EventArgs e)
...
+        }
+
+        private async void b2cSilentFlow_Click(object sender, EventArgs e)
+        {
+            using (new UIProgressScope(this))
+            {
+                ClearResultPageInfo();
+
+                _publicClientHandler.InteractiveAuthority = B2CAuthority;
+
```
bgavrilMS, line 578: I think you should also set the client ID #Resolved

bgavrilMS (changes requested):
Lazy initialisation for the B2C account, otherwise LGTM

msal/tests/dev apps/DesktopTestApp/MainForm.cs
jennyf19: nice catch! thank you!

---
In reply to: [237127805](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1424#discussion_r237127805) [](ancestors = 237127805)

msal/tests/dev apps/DesktopTestApp/MainForm.cs
jennyf19: awesome! thanks.

---
In reply to: [237133848](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1424#discussion_r237133848) [](ancestors = 237133848)

msal/tests/dev apps/DesktopTestApp/MainForm.cs
bgavrilMS: Yes, it is good to fetch it here, but I meant _publicClientHandler.ApplicationId = _b2CClientId;  #Resolved

bgavrilMS (approved):

msal/tests/dev apps/DesktopTestApp/MainForm.cs
```diff
@@ -29,28 +29,31 @@
...
 {
     public partial class MainForm : Form
     {
-        private const string publicClientId = "0615b6ca-88d4-4884-8729-b178178f7c27";
-
-        private readonly PublicClientHandler _publicClientHandler = new PublicClientHandler(publicClientId);
-        private CancellationTokenSource _cancellationTokenSource;
-
+        private const string PublicClientId = "0615b6ca-88d4-4884-8729-b178178f7c27";
+        private static string _b2CClientId = null;
```
henrik-me: >static [](start = 16, length = 6)

I don't think this needs to be static? #Closed

msal/src/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
```diff
@@ -285,6 +285,11 @@ private void CheckForDuplicateQueryParameters(string queryParams, IDictionary<st
                 authorizationRequestParameters[kvp.Key] = kvp.Value;
             }
 
+            if (_uiBehavior.PromptValue == UIBehavior.NoPrompt.PromptValue)
+            {
+                return authorizationRequestParameters;
+            }
+
             authorizationRequestParameters[OAuth2Parameter.Prompt] = _uiBehavior.PromptValue;
```
henrik-me: this is a bit of a special scenario.... I think to best ensure that this doesn't break some time in the future (by someone not returning anymore....) 
I suggest that you invert the assignment and then only assign the prompt in case the uibehavior is not NoPrompt. makes sense? #Resolved

henrik-me:
>         public static LabResponse GetLabResponseWithB2CLocalAccountProvider()

nit: Consider removing the return type from all the methods... seems a bit too much information. #Resolved

---
Refers to: core/tests/Test.Microsoft.Identity.LabInfrastructure/LabUserHelper.cs:96 in 0952f29. [](commit_id = 0952f29d8b6e7265bbcdbb10748af85517284c0e, deletion_comment = False)

msal/tests/Test.MSAL.NET.UIAutomation/XamarinDroidMSALTests.cs
henrik-me: yeah, imo not needed. The methods should be made for that purpose instead of having to add all the param names in the mehtod call... 

personal opinion: use consts for that instaed if needed.

---
In reply to: [237127305](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1424#discussion_r237127305) [](ancestors = 237127305)

core/src/Instance/B2CAuthority.cs
```diff
@@ -81,14 +80,6 @@ internal B2CAuthority(IValidatedAuthoritiesCache validatedAuthoritiesCache, stri
             }
         }
 
-        private string GetCanonicalAuthorityB2C()
-        {
-            Uri b2cAuthority = new Uri(CanonicalAuthority);
-
-            string canonicalAuthorityUri = string.Format(CultureInfo.InvariantCulture, MicrosoftOnline + b2cAuthority.AbsolutePath);
-            return canonicalAuthorityUri;
-        }
```
henrik-me, line 87: :) #Resolved

henrik-me (approved):
:shipit:

PR #1428, opened 2018-11-26
Make UIParent cross platform friendly

bgavrilMS:
I propose reviewers have a look at each commit separately, with emphasis on the second commit. Also, for reference - 

image

And the wiki article: 
https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/wiki/Understanding-multi-targetting-and-NetStandard


msal/src/Microsoft.Identity.Client/Platforms/Android/UIParent.cs
```diff
@@ -43,7 +43,7 @@ namespace Microsoft.Identity.Client
     /// Android specific UI properties for interactive flows, such as the parent activity and 
     /// which browser to use 
     /// </summary> 
-    public sealed class UIParent
+    public sealed partial class UIParent
```
bgavrilMS: partial is not needed, I will remove it

msal/src/Microsoft.Identity.Client/Platforms/Android/UIParent.cs
```diff
@@ -53,9 +53,11 @@ static UIParent()
             ModuleInitializer.EnsureModuleInitialized();
         }
 
-        /// <summary>
+
+        /// <summary>  
         /// Default constructor.
         /// </summary>
+        [Obsolete("This constructor should not be used because this object requires a parameters of type Activity. ")]
```
bgavrilMS, line 61: Using this ctor on Android leads to a Null Ref later on.

msal/src/Microsoft.Identity.Client/Platforms/Android/UIParent.cs
```diff
@@ -81,6 +83,37 @@ public UIParent(Activity activity, bool useEmbeddedWebview) : this(activity)
             CoreUIParent.UseEmbeddedWebview = useEmbeddedWebview;
         }
 
+        /// <summary>
+        /// Platform agnostic constructor that allows building an UIParent from a NetStandard assembly.
```
bgavrilMS: @jmprieur - please have a look at these new ctors

msal/src/Microsoft.Identity.Client/Platforms/iOS/UIParent.cs
```diff
@@ -62,5 +62,15 @@ public UIParent(bool useEmbeddedWebview) : this()
             CoreUIParent.UseEmbeddedWebview = useEmbeddedWebview;
         }
 
+        /// <summary>
+        /// Platform agnostic constructor that allows building an UIParent from a NetStandard assembly.
+        /// On iOS, the parent is ignored, you can pass null.
+        /// </summary>
+        /// <param name="parent">Ignored on iOS</param>
+        /// <param name="useEmbeddedWebview">Flag to determine between embedded vs system browser. See https://aka.ms/msal-net-uses-web-browser </param>
+        public UIParent(object parent, bool useEmbeddedWebview)         :
```
bgavrilMS: On other platforms, we will require this exact constructor. I was thinking to enforce this with an interface, but ctors cannot be part of an interface. Instead, I could expose a builder object and add an interface to that. It would be more like "IUIParentBuilder" with methods "WithParent(object); UseEmbeddedWebview(bool); The downside is that the builder is less discoverable than a ctor.

msal/src/Microsoft.Identity.Client/Platforms/netstandard1.3/UIParent.cs
```diff
@@ -59,5 +51,24 @@ public UIParent()
...
+        /// On Android, it is mandatory to pass in an Activity.
+        /// On all other platforms, it is not required.
+        /// On .net desktop, it is optional - you can either pass a System.Windows.Forms.IWin32Window or an System.IntPtr
+        /// to a window handle or null. This is used to center the webview. </param>
+        /// <param name="useEmbeddedWebview">Flag to determine between embedded vs system browser. Currently affects only iOS and Android. See https://aka.ms/msal-net-uses-web-browser </param>
+        public UIParent(object parent, bool useEmbeddedWebview)
+        {
+            // for the rare case when an application actually uses the netstandard implementation
+            // i.e. other frameworks - e.g. Xamarin.MAC - or MSAL.netstandard loaded via reflection
+            throw new PlatformNotSupportedException("Interactive Authentication flows are not supported when the NetStandard assembly is at runtime. " +
```
bgavrilMS: NetStandard assembly will not be used at runtime in 99% of the cases. But in the rare cases where it will, better to throw a good exception. 

image

msal/src/Microsoft.Identity.Client/Platforms/netcore/UIParent.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        /// <summary>
+        /// Platform agnostic constructor that allows building an UIParent from a NetStandard assembly.
+        /// </summary>
+        /// <param name="ownerWindow">Parent window object reference. OPTIONAL. The expected parent window
+        /// are either of type <see cref="System.Windows.Forms.IWin32Window"/> or <see cref="System.IntPtr"/> (for window handle)</param>
+        /// <param name="useEmbeddedWebview">Ignored, on .net desktop an embedded webview is always used</param>
+        public UIParent(object parent, bool useEmbeddedWebview)
+        {
+            throw new PlatformNotSupportedException("Interactive Authentication flows are not supported on .net core. " +
```
bgavrilMS, line 72: Instead of this run-time exception, we could employ a bait-and-switch technique which would ensure: 

- If the user is targetting .net core DIRECTLY, then he will not see this ctor
- If they are targetting .net core via a NetStandard dll, they will get this exception

henrik-me:
Seems like we can't really break free of the platform specifics due to the Activity requirement on Android. Currently not convinced to go down this route.

msal/src/Microsoft.Identity.Client/Platforms/Android/UIParent.cs
```diff
@@ -115,27 +99,16 @@ public UIParent(Activity activity, bool useEmbeddedWebview) : this(activity)
...
-            }
-            else if (isBrowserWithCustomTabSupportAvailable)
-            {
-                return true;
-            }
-            else
-            {
-                return false;
-            }
+            return (isBrowserWithCustomTabSupportAvailable || IsChromeEnabled()) &&
```
MarkZuber, line 141: Isn't this the same as "isBrowserWithCustomTabSupportAvailable" by itself since you're AND'ing to it...

msal/src/Microsoft.Identity.Client/Platforms/iOS/UIParent.cs
MarkZuber: use a factory method to create if you need a specific interface for construction?

msal/src/Microsoft.Identity.Client/Platforms/Android/UIParent.cs
bgavrilMS: This is just R# based refactoring ... but I did have a look at this. AFAIK (a ^ b) | a = (a | a) ^ (b | a) = a ^ (b | a)  - am I missing a simplification ?

msal/src/Microsoft.Identity.Client/Platforms/netcore/UIParent.cs
bgavrilMS: @henrik-me, @MarkZuber   - I want to use reference assemblies to hide these methods. On .net core, we need to have the ctor UIParent(object, bool), in order to support NetStandard entry point. However, if the user programs directly against .net core, he should not see this method. 

With ref assemblies, this will be achieved: 

Behaviour via NetStandard - PlatformNotSupportedException
Behavior via NetCore - methods are hidden in the editor / at build time

msal/src/Microsoft.Identity.Client/Platforms/iOS/UIParent.cs
bgavrilMS: I plan to use the bait and switch technique to hide some of these methods (see PR about using ref assemblies). I don't think we can use interfaces in this case. I'll experiement to see.

msal/src/Microsoft.Identity.Client/Platforms/iOS/UIParent.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// system browser. See https://aka.ms/msal-net-uses-web-browser
+        /// </summary>
+        /// <remarks>This method is likely to be removed (replaced) before final release</remarks>
+        public UIParent(bool useEmbeddedWebview) : this()
+        {
+            CoreUIParent.UseEmbeddedWebview = useEmbeddedWebview;
+        }
+
+        /// <summary>
+        /// Platform agnostic constructor that allows building an UIParent from a NetStandard assembly.
```
bgavrilMS, line 68: @henrik-me, @MarkZuber  - on iOS, this ctor is a bit out of place, because the "parent" param is always ignored. The same happens on net45 (useEmbeddedWebview is ignored), UWP (both params are ignored). However, the ctors need to exist to support calling the equivalent ctor from NetStanadrd. Otherwise, we'd get a MethodMissingException. 

Using the bait and switch technique (i.e. ref assemblies), we get the best of both worlds: 

- constructors exist at runtime, to allow creation from NetStandard 
- constructors are hidden at edit / build time if user programs directly against iOS so as not to increase public API / pollute the type

msal/src/Microsoft.Identity.Client/Platforms/uap/UIParent.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// <summary>
+        /// Platform agnostic constructor that allows building an UIParent from a NetStandard assembly.
+        /// On UWP, both arguments are currently ignored.
+        /// </summary>
+        /// <param name="parent">Ignored on UWP</param>
+        /// <param name="useEmbeddedWebview">UWP supports only embeddedWebview</param>
+        public UIParent(object parent, bool useEmbeddedWebview) 
+            : this()
+        {
+        }
```
henrik-me, line 67: was assuming you would remove this one? or at least hide it in the reference assembly?

msal/src/Microsoft.Identity.Client/Platforms/netstandard1.3/UIParent.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <remarks>Interactive auth is not currently implemented in .net core</remarks>
+        /// <param name="parent">An owner window to which to attach the webview to.
+        /// On Android, it is mandatory to pass in an Activity.
+        /// On all other platforms, it is not required.
+        /// On .net desktop, it is optional - you can either pass a System.Windows.Forms.IWin32Window or an System.IntPtr
+        /// to a window handle or null. This is used to center the webview. </param>
+        /// <param name="useEmbeddedWebview">Flag to determine between embedded vs system browser. Currently affects only iOS and Android. See https://aka.ms/msal-net-uses-web-browser </param>
+        public UIParent(object parent, bool useEmbeddedWebview)
+        {
+            ThrowPlatformNotSupported();
```
henrik-me, line 67: >ThrowPlatformNotSupported(); [](start = 12, length = 28)

I thought this was required in netstandard...

msal/src/Microsoft.Identity.Client/Platforms/Android/UIParent.cs
```diff
@@ -25,68 +25,53 @@
 //
 //------------------------------------------------------------------------------
 
-#if ANDROID
+
```
bgavrilMS, line 28: I've split UIParent.cs into platform specific UIParent because the class was too fragmented already.

msal/src/Microsoft.Identity.Client/Platforms/Android/UIParent.cs
```diff
@@ -99,6 +84,43 @@ public UIParent(Activity activity, bool useEmbeddedWebview) : this(activity)
             CoreUIParent.UseEmbeddedWebview = useEmbeddedWebview;
         }
 
+        #if ANDROID_RUNTIME
```
bgavrilMS, line 87: @MarkZuber  - how about these flags? I also have defined ANDROID_BUILD etc. but these aren't used here.

msal/src/Microsoft.Identity.Client/Platforms/iOS/UIParent.cs
MarkZuber: Looks good.  CAn you add your comments from the PR here into the code itself?  I think that's a helpful reminder of what's going on here and it would be great to persist that knowledge into the codebase.

PR #1430, opened 2018-11-26
Add Microsoft.Identity.Json infra to build for upcoming cache work

MarkZuber:


msal/src/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -41,28 +42,28 @@
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp1.0' ">
     <DefineConstants>$(DefineConstants);NET_CORE</DefineConstants>
-  </PropertyGroup>
+    <DefineConstants>$(DefineConstants);NETSTANDARD1_0;DOTNET;PORTABLE;HAVE_ASYNC;HAVE_COVARIANT_GENERICS;HAVE_DATA_CONTRACTS;HAVE_DATE_TIME_OFFSET;HAVE_FSHARP_TYPES;HAVE_GUID_TRY_PARSE;HAVE_HASH_SET;HAVE_IGNORE_DATA_MEMBER_ATTRIBUTE;HAVE_INOTIFY_COLLECTION_CHANGED;HAVE_ISET;HAVE_LINQ;HAVE_METHOD_IMPL_ATTRIBUTE;HAVE_READ_ONLY_COLLECTIONS;HAVE_REFLECTION_BINDER;HAVE_SERIALIZATION_BINDER_BIND_TO_NAME;HAVE_STRING_JOIN_WITH_ENUMERABLE;HAVE_TIME_SPAN_PARSE_WITH_CULTURE;HAVE_TIME_SPAN_TO_STRING_WITH_CULTURE;HAVE_TIME_ZONE_INFO;HAVE_TYPE_DESCRIPTOR;HAVE_UNICODE_SURROGATE_DETECTION;HAVE_VARIANT_TYPE_PARAMETERS;HAVE_VERSION_TRY_PARSE;HAVE_OBSOLETE_FORMATTER_ASSEMBLY_STYLE;$(AdditionalConstants)</DefineConstants>  </PropertyGroup>
```
bgavrilMS: Hmm, Where were these picked up from ? The whole "flags for features" are ok, but I am hesitant about adding the flag NETSTANDARD1_3 to Xamarin.iOS (and the other "platform" flags, but this one in particular is worrisome). This is because currently we associate a platform to a target framework against which we build our assembly. Adding new flags breaks this association (mental association and in the case of iOS == netstandard1_3, also real association) - how will we be able to define an object that lives in the netstandard1.3 version of our DLL but not in iOS (and vice versa) ?

Can the code you added be refactored so that it only uses the feature flags?

MarkZuber: These were picked up from the newtonsoft.json csproj file.  These are the constants they use to build with.  I got these from adding newtonsoft as a dependency to msal and then looking at which newtonsoft build version it picked up and then dropping them in.  If we want to remove the netstandard ones, i can look at where they're defined and see if there's some other flag we could use.  I had to fix the iOS case in our code in a few places which you'll see in changes outside of the /json subdir...  extremely happy to get input/guidance here since you're so deep in the netstandard configuration stuff right now and i want to ensure we're aligned properly there.

MarkZuber: OK.  I've fixed these to use our defines as necessary and removed the netstandard checks.

henrik-me:
There are a few public types (enums). Can those be internal for now and if we need to expose anything we can do that on a per needed basis? (ideally I would have liked to keep this internal)

msal/src/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -41,27 +42,33 @@
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp1.0' ">
     <DefineConstants>$(DefineConstants);NET_CORE</DefineConstants>
+    <DefineConstants>$(DefineConstants);DOTNET;PORTABLE;HAVE_ASYNC;HAVE_COVARIANT_GENERICS;HAVE_DATA_CONTRACTS;HAVE_DATE_TIME_OFFSET;HAVE_FSHARP_TYPES;HAVE_GUID_TRY_PARSE;HAVE_HASH_SET;HAVE_IGNORE_DATA_MEMBER_ATTRIBUTE;HAVE_INOTIFY_COLLECTION_CHANGED;HAVE_ISET;HAVE_LINQ;HAVE_METHOD_IMPL_ATTRIBUTE;HAVE_READ_ONLY_COLLECTIONS;HAVE_REFLECTION_BINDER;HAVE_SERIALIZATION_BINDER_BIND_TO_NAME;HAVE_STRING_JOIN_WITH_ENUMERABLE;HAVE_TIME_SPAN_PARSE_WITH_CULTURE;HAVE_TIME_SPAN_TO_STRING_WITH_CULTURE;HAVE_TIME_ZONE_INFO;HAVE_TYPE_DESCRIPTOR;HAVE_UNICODE_SURROGATE_DETECTION;HAVE_VARIANT_TYPE_PARAMETERS;HAVE_VERSION_TRY_PARSE;HAVE_OBSOLETE_FORMATTER_ASSEMBLY_STYLE;$(AdditionalConstants)</DefineConstants>
```
henrik-me, line 46: ><DefineConst [](start = 4, length = 12)

.... comments are not often seen here, however I would suggest adding a comment saying those feature flags are for newtonsoft.json

MarkZuber: will do 

> There are a few public types (enums). Can those be internal for now and if we need to expose anything we can do that on a per needed basis? (ideally I would have liked to keep this internal)

i missed a few public types.  full intention on this being internal only.  will fix.

henrik-me (approved):
:shipit:

msal/src/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -150,7 +150,7 @@ public void LogState()
             builder.AppendLine("Extra Query Params Keys (space separated) - " + dict.Keys.AsSingleString());
             dict = CoreHelpers.ParseKeyValueList(SliceParameters, '&', true, RequestContext);
             builder.AppendLine("Slice Parameters Keys(space separated) - " + dict.Keys.AsSingleString());
-#if DESKTOP || NETSTANDARD1_3 || NET_CORE
+#if (DESKTOP || NETSTANDARD1_3 || NET_CORE) && !iOS
```
bgavrilMS, line 153: The changes to this file should be undone now that iOS != netstandard

msal/src/Microsoft.Identity.Client/Internal/Requests/RequestValidationHelper.cs
```diff
@@ -36,7 +36,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
 {
     internal static class RequestValidationHelper
     {
-#if DESKTOP || NETSTANDARD1_3 || NET_CORE
+#if (DESKTOP || NETSTANDARD1_3 || NET_CORE) && !iOS
```
bgavrilMS, line 39: changes to be undone

msal/src/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
bgavrilMS: LGTM!

PR #1432, opened 2018-11-26
Add support for reference assemblies

bgavrilMS:
This is based on Oren's - @onovotny - work to define Reference assemblies - see more details and use cases here: https://oren.codes/2018/07/09/create-and-pack-reference-assemblies-made-easy/

We can use this to manipulate the public API surface of NetStandard / actual platforms, for example to hide methods that make sense on NetStandard but do not make sense on specific platforms.


core/src/Microsoft.Identity.Core.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="MSBuild.Sdk.Extras/1.5.4">
```
bgavrilMS, line 1: Version is now defined in global.json

msal/src/MSAL.Common.props
```diff
@@ -0,0 +1,204 @@
+<Project>
```
bgavrilMS, line 1: I moved most of the MSAL.csproj into this file, which will be used by MSAL.ref.csproj as well. I will point out differences, as csproj files are a joy to read.

msal/src/MSAL.Common.props
```diff
@@ -0,0 +1,204 @@
...
+      MSAL makes it easy to obtain tokens from Azure AD v2 (work &amp; school accounts, MSA) and Azure AD B2C, gaining access to Microsoft Cloud API and any other API secured by Microsoft identities. This version supports adding authentication functionality to your .NET based client on Windows desktop (.NET 4.5+), UWP, .NET Core, Xamarin iOS and Xamarin Android.
+    </Description>
+    <PackageTags>Microsoft Authentication Library MSA MSAL B2C Azure Active Directory AAD Identity Authentication .NET Windows Store Xamarin iOS Android</PackageTags>
+    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
+    <Company>Microsoft Corporation</Company>
+    <Product>Microsoft Authentication Library</Product>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <PathToMsalSources>$(MSBuildThisFileDirectory)\Microsoft.Identity.Client</PathToMsalSources>
```
bgavrilMS, line 41: I defined an extra 2 properties for paths so that they can be used from 2 different projects

msal/src/MSAL.Common.props
```diff
@@ -0,0 +1,204 @@
...
+    <PathToCoreSources>$(MSBuildThisFileDirectory)\..\..\core\src</PathToCoreSources>
+    <!-- Need to include / exclude / remove cs files manually so that they are present in both assemblies -->
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
```
bgavrilMS, line 44: Including the default files causes errors now that these files are used by 2 projects

msal/src/MSAL.Common.props
```diff
@@ -0,0 +1,204 @@
...
+
+  <ItemGroup>
+    <!-- This None is here so the conditionally included files show up in the Solution Explorer -->
+    <None Include="$(PathToMsalSources)\**\*.cs;$(PathToMsalSources)\**\*.xml;$(PathToMsalSources)\**\*.axml" 
+          Exclude="$(PathToMsalSources)\obj\**\*.*;$(PathToMsalSources)\bin\**\*.*" />
+    <!-- Manually include the cs files -->    
+    <Compile Include="$(PathToMsalSources)\**\*.cs" Exclude="$(PathToMsalSources)\obj\**\*.*" />    
+    <Compile Remove="$(PathToMsalSources)\Platforms\**\*.*" />
+    <Compile Remove="$(PathToMsalSources)\Features\**\*.*" />
+
```
bgavrilMS, line 90: I had to add lines 77:80 to emulate "include default files" property that had to be set to false.

msal/src/Microsoft.Identity.Client.Ref/Microsoft.Identity.Client.Ref.csproj
```diff
@@ -0,0 +1,11 @@
+﻿<Project Sdk="MSBuild.Sdk.Extras">
+
+  <!-- This project is used to generated a ref assembly for MSAL. Use #if REFERENCE_ASSEMBLY in the code files. 
+  See https://oren.codes/2018/07/09/create-and-pack-reference-assemblies-made-easy/ for details-->
```
bgavrilMS, line 4: If you do an msbuild /t:pack, you will notice it now contains ref assemblies. Note that I haven't updated the release build yet.

msal/src/MSAL.Common.props
```diff
@@ -0,0 +1,204 @@
...
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.3'">
+    <Compile Include="$(PathToMsalSources)\Platforms\netstandard1.3\**\*.cs" LinkBase="Platforms\netstandard1.3" />
+    <Compile Include="$(PathToCoreSources)\Platforms\netstandard1.3\**\*.cs" LinkBase="Core\Platforms\netstandard1.3" />
+    <Compile Include="$(PathToCoreSources)\Features\Cache\**\*.cs" LinkBase="Core\Features\Cache" />
+    <Compile Include="$(PathToMsalSources)\Features\PublicClientWithTokenCache\**\*.cs" LinkBase="Features\PublicClientWithTokenCache" />
+    <Compile Include="$(PathToMsalSources)\Features\ConfidentialClient\**\*.cs"  LinkBase="Features\ConfidentialClient" />
+    <Compile Include="$(PathToMsalSources)\Features\EventSource\**\*.cs" LinkBase="Features\EventSource" />
+    <Compile Include="$(PathToMsalSources)\Features\TokenCache\**\*.cs" LinkBase="Features\TokenCache" />
+    <Compile Include="$(PathToMsalSources)\Features\DeviceCode\**\*.cs" LinkBase="Features\DeviceCode"/>
+    <PackageReference Include="System.Runtime.Serialization.Primitives" Version="4.3.0" />
```
MarkZuber, line 112: since these are now in a props file and not the csproj, will nuget update from within VS work properly?

henrik-me:
overall looks ok to me. will follow up with a few q's separately

msal/src/MSAL.Common.props
bgavrilMS: The VS experience looks unchanged on my machine. I think it is because it is using the new csproj flavour, where reloading is dynamic.

henrik-me:
@trwalke : this will impact your current change

henrik-me (approved):
:shipit:

msal/src/Microsoft.Identity.Client.Ref/Microsoft.Identity.Client.Ref.csproj
```diff
@@ -0,0 +1,11 @@
...
+  <PropertyGroup>
+    <TargetFrameworks>net45;netstandard1.3;uap10.0;MonoAndroid8.1;Xamarin.iOS10;netcoreapp1.0</TargetFrameworks>
```
trwalke, line 6: So I am adding conditional targets for the macOS build in the ADAL/MSAL/CORE csproj files. With these changes, do I need to also modify the new Ref.csproj files for the macOS build?

MarkZuber (approved):

msal/src/Microsoft.Identity.Client.Ref/Microsoft.Identity.Client.Ref.csproj
clairernovotny: the targets in the ref do not need to match the implementation, but the public surface area should be there that you want the compiler to see on each platform.

bgavrilMS: Thanks Oren. Discussed this offline with Travis - for the build on Mac (which we need for test purposes) we need to use a subset of the target frameworks. Travis is right - the simplest will be to use the same targets (on Win and on Mac) on both MSAL and MSAL.ref csproj files.

jmprieur:
I like this!

jennyf19 (approved):

PR #1433, opened 2018-11-27
First round of CacheV2 infrastructure

MarkZuber:
This brings in the first round of code needed for interaction with per-platform persistence (like MSALC++ is doing).  It's based off of the msal c++ cachemanager/storagemanager architecture with the goal of being able to share token persistence with those platforms (e.g. VS share with Office).

The cachev2 is NOT wired up in any way to real code yet.  I have some of that for a later checkin, but want to start with the v2 cache infra isolated, get feedback and polish in place, and flesh out remaining test infrastructure.


bgavrilMS:
This is a bit too much to review, but I assume most of the code is adapted ? Would be great if we can sync up on this. Otherwise, I don't have any comments.

MarkZuber:
> This is a bit too much to review, but I assume most of the code is adapted ? Would be great if we can sync up on this. Otherwise, I don't have any comments.

I debated sending all of this out or doing it in smaller chunks.  If you'd prefer, I can break this down into smaller pieces for more detailed review.  Please let me know.

Let's chat today in our 1x1 conf call...

msal/src/Microsoft.Identity.Client/CacheV2/CacheManager.cs
```diff
@@ -0,0 +1,432 @@
...
+
+namespace Microsoft.Identity.Client.CacheV2
+{
+    /// <inheritdoc />
+    internal class CacheManager : ICacheManager
+    {
+        private readonly AuthenticationRequestParameters _authParameters;
+        private readonly IStorageManager _storageManager;
+
+        public CacheManager(IStorageManager storageManager, AuthenticationRequestParameters authParameters)
```
MarkZuber, line 45: This file is a port from msal c++, it's meant to be constructed in context with the current call which is why it takes AuthParameters as constructor.

msal/src/Microsoft.Identity.Client/CacheV2/Impl/FileSystemCredentialPathManager.cs
```diff
@@ -0,0 +1,166 @@
...
+
+using System;
+using System.IO;
+using Microsoft.Identity.Client.CacheV2.Impl.Utils;
+using Microsoft.Identity.Client.CacheV2.Schema;
+using Microsoft.Identity.Core;
+
+namespace Microsoft.Identity.Client.CacheV2.Impl
+{
+    internal class FileSystemCredentialPathManager : ICredentialPathManager
```
MarkZuber, line 36: This file follows the msal c+ directory structure building algorithm.

msal/src/Microsoft.Identity.Client/CacheV2/Impl/IStorageManager.cs
```diff
@@ -0,0 +1,91 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+// 
+// ------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using Microsoft.Identity.Client.CacheV2.Schema;
+
+namespace Microsoft.Identity.Client.CacheV2.Impl
```
MarkZuber, line 31: This is the primary interface agreed on by msal c++ and all of the native libraries for abstracting token/credential/account storage persistence.

msal/src/Microsoft.Identity.Client/CacheV2/Schema/StorageJsonUtils.cs
```diff
@@ -0,0 +1,232 @@
...
+using System.Globalization;
+using Microsoft.Identity.Client.CacheV2.Impl.Utils;
+using Microsoft.Identity.Json.Linq;
+
+namespace Microsoft.Identity.Client.CacheV2.Schema
+{
+    /// <summary>
+    /// This class contains the methods for encoding/decoding our object representations of cache data.
+    /// If you're modifying this class, you're updating the schema persistence behavior so ensure you're
+    /// aligned with the other cache schema models.
```
MarkZuber, line 38: this is the serialization/deserialization block for persistence

core/src/Platforms/net45/NetDesktopCryptographyManager.cs
```diff
@@ -36,16 +36,7 @@ internal class NetDesktopCryptographyManager : ICryptographyManager
...
-            {
-                return null;
-            }
-
-            using (SHA256Cng sha = new SHA256Cng())
-            {
-                UTF8Encoding encoding = new UTF8Encoding();
-                return Base64UrlHelpers.Encode(sha.ComputeHash(encoding.GetBytes(input)));
-            }
+            return string.IsNullOrEmpty(input) ? null : Base64UrlHelpers.Encode(CreateSha256HashBytes(input));
```
jennyf19, line 39: >IsNullOrEmpty [](start = 26, length = 13)

In the iOS and Android versions you have `.IsNullOrWhiteSpace` here. #Closed

core/src/Platforms/net45/NetDesktopCryptographyManager.cs
```diff
@@ -61,17 +52,16 @@ public string GenerateCodeVerifier()
 
         public string CreateSha256Hash(string input)
         {
-            if (string.IsNullOrEmpty(input))
-            {
-                return null;
-            }
+            return string.IsNullOrEmpty(input) ? null : Convert.ToBase64String(CreateSha256HashBytes(input));
```
jennyf19, line 55: >IsNullOrEmpty [](start = 26, length = 13)

same as above. #Closed

core/src/Platforms/netcore/NetCoreCryptographyManager.cs
```diff
@@ -36,16 +36,7 @@ internal class NetCoreCryptographyManager : ICryptographyManager
...
-            {
-                return null;
-            }
-
-            using (SHA256 sha = SHA256.Create())
-            {
-                UTF8Encoding encoding = new UTF8Encoding();
-                return Base64UrlHelpers.Encode(sha.ComputeHash(encoding.GetBytes(input)));
-            }
+            return string.IsNullOrEmpty(input) ? null : Base64UrlHelpers.Encode(CreateSha256HashBytes(input));
```
jennyf19, line 39: >IsNullOrEmpty [](start = 26, length = 13)

In the iOS and Android versions you have `.IsNullOrWhiteSpace` here. #Closed

msal/src/Microsoft.Identity.Client/CacheV2/CacheManager.cs
```diff
@@ -0,0 +1,432 @@
...
+                        environment,
+                        clientId,
+                        cachedAt,
+                        tokenResponse.RefreshToken,
+                        string.Empty));
+            }
+
+            if (tokenResponse.HasAccessToken)
+            {
+                long expiresOn = tokenResponse.ExpiresOn.Ticks; // todo: this is probably wrong
```
jennyf19, line 236: >Ticks [](start = 57, length = 5)

i think we currently do minutes and from 5 as default...

msal/src/Microsoft.Identity.Client/CacheV2/CacheManager.cs
```diff
@@ -0,0 +1,432 @@
...
+                authorityType,
+                idTokenJwt.PreferredUsername,
+                idTokenJwt.GivenName,
+                idTokenJwt.FamilyName,
+                idTokenJwt.MiddleName,
+                idTokenJwt.Name,
+                idTokenJwt.AlternativeId,
+                tokenResponse.RawClientInfo,
+                string.Empty);
+
```
jennyf19, line 295: if some of these values are not returned, will that impact anything? for example, b2c idTokens might be different from aad idTokens as far as the claims that are returned. Also, why do we need all  these values for the account (ex. MiddleName)? #Closed

msal/src/Microsoft.Identity.Client/CacheV2/ITokenCacheAdapter.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        Task<MsalAccessTokenCacheItem> FindAccessTokenAsync(AuthenticationRequestParameters authenticationRequestParameters);
+        MsalIdTokenCacheItem GetIdTokenCacheItem(MsalIdTokenCacheKey msalIdTokenCacheKey, RequestContext requestContext);
+        Task<MsalRefreshTokenCacheItem> FindRefreshTokenAsync(AuthenticationRequestParameters authenticationRequestParameters);
+
+        void SetKeychainSecurityGroup(string keychainSecurityGroup);
+        ICollection<string>  GetAllAccessTokenCacheItems(RequestContext requestContext);
+        ICollection<MsalAccessTokenCacheItem> GetAllAccessTokensForClient(RequestContext requestContext);
+        ICollection<MsalAccountCacheItem> GetAllAccounts(RequestContext requestContext);
+        ICollection<string> GetAllAccountCacheItems(RequestContext requestContext);
```
jennyf19: >Account [](start = 34, length = 7)

missing an s ? -> `Accounts`

msal/src/Microsoft.Identity.Client/CacheV2/ITokenCacheAdapter.cs
```diff
@@ -0,0 +1,75 @@
...
+        ICollection<MsalIdTokenCacheItem> GetAllIdTokensForClient(RequestContext requestContext);
+        ICollection<string> GetAllIdTokenCacheItems(RequestContext requestContext);
+        ICollection<MsalRefreshTokenCacheItem> GetAllRefreshTokensForClient(RequestContext requestContext);
+        ICollection<string> GetAllRefreshTokenCacheItems(RequestContext requestContext);
+        void RemoveMsalAccount(IAccount user, RequestContext requestContext);
```
jennyf19: >user [](start = 40, length = 4)

should this be a user or an account? #Closed

msal/src/Microsoft.Identity.Client/CacheV2/Schema/Credential.cs
```diff
@@ -0,0 +1,256 @@
...
+        public long ExtendedExpiresOn { get; set; }
+        public string Secret { get; set; }
+
+        public string AdditionalFieldsJson
+        {
+            get => _additionalFieldsJson;
+            set => _additionalFieldsJson = string.IsNullOrWhiteSpace(value) ? string.Empty : JObject.Parse(value).ToString();
+        }
+
+        public bool Equals(Credential other)
```
jennyf19, line 61: >other [](start = 38, length = 5)

what is other? can you use a more specific name? #Closed

msal/src/Microsoft.Identity.Client/CacheV2/Schema/StorageJsonValues.cs
```diff
@@ -0,0 +1,41 @@
...
+    internal static class StorageJsonValues
+    {
+        public const string AuthorityTypeAdfs = "ADFS";
+        public const string AuthorityTypeMsa = "MSA";
+        public const string AuthorityTypeMsSts = "MSSTS";
+        public const string AuthorityTypeOther = "Other";
+        public const string CredentialTypeRefreshToken = "RefreshToken";
+        public const string CredentialTypeAccessToken = "AccessToken";
+        public const string CredentialTypeIdToken = "IdToken";
+        public const string CredentialTypeOther = "Other";
```
jennyf19, line 39: what about B2C? Does that need to be included? #Closed

jennyf19:
i made it through all but the tests...looks pretty cool. Might be a good idea to add some logging in and around some of the cache stuff. will follow-up w/some other questions off-line. 

---
In reply to: [442116485](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1433#issuecomment-442116485) [](ancestors = 442116485)

core/src/ICryptographyManager.cs
```diff
@@ -32,6 +32,7 @@ internal interface ICryptographyManager
         string CreateBase64UrlEncodedSha256Hash(string input);
         string GenerateCodeVerifier();
         string CreateSha256Hash(string input);
+        byte[] CreateSha256HashBytes(string input);
```
henrik-me, line 35: aside: we should also be supporting 512... possibly something to align with other libraries on afaik evo supports both right now.

core/src/Platforms/uap/UapCryptographyManager.cs
```diff
@@ -72,15 +72,24 @@ public string CreateSha256Hash(string input)
...
             IBuffer inputBuffer = CryptographicBuffer.ConvertStringToBinary(input, BinaryStringEncoding.Utf8);
 
             var hasher = HashAlgorithmProvider.OpenAlgorithm("SHA256");
             IBuffer hashed = hasher.HashData(inputBuffer);
 
             return CryptographicBuffer.EncodeToBase64String(hashed);
         }
-        
+
+        public byte[] CreateSha256HashBytes(string input)
```
henrik-me, line 89: input validation? #Closed

core/src/Platforms/uap/UapCryptographyManager.cs
```diff
@@ -72,15 +72,24 @@ public string CreateSha256Hash(string input)
             {
                 return null;
             }
-            
+
             IBuffer inputBuffer = CryptographicBuffer.ConvertStringToBinary(input, BinaryStringEncoding.Utf8);
 
             var hasher = HashAlgorithmProvider.OpenAlgorithm("SHA256");
             IBuffer hashed = hasher.HashData(inputBuffer);
```
henrik-me, line 85: consider calling the method below instead of duplicating #Closed

msal/tests/Test.MSAL.NET.Common/Core/Helpers/RandomDataUtils.cs
```diff
@@ -0,0 +1,43 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+// 
+// ------------------------------------------------------------------------------
+
+using System;
+
+namespace Test.MSAL.NET.Common.Core.Helpers
+{
+    public static class RandomDataUtils
```
henrik-me, line 32: nit: naming., funny enough this utils is in a helper folder... #Pending

msal/src/Microsoft.Identity.Client/CacheV2/TokenCacheV2.cs
```diff
@@ -0,0 +1,234 @@
...
+    /// </summary>
+    internal class TokenCacheV2 : ITokenCache
+    {
+        private readonly object _lock = new object();
+
+        //private TokenCacheNotification _afterAccess;
+        //private TokenCacheNotification _beforeAccess;
+        //private TokenCacheNotification _beforeWrite;
+        private CacheData _deserializedCacheDataHolding = new CacheData();
+        private IStorageManager _storageManager;
```
henrik-me, line 49: not sure I understand the scenario for either having a storage manager or not... why is the facade doing 2 things. #Pending

msal/src/Microsoft.Identity.Client/CacheV2/TokenCacheV2.cs
```diff
@@ -0,0 +1,234 @@
...
+            if (storageManager == null)
+            {
+                throw new ArgumentNullException(nameof(storageManager));
+            }
+
+            lock (_lock)
+            {
+                _storageManager = storageManager;
+
+                // If the user has deserialized any data into the cache before binding...
```
henrik-me, line 190: >If the user has deserialized any data into the cache before binding... [](start = 19, length = 70)

per above. not sure why this would not always be bound, probably my real question #Closed

msal/src/Microsoft.Identity.Client/CacheV2/Impl/Jwt.cs
```diff
@@ -0,0 +1,63 @@
...
+            Payload = Base64UrlHelpers.DecodeToString(sections[1]);
+            Json = JObject.Parse(Payload);
+            IsSigned = !string.IsNullOrEmpty(sections[2]);
+        }
+
+        protected JObject Json { get; }
+        public string Raw { get; }
+        public string Payload { get; }
+        public bool IsSigned { get; }
+        public bool IsEmpty => Json.IsEmpty();
```
henrik-me, line 61: nit: are these kept here because that is the C++ style guide?  #Closed

msal/src/Microsoft.Identity.Client/CacheV2/Impl/Jwt.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (string.IsNullOrWhiteSpace(raw))
+            {
+                // warning: constructed jwt from empty string
+                return;
+            }
+
+            string[] sections = Raw.Split('.');
+            if (sections.Length != 3)
+            {
+                throw new ArgumentException("failed jwt decode: wrong number of sections", nameof(raw));
```
henrik-me: >"failed jwt decode: wrong number of sections" [](start = 44, length = 45)

goal other places has been to keep strings in one file (is easier to review as a whole and it will be easier to make general updates to all errors, messages, etc. #Pending

msal/src/Microsoft.Identity.Client/CacheV2/Impl/Jwt.cs
```diff
@@ -0,0 +1,63 @@
...
+        public Jwt(string raw)
+        {
+            Raw = raw;
+            if (string.IsNullOrWhiteSpace(raw))
+            {
+                // warning: constructed jwt from empty string
+                return;
+            }
+
+            string[] sections = Raw.Split('.');
```
henrik-me, line 46: nit: slight inconsistency in whether the param raw is used or the property Raw is used. #Closed

msal/src/Microsoft.Identity.Client/CacheV2/Impl/PathStorageWorker.cs
```diff
@@ -0,0 +1,551 @@
...
+
+        public byte[] Encrypt(string input)
+        {
+            return Encoding.UTF8.GetBytes(input);
+        }
+
+        private byte[] Decrypt(byte[] input)
+        {
+            return input;
+        }
```
henrik-me, line 543: assuming there should be a feature flag or similar to flip encryption on/off. #Closed

msal/src/Microsoft.Identity.Client/CacheV2/Impl/PathStorageWorker.cs
```diff
@@ -0,0 +1,551 @@
...
+                credential.Environment,
+                credential.Realm,
+                credential.ClientId,
+                credential.FamilyId,
+                credential.CredentialType);
+        }
+
+        public void AddAccessTokenWithScopes(JObject accessTokens, JObject tokenToAdd)
+        {
+            if (!tokenToAdd.ContainsKey(StorageJsonKeys.Target))
```
henrik-me, line 382: >StorageJsonKeys.Target [](start = 40, length = 22)

is this really the returned scopes? or what does target mean? in that case, imo this should be named returnedScopes or similar? (potentially targetScopes if that makes sense in a bigger context) #Closed

henrik-me:
there are a few places of inconsistency of placing var's before or after mehtods, also there are a few places where there is both an enum and a class in the same file (and I believe also two classes in one file). Consider having one file for each class/enum and unify the field locations in the files. #Closed

henrik-me:
as discussed logging and exceptions are not part of this pass. will be added later. also assuming various todo's are to be done in an upcoming pass. #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/InMemoryCachePathStorageTests.cs
```diff
@@ -0,0 +1,389 @@
...
+
+            _io.DeleteFile(FileName);
+            Assert.IsFalse(_io.RootDirectory.FileExists(FileName));
+        }
+
+        // Tests FileIO::DeleteFile() and FileIO::Write() concurrently
+        [TestMethod]
+        public void WriteDeleteConcurrently()
+        {
+            Debug.WriteLine("TEST START: WriteDeleteConcurrently");
```
henrik-me: >Debug.WriteLine("TEST START: WriteDeleteConcurrently"); [](start = 12, length = 55)

we are not doing this in other tests. something we are going to do elsewhere going forward? Any reason the test framework is not doing enough already? ... looking for a purpose to have this. #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/InMemoryCachePathStorageTests.cs
```diff
@@ -0,0 +1,389 @@
...
+
+            var actions = new List<Action>();
+            for (int i = 0; i < 100; i++)
+            {
+                actions.Add(
+                    () =>
+                    {
+                        Thread.Sleep(new Random().Next() % 10);
```
henrik-me, line 110: >(new Random().Next() [](start = 36, length = 20)

is this going to be "truly" random? likely not important though. #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/InMemoryCachePathStorageTests.cs
```diff
@@ -0,0 +1,389 @@
...
+                        byte[] actualData = _io.Read(FileName);
+
+                        // Either the file doesn't exist
```
henrik-me: >Either [](start = 27, length = 6)

the remaining part of the sentence is missing. 
Either the file doesn't exist or the file is empty ? #Pending

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/InMemoryCachePathStorageTests.cs
```diff
@@ -0,0 +1,389 @@
...
+    {
+        private const string FileName = "FileIOTests.bin";
+        private const string TestFolderBase = "FileIOTestsFolder";
+        private byte[] _data;
+        private InMemoryCachePathStorage _io;
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            TestCleanup();
```
henrik-me: why would this be called first? I suppose we should assume that the tests are properly cleaning up? (aside cleanup is not really doing anything for the moment). Given this is the memory test I assume that as soon as we are out of the scope of the test the GC will handle the cleanup.... however for the file persistence layer I'm assuming proper cleanup is required? #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/JObjectAssert.cs
```diff
@@ -0,0 +1,67 @@
...
+            foreach (KeyValuePair<string, JToken> kvpExpected in expected)
+            {
+                if (actual.TryGetValue(kvpExpected.Key, out var value))
+                {
+                    Assert.AreEqual(kvpExpected.Value.Type, value.Type);
+                    Assert.AreEqual(kvpExpected.Value, value);
+                }
+                else
+                {
+                    Assert.Fail();
```
henrik-me: for every Assert.Fail .... would it make sense to add a message as to what went wrong? ... I suppose the stack trace will tell however perhaps initial troubleshooting will be easier? #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/TokenCacheV2Tests.cs
```diff
@@ -0,0 +1,126 @@
...
+namespace Test.MSAL.NET.Unit.net45.CacheV2Tests
+{
+    [TestClass]
+    public class TokenCacheV2Tests
+    {
+        private InMemoryCachePathStorage _cachePathStorage;
+        private FileSystemCredentialPathManager _credentialPathManager;
+        private StorageManager _storageManager;
+        private PathStorageWorker _storageWorker;
+        private TokenCacheV2 _tokenCache;
```
henrik-me, line 50: consider if these should really be part of the class or if the test itself should have an instance. as there is only one test right now I suppose it doesn't matter... but perhaps later it will. #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/FileIOTests.cs
```diff
@@ -0,0 +1,779 @@
...
+        {
+            if (Directory.Exists(TestFolderBase))
+            {
+                Directory.Delete(TestFolderBase, true);
+            }
+
+            if (File.Exists(TestFolderBase))
+            {
+                File.Delete(TestFolderBase);
+            }
```
henrik-me, line 73: Seems like Directory.Delete(,true) should be able to able to handle deleting both files and sub folders?
aside: if not the next line should probably have been first. #Closed

core/src/Platforms/net45/NetDesktopCryptographyManager.cs
MarkZuber: good catch.  fixed #Closed

core/src/Platforms/net45/NetDesktopCryptographyManager.cs
MarkZuber: good catch.  fixed #Closed

core/src/Platforms/uap/UapCryptographyManager.cs
MarkZuber: fixed #Closed

msal/src/Microsoft.Identity.Client/CacheV2/TokenCacheV2.cs
MarkZuber: Because in our current model, the TokenCache is constructed by the developer using our libraries, so when it's constructed none of this is bound up.  Eventually TokenCacheV2 will replace the public TokenCache object for public consumption and we need to handle them constructing it and then handing it to us (where we then bind up the types). #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/FileIOTests.cs
MarkZuber: Directory.Delete doesn't work if you pass it a filename.  And sometimes we're deleting a file (we have a test that creates a file named what the folder should be and ensure we fail properly) and other times we're deleting a folder (the normal case). #Closed

core/src/Platforms/netcore/NetCoreCryptographyManager.cs
MarkZuber: Fixed #Closed

msal/src/Microsoft.Identity.Client/CacheV2/ITokenCacheAdapter.cs
MarkZuber: account.  fixed. #Closed

msal/src/Microsoft.Identity.Client/CacheV2/Schema/Credential.cs
MarkZuber: This is pretty standard for the Equals comparison method.  It means "the other one you're comparing to" #Closed

msal/src/Microsoft.Identity.Client/CacheV2/CacheManager.cs
MarkZuber: If they come back empty, then they come back empty and we'll return empty fields for them.  This is all of the id token fields that were in C++ and we're going for functional consistency so I have them included here. #Closed

msal/src/Microsoft.Identity.Client/CacheV2/Impl/Jwt.cs
MarkZuber: I don't understand your question.  Are you asking about the placement of the public properties?  They're all listed after the constructor.  That's normal in our code as well, and the formatter puts them this way too... #Closed

msal/src/Microsoft.Identity.Client/CacheV2/Impl/Jwt.cs
MarkZuber: All exceptions/logging/errors need a 2nd round cleanup pass under Cachev2/... #Pending

msal/src/Microsoft.Identity.Client/CacheV2/Impl/Jwt.cs
MarkZuber: fixed #Closed

msal/src/Microsoft.Identity.Client/CacheV2/Impl/PathStorageWorker.cs
MarkZuber: The field in the json needs to be "target" and this property is named Target for consistency with what's being put there.  I think they named it Target instead of Scopes because the cache can store credentials that aren't OAuth2 compliant (and thus don't have scopes) #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/InMemoryCachePathStorageTests.cs
MarkZuber: I'll remove these, they were there for some initial debugging... #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/InMemoryCachePathStorageTests.cs
MarkZuber: I don't think it's critical that this is crypto-random or anything that crazy. #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/InMemoryCachePathStorageTests.cs
MarkZuber: Fixed and cleaned up #Pending

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/InMemoryCachePathStorageTests.cs
MarkZuber: Remnants of earlier debugging.  Since TestCleanup is also empty, i've removed TestCleanup and the call here. #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/JObjectAssert.cs
MarkZuber: Yup, great feedback.  fixed. #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CacheV2Tests/TokenCacheV2Tests.cs
MarkZuber: Yes, I expect this test class to change a bit as I roll in the rest of the tests.  I'm going to not fix this now but I'll keep it in mind as I bring in the rest of the test coverage here. #Closed

MarkZuber:
> there are a few places of inconsistency of placing var's before or after mehtods, also there are a few places where there is both an enum and a class in the same file (and I believe also two classes in one file). Consider having one file for each class/enum and unify the field locations in the files.

I've fixed the enums and moved them to their own classes.  FYI:  there are many places in the code where the enum is in the file the enum is most related to.  In the past I've traditionally not always had enums in their own files but I'm fine having that convention.  We should consider making that consistent across the codebase if that's what we want.

I didn't see anywhere that I have more than one public/internal class except where I have nested private classes.  If you see something you think should be in a separate file, please point it out and i'll fix. #Closed

msal/src/Microsoft.Identity.Client/CacheV2/Impl/PathStorageWorker.cs
MarkZuber: added TODO block here #Closed

msal/src/Microsoft.Identity.Client/CacheV2/Schema/StorageJsonValues.cs
MarkZuber: It will need to be.  This is the first run and msalc++ doesn't support b2c now so it wasn't in there.  as we fill in scenarios and test cases we'll need to expand support here. #Closed

henrik-me (approved):
:shipit:

msal/src/Microsoft.Identity.Client/CacheV2/ITokenCache.cs
```diff
@@ -0,0 +1,83 @@
...
+    /// and before/after notifications used in that scenario.
+    /// TODO: This interface should eventually be PUBLIC.
+    /// </summary>
+    internal interface ITokenCache
+    {
+        ///// <summary>
+        ///// 
+        ///// </summary>
+        ///// <param name="beforeAccess"></param>
+        //void SetBeforeAccess(TokenCacheNotification beforeAccess);
```
bgavrilMS, line 45: Why are the hooks commented out?

msal/src/Microsoft.Identity.Client/CacheV2/Impl/FileSystemCredentialPathManager.cs
```diff
@@ -0,0 +1,166 @@
...
+            return GetCredentialPath(
+                credential.HomeAccountId,
+                credential.Environment,
+                credential.Realm,
+                credential.ClientId,
+                credential.FamilyId,
+                credential.CredentialType);
+        }
+
+        public string ToSafeFilename(string data)
```
bgavrilMS, line 63: I wonder why this is required and what is the likelyhood / impact name of collisions that this brings.

msal/src/Microsoft.Identity.Client/CacheV2/Impl/IStorageManager.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    /// <summary>
+    /// Equivalence layer with MSAL C++ and other native platforms for handing read/write/query operations
+    /// on the various credential and account types in the unified cache.
+    /// Also provides (on msal.net) access to the Adal Legacy Cache Manager for ensuring legacy cache interop
+    /// during storage access.
+    /// </summary>
+    internal interface IStorageManager
```
bgavrilMS, line 39: I really like that they name things properly. Ours is not a token "cache", it's a store. Cache implies expiration etc.

msal/src/Microsoft.Identity.Client/CacheV2/ITokenCache.cs
MarkZuber: I was working on wiring them up but they're not ready for both implementations.  So they're here and will be enabled on a near-future checkin.

msal/src/Microsoft.Identity.Client/CacheV2/Impl/FileSystemCredentialPathManager.cs
MarkZuber: Valentin did a bunch of analysis with the base32 hex namespace and calculated out the collision rates.  I'll see if i can find anything about it and add it in as comments.  It's a crazy rare occurrence in the normal case of a few accounts and many products.

msal/src/Microsoft.Identity.Client/CacheV2/Impl/IStorageManager.cs
MarkZuber: <you're welcome> ;)

bgavrilMS (approved):

jennyf19 (approved):

PR #1434, opened 2018-11-27
Create ServiceBundle to reduce DI verbosity

MarkZuber:


bgavrilMS (approved):
LGTM

msal/tests/Test.MSAL.NET.Unit.net45/RequestsTests/InteractiveRequestTests.cs
```diff
@@ -52,28 +52,18 @@ namespace Test.MSAL.NET.Unit.RequestsTests
...
-            };
         }
 
         [TestCleanup]
         public void TestCleanup()
         {
-            _cache.TokenCacheAccessor.ClearAccessTokens();
-            _cache.TokenCacheAccessor.ClearRefreshTokens();
+            //_cache.TokenCacheAccessor.ClearAccessTokens();
+            //_cache.TokenCacheAccessor.ClearRefreshTokens();
```
jennyf19: what's going to happen here? #Closed

jennyf19 (approved):
cool.

msal/tests/Test.MSAL.NET.Unit.net45/CoreTests/InstanceTests/AadAuthorityTests.cs
```diff
@@ -304,6 +300,8 @@ public void FailedTenantDiscoveryMissingEndpointsTest()
         [TestCategory("AadAuthorityTests")]
         public void CanonicalAuthorityInitTest()
         {
+            var serviceBundle = ServiceBundle.CreateForProduction();
```
henrik-me: >CreateForProduction [](start = 46, length = 19)

is this on purpose? if so why is it different from the others? #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CoreTests/InstanceTests/B2cAuthorityTests.cs
```diff
@@ -64,15 +59,14 @@ public void NotEnoughPathSegmentsTest()
         {
             try
             {
-                var instance = Authority.CreateAuthority(_validatedAuthoritiesCache, _aadInstanceDiscovery, "https://login.microsoftonline.in/tfp/", false);
+                var serviceBundle = ServiceBundle.CreateForProduction();
```
henrik-me: for test? why not? #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CoreTests/InstanceTests/B2cAuthorityTests.cs
```diff
@@ -171,6 +163,8 @@ public void B2CMicrosoftOnlineCreateAuthority()
         [TestCategory("B2CAuthorityTests")]
         public void CanonicalAuthorityInitTest()
         {
+            var serviceBundle = ServiceBundle.CreateForProduction();
```
henrik-me: >CreateForProduction [](start = 46, length = 19)

would really be great to understand why some tests does one thing and others the other thing. #Closed

core/src/ServiceBundle.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public static ServiceBundle CreateForProduction(ITelemetryReceiver telemetryReceiver = null)
+        {
+            return new ServiceBundle(telemetryReceiver: telemetryReceiver);
+        }
+
+        public static ServiceBundle CreateForTest(IHttpManager httpManager, ITelemetryReceiver telemetryReceiver = null)
+        {
+            return new ServiceBundle(httpManager: httpManager, telemetryReceiver: telemetryReceiver, shouldClearCaches: true);
+        }
```
henrik-me, line 85: based on the current test code I would like that these two have a description on when to be used and what the difference really is. Perhaps instead of test and prod those should be named CreateDefaultBundle() and CreateBundleWithCleanCache() ? #Closed

msal/tests/Test.MSAL.NET.Unit.net45/RequestsTests/InteractiveRequestTests.cs
henrik-me: I suppose the cache has been removed, thus the clearnup can be removed as well?

---
In reply to: [236820555](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1434#discussion_r236820555) [](ancestors = 236820555)

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -108,15 +108,15 @@ public AuthenticationContext(string authority, bool validateAuthority, TokenCach
...
             TokenCache tokenCache)
         {
             // If authorityType is not provided (via first constructor), we validate by default (except for ASG and Office tenants).
             this.Authenticator = new Authenticator(authority, (validateAuthority != AuthorityValidationType.False));
             this.TokenCache = tokenCache;
 
-            _httpManager = httpManager ?? new HttpManager();
-            _wsTrustWebRequestManager = new WsTrustWebRequestManager(_httpManager);
+            _serviceBundle = serviceBundle ?? ServiceBundle.CreateForProduction();
+            _wsTrustWebRequestManager = new WsTrustWebRequestManager(_serviceBundle);
```
henrik-me: >_wsTrustWebRequestManager = new WsTrustWebRequestManager(_serviceBundle); [](start = 12, length = 73)

would it make sense to put this in the ServiceBundle as well even though it will be kind of a fun "dependency"? #Closed

henrik-me (changes requested):
:clock1:

henrik-me:
name suggestions (instead of ServiceBundle): DependencyContainer, both service and bundle seems off by a bit to me (though there could also be more put into the DI Container term than what is being used here). #Closed

bgavrilMS:
Name suggestion: I like ServiceBundle, "bundle" is a widely used term in the javascript world where it means "package all the code and all the dependencies in one file". Maybe DepedencyBundle if ppl think "Service" is too overloaded. I don't link DepdencyContainer because this term is used by proper DI libraries - this is not DI, it's not even a ServiceLocator.

msal/tests/Test.MSAL.NET.Unit.net45/RequestsTests/InteractiveRequestTests.cs
MarkZuber: Correct, the cache isn't a member so we'll create a new one.  i think these were unnecessary even previously since we were recreating the _cache variable during TestInitialize.  I've removed the dead code now. #Closed

msal/tests/Test.MSAL.NET.Unit.net45/CoreTests/InstanceTests/AadAuthorityTests.cs
MarkZuber: Yes, this is on purpose.  CreateForTest takes in a MockHttpManager.  This doesn't have/need one so we just create a Production servicebundle.  I could change this name to CreateDefault and CreateWithMockHttp if that makes this usage clearer. #Closed

core/src/ServiceBundle.cs
MarkZuber: I've renamed them to CreateDefault and CreateWithCustomHttpManager since those are the actual usages. #Closed

adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
MarkZuber: done #Closed

henrik-me:
I'm ok with DependencyBundle ... at least it makes it more clear to me what it is. But not a strong point either.

---
In reply to: [442407487](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1434#issuecomment-442407487) [](ancestors = 442407487)

henrik-me (approved):
:shipit:

PR #1435, opened 2018-11-27
(Partial PR) Add ClaimsPrincipalExtension methods to MSAL

bgavrilMS:


msal/src/Microsoft.Identity.Client/ClaimsPrincipalExtension.cs
```diff
@@ -0,0 +1,108 @@
...
+
+using System;
+using System.Security.Claims;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Extension methods for ClaimsPrincipal
+    /// </summary>
+    public static class ClaimsPrincipalExtension
```
henrik-me, line 36: >ClaimsPrincipalExtension [](start = 24, length = 24)

what about .net desktop?

msal/src/Microsoft.Identity.Client/ClaimsPrincipalExtension.cs
```diff
@@ -0,0 +1,108 @@
...
+                displayName = claimsPrincipal.FindFirstValue("name");
+            }
+            return displayName;
+        }
+
+         /// <summary>
+        ///     Return the claim value for the first claim with the specified type if it exists, null otherwise
+        /// </summary>
+        /// <param name="principal"></param>
+        /// <param name="claimType"></param>
```
henrik-me, line 96: empty params and indentation seems a bit off

msal/src/Microsoft.Identity.Client/ClaimsPrincipalExtension.cs
```diff
@@ -0,0 +1,108 @@
...
+//
+//------------------------------------------------------------------------------
+
+using System;
+using System.Security.Claims;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Extension methods for ClaimsPrincipal
```
henrik-me, line 34: >/ Extension methods for ClaimsPrincipal [](start = 6, length = 39)

While I understand the benefit it can bring, what are the scenarios? For web only? Also what is the next step?

henrik-me (changes requested):
:clock1:

msal/src/Microsoft.Identity.Client/ClaimsPrincipalExtension.cs
jmprieur: that was not in my file ? and the argument exception seems wrong?

msal/src/Microsoft.Identity.Client/ClaimsPrincipalExtension.cs
jmprieur: It's for web, when ASP.NET sets a ClaimPrincipal for the signed-in user, available everywhere in the application. If customers want to get a token for an account, it's a confort method to get it from the claims principal.
not sure for DisplayName, BTW.

msal/src/Microsoft.Identity.Client/ClaimsPrincipalExtension.cs
bgavrilMS: @jmprieur - this was not in your file, indeed. You were using an ASP.NET extension method for FindFirstValue. We definitely don't want to take a dependency on ASP.NET so I wrote my own. I'll delete the comments and fix the arg null.

msal/src/Microsoft.Identity.Client/ClaimsPrincipalExtension.cs
bgavrilMS: I don't understand your question - this class is now compiled on all our target frameworks, including .net classic. Should it be added only to .net core / .net standard?

msal/src/Microsoft.Identity.Client/ClaimsPrincipalExtension.cs
bgavrilMS: @jmprieur, @henrik-me  - let's decide on whether we want both, one or none of these methods before cleaning up / adding unit tests.

msal/src/Microsoft.Identity.Client/ClaimsPrincipalExtension.cs
jmprieur: ah :) that explains :) Thanks @bgavrilMS

msal/src/Microsoft.Identity.Client/ClaimsPrincipalExtension.cs
jmprieur: yes. Let's decide first. I'm starting to think that we would rather have another library:
Microsoft.Identity.Client.AspNetWeb or something like that.

msal/src/Microsoft.Identity.Client/ClaimsPrincipalExtension.cs
henrik-me: IMO, this only makes sense in web scenarios. perhaps I'm missing something.

---
In reply to: [237044980](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1435#discussion_r237044980) [](ancestors = 237044980)

jmprieur: indeed it's only for Web scenarios.

bgavrilMS:
Abandoning this, it was meant as a proof of concept, not approved.

PR #1443, opened 2018-11-30
PublicClientApp unified NetStandard API

bgavrilMS:


msal/src/Microsoft.Identity.Client.Ref/Microsoft.Identity.Client.Ref.csproj
```diff
@@ -11,22 +11,22 @@
...
-    <DefineConstants>$(DefineConstants);DESKTOP_BUILD</DefineConstants>
+    <DefineConstants>$(DefineConstants);DESKTOP_BUILDTIME</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFramework)' == 'Xamarin.iOS10'">
-    <DefineConstants>$(DefineConstants);iOS_BUILD</DefineConstants>
+    <DefineConstants>$(DefineConstants);iOS_BUILDTIME</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard1.3'">
-    <DefineConstants>$(DefineConstants);NETSTANDARD_BUILD</DefineConstants>
+    <DefineConstants>$(DefineConstants);NETSTANDARD_BUILDTIME</DefineConstants>
```
MarkZuber, line 29: does this need to be netstandard_13_buildtime?  would this behavior change if we changed our minbar to netstandard15?  i'm assuming (HOPING) we only support one netstandard min level and not multiple netstandard targets.

MarkZuber (approved):

msal/src/Microsoft.Identity.Client.Ref/Microsoft.Identity.Client.Ref.csproj
bgavrilMS: No, I don't think we'd need to change anythin if we decided to support NetStandard2.0 instead. There are some default flags (NETSTANDARD_13, NETSTNANDARD_20) that would change, but we define our own flags so this is ok.

Btw, on ADAL we support 2 netstandard targets - 1.1 and 1.3. We really wanted to make sure we don't break anybody :).

PR #1446, opened 2018-11-30
enable username password request everywhere but ios/android

MarkZuber:
This is helping out with Bogdan's work on the NetStandard unification.  I've made a feature flag for username password and enabled it in the proper projects and ifdef'd the code based on the feature flag.


msal/src/Microsoft.Identity.Client/PublicClientApplication.cs
```diff
@@ -509,6 +509,35 @@ internal IWebUI CreateWebAuthenticationDialog(UIParent parent, UIBehavior behavi
         // endif for !NET_CORE
 #endif
 
+#if MSAL_FEATURE_USERNAMEPASSWORD
+        /// <summary>
+        /// Non-interactive request to acquire a security token from the authority, via Username/Password Authentication.
+        /// Available only on .net desktop and .net core. See https://aka.ms/msal-net-up for details.
```
jmprieur, line 515: Not available on Xamarin.iOS and Xamarin.Android. See https://aka.ms/msal-net-up for details 

We'll have to update https://aka.ms/msal-net-up BTW

jmprieur (approved):
LGTM.
- A comment to update.
- interesting the approach of having a #define by feature and centralizing in the MSAL.Common.props which features are implemented in which platform. This is way simpler to read. Of course between this and the files only included in some platforms, this is still complicated to get the big picture by looking at the source code :)

msal/src/Microsoft.Identity.Client/IPublicClientApplication.cs
```diff
@@ -288,5 +288,20 @@ public partial interface IPublicClientApplication : IClientApplicationBase
         // endif !NOT_CORE
 #endif
 
+#if MSAL_FEATURE_USERNAMEPASSWORD
```
jennyf19, line 291: do we have to say `MSAL` can it just be `FEATURE`?

MarkZuber:
This may also be blocked behind the split since we need proper handling for UsernamePasswordInput which is in core, but I can't put this code behind the proper feature flag since ADAL doesn't support SecureString in all of the scenarios.  So we can't use the existing #if NETCORE / NET45 since that will break how MSAL wants to include these.

msal/src/Microsoft.Identity.Client/IPublicClientApplication.cs
bgavrilMS: I like MSAL_FEATURE because if we have to add it to core files, it will clearly disambiguate between an ADAL and an MSAL feature

msal/src/MSAL.Common.props
```diff
@@ -47,16 +47,22 @@
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp1.0' ">
     <DefineConstants>$(DefineConstants);NET_CORE</DefineConstants>
+    <!-- Feature Flags -->
+    <DefineConstants>$(DefineConstants);MSAL_FEATURE_USERNAMEPASSWORD</DefineConstants>
```
bgavrilMS, line 51: Technically this is pretty cool, and a lot of libraries do this. But if our goal is to have a unified API across all tfms, we'll have to include each and every feature to each tfm, and hide some APIs from Intellisense / Build.

bgavrilMS:
@MarkZuber - currently the flags iOS_RUNTIME / iOS_BUILDTIME are only defined in MSAL. )We might want to prefix them with MSAL_). So you could expose U/P without affecting ADAL. I agree that we'll have to take a reference to SecureString from Android and iOS, which is a bit wastefull because we'd only use it so that we can throw a nice "platform not supported' exception. Not sure if reference footprint is that big of an issue (if it is, don't use xamarin :D).

PR #1447, opened 2018-12-03
[NetStandard Unification] Confidential Client

bgavrilMS:


msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
```diff
@@ -78,6 +80,7 @@ static ConfidentialClientApplication()
             ClientCredential clientCredential, TokenCache userTokenCache, TokenCache appTokenCache)
             : this(clientId, DefaultAuthority, redirectUri, clientCredential, userTokenCache, appTokenCache)
         {
+            GuardMobileFrameworks();
```
MarkZuber, line 83: aside thought (not pending for this PR):  Going forward with the config object, if we have the config object builder actually return a IPublicClientApplication / IConfidentialClientApplication then we can have it return implementations that throw per platform instead of having to embed it in the default type

MarkZuber (approved):

msal/src/Microsoft.Identity.Client/Features/ConfidentialClient/ConfidentialClientApplication.cs
bgavrilMS: Yes, that will work well particularly for CCA. For PCA, platform support is more heterogeneous - platform X supports IWA, platform Y supports IWA and U/P etc.

PR #1448, opened 2018-12-03
 [NetStandard Unification] IsSystemWebviewAvailable implementation

bgavrilMS:


msal/src/Microsoft.Identity.Client/Platforms/netstandard1.3/UIParent.cs
```diff
@@ -67,11 +67,20 @@ public UIParent(object parent, bool useEmbeddedWebview)
             ThrowPlatformNotSupported();
         }
 
+        /// <summary>
+        /// Checks if the system weview can be used. 
+        /// </summary>
+        public static bool IsSystemWebviewAvailable() // This is part of the NetStandard "interface" 
+        {
+            ThrowPlatformNotSupported();
```
MarkZuber, line 75: If this is a capability check, why should this throw?  This should just return false.

msal/src/Microsoft.Identity.Client/Platforms/iOS/UIParent.cs
```diff
@@ -76,5 +76,15 @@ public UIParent(bool useEmbeddedWebview) : this()
         {
         }
 #endif
+
+        /// <summary>
+        /// Checks if the system weview can be used. 
+        /// Currently, on iOS, only the embedded webview is available, so this always returns false.
+        /// </summary>
+        public static bool IsSystemWebviewAvailable() // This is part of the NetStandard "interface"
```
jennyf19, line 84: we can use system webview on iOS as well. This is for Android only because if Chrome custom tabs is not available on the device, system webview may not work, so instead of the library making the decision for the developer to switch automatically from system webview to embedded webview, we add this to the api so the developer can check the device first and then switch to embedded webview. An iOS developer should never call this method. #Closed

bgavrilMS: Great, thanks Jenny. I will have iOS return true here. 

I understand why you say that an iOS developer should never have to call this method, but when developing a Xamarin app, developers are writing code against NetStandard, but it is ran against MSAL.Android, MSAL.iOS etc. So this needs to exist on iOS as well.  #Closed

jennyf19 (approved):
one comment, but looks good.

msal/src/Microsoft.Identity.Client/Platforms/iOS/UIParent.cs
jennyf19: Cool. I was thinking more for the comment, you could add that an iOS dev shouldn't call this method (in case someone actually reads it). excited to have these changes in! will help a lot of devs using xamarin.

---
In reply to: [238377971](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1448#discussion_r238377971) [](ancestors = 238377971)

PR #1461, opened 2018-12-13
Jennyf/multi resource

jennyf19:
@henrik-me Initial commit...just trying some things out. need to do more work, but pushed up what I have in case you want to take a look or had feedback (mainly for testing and what should be covered). I just did a basic silent utest w/ a different tenantId, but need to do more in-depth testing. also...need to review when i'm not hangry. 😸


src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -418,6 +418,8 @@ internal AdalResultWrapper LoadFromCacheCommon(CacheQueryData cacheQueryData, Re
                         // this is a cross-tenant result. use RT only
                         resultEx.Result.AccessToken = null;
 
+                        resultEx.Result.Authority = cacheQueryData.Authority;
```
henrik-me, line 421: >resultEx.Result.Authority = cacheQueryData.Authority; [](start = 24, length = 53)

would that be because it is generally saved with the migrated authority instead of what the user has entered? ... in that case it may not be a cross tenant token which was found but rather simply the base part of the authority which doesn't match... perhaps there should be validation that the tenant is the same vs. the authority base url? #Resolved

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -471,7 +473,7 @@ internal AdalResultWrapper LoadFromCacheCommon(CacheQueryData cacheQueryData, Re
                         resultEx = null;
                     }
 
-                    if (resultEx != null)
+                    if (resultEx?.Result.Authority == null)
                     {
                         resultEx.Result.Authority = cacheKey.Authority;
                         requestContext.Logger.Info("A matching item (access token or refresh token or both) was found in the cache");
```
henrik-me, line 479: >requestContext.Logger.Info("A matching item (access token or refresh token or both) was found in the cache"); [](start = 24, length = 109)

not sure I get this comment based on the if statement above... can you explain?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -604,8 +606,9 @@ internal AdalResultWrapper LoadFromCacheCommon(CacheQueryData cacheQueryData, Re
                 // this check only applies to user tokens. client tokens should be ignored.
                 if (returnValue == null && cacheQueryData.SubjectType != TokenSubjectType.Client)
                 {
+
                     List<KeyValuePair<AdalTokenCacheKey, AdalResultWrapper>> itemsForAllTenants = this.QueryCache(
```
henrik-me, line 610: nit: ...

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -628,9 +631,9 @@ internal AdalResultWrapper LoadFromCacheCommon(CacheQueryData cacheQueryData, Re
             }
         }
 
-        private static bool IsSameCloud(string authority, string authority1)
+        private static bool IsSameCloud(string cachedAuthority, string cacheQueryDataAuthority)
```
henrik-me, line 634: >cacheQueryDataAuthority [](start = 71, length = 23)

it's not really obvious (to me) what this is. I'm assuming this is the value which was provided by the user/developer as a parameter? (same comment above). perhaps renaming?

henrik-me:
Looks interesting. Added a few thoughts and wonderings.

henrik-me (changes requested):
:clock1:

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
jennyf19: i don't think this is needed.

---
In reply to: [241267403](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1461#discussion_r241267403) [](ancestors = 241267403)

henrik-me: Did you try your test with-out this one?

---
In reply to: [241910097](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1461#discussion_r241910097) [](ancestors = 241910097,241267403)

jennyf19: Yes. I wrote up something earlier today, but I think I forgot to hit "publish". The authority is not part of the data contract on the AdalResult, so when the Clone() is created and the auth result is deserialized, the authority is null. so resultEx.Result.Authority is always null and always returned as null in the silent call. I'm not sure if this is a problem though. That's what I need to figure out. and it's not related to the resource issue. I think they are separate.  #Resolved

henrik-me: Makes sense.

---
In reply to: [241939559](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1461#discussion_r241939559) [](ancestors = 241939559)

Issue #1463, opened 2018-12-17
When asking for a token to a different resource through a broker, the token cache deletes the original AT


bgavrilMS:
**Which Version of ADAL are you using ?**
Note that to get help, you need to run the latest preview or non-preview version
For MSAL, please log issues to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet
latest

**Which platform has the issue?**
Android, possibly iOS (?)

**What authentication flow has the issue?**
Other: AcquireTokenInternactive + AcquireTokenSilent via broker

**Repro**

1. AcquireTokenInteractive resource1 (via broker) 
**Actual**: adds an AT and IdT to the cache, but the RT (which is an MRRT) is stored by the broker. This is expected.

2. AcquireTokenSilent resource2 (via broker) - a silent broker calls is happening, but the original AT is deleted 

3. Look at the token cache. 

**Expected behavior**
The token cache should still contain the AT obtained from step1

**Actual behavior**
The token cache deletes the AT obtained from step1

**User Impact**
The impact is not significant because the logic falls back to a silent broker call will also produce a token. However this is slower than fetching the result from the token cache. 

**What is actually happening**

a. The tokenCache holds an entry, but this entry is not resource specific and [no MRRT exists](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/23fd79c5adeb01d4d01199a4cc68efdb306c2985/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs#L592) because borkers hold onto the MRRT 
b. The tokenCache continues to look at the [original AT](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/23fd79c5adeb01d4d01199a4cc68efdb306c2985/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs#L616)
c. The tokenCache continues to think there is an [MRRT available locally](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/23fd79c5adeb01d4d01199a4cc68efdb306c2985/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs#L429)
[...]

Issue #1353, opened 2018-10-25
Android + Broker - interactive auth is forced when requesting different resources


EldinHb:
**Which Version of ADAL are you using ?**
4.3

**Which platform has the issue?**
Xamarin.Android

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive

**Repro**
While using the broker to authenticate users the tokencache does not have a refreshtoken.

See code:
```csharp
if (authContext.TokenCache.ReadItems().Any())
    authContext = new AuthenticationContext(authContext.TokenCache.ReadItems().First().Authority);
else
    var authContext = new AuthenticationContext(authority);

var uri = new Uri(redirectUri);
var platformParams = new PlatformParameters(Plugin.CurrentActivity.CrossCurrentActivity.Current.Activity, true);
authResult = await authContext.AcquireTokenAsync(resource, clientId, uri, platformParams);
```

**Expected behavior**
TokenCache has a RefreshToken that is not null.

**Actual behavior**
RefreshToken in TokenCache is null when using a broker. When platformParams.UseBroker = false; the behavior is as expected and RefreshToken in TokenCache is not null.

jmprieur:
@Eldin66 : 
What are you trying to do? What is your scenario?

I believe that what you observe is the expected behavior as, when an app uses the broker, the Refresh token is really stored and refreshed by the broker itself. (ADAL delegates token acquisition to the broker)

Did you ever see that the Refresh token was not null? 
cc: @jennyf19

EldinHb:
@jmprieur 
In our scenario the user should be able to login with the broker (this works), then the token should get refreshed if necessary to reduce the amount of logins the user should do. However, the TokenCache RefreshToken is null so we cannot refresh the session meaning the user has to login every time the token expires or gets removed.

**Did you ever see that the Refresh token was not null?**
Yes, this works perfectly on Xamarin.iOS. Which is why I thought it is a bug on the Android version.

jmprieur:
@Eldin66 : @jennyf19 will confirm, but I believe that the reason why the refresh token is kept locally in ADAL.NET for Xamarin.iOS, is otherwise the application would be interrupted to display the broker app, and therefore the experience would not be good. On Android, this is not needed as apps can be called in the background, and therefore the Refresh token is kept in the broker.

[...]

PR #1465, opened 2018-12-18
Fix for bug #1463 and more logging when using broker

bgavrilMS:


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Broker/NullBroker.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Identity.Core.Cache;
+
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Broker
+{
+    internal class NullBroker : IBroker
```
bgavrilMS, line 13: Used everywhere except Android and iOS. This is very similar to the BrokerHelper that already existed on these platforms (and which does nothing), but now it implements an interface, avoiding potential MethodMissingException if we ever change the broker methods.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Broker/IBroker.cs
```diff
@@ -23,26 +27,16 @@
...
-using Microsoft.Identity.Core;
-using Microsoft.Identity.Core.Cache;
-using System.Collections.Generic;
-using System.Threading.Tasks;
+//
 
-namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
+namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Broker
 {
-    internal class BrokerHelper
```
bgavrilMS, line 35: All the broker changes should be pretty isolated. I needed to them because it was impossible to inject a logger into the BrokerProxy and also because it will make the code much easier to understand.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/AndroidBroker.cs
```diff
@@ -37,63 +37,55 @@
...
-        public RequestContext RequestContext { get; set; }
+        public AndroidBroker(ICoreLogger logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _brokerProxy =  new AndroidBrokerProxy(Application.Context, logger);
+        }
 
         public IPlatformParameters PlatformParameters { get; set; }
 
-        private bool WillUseBroker()
```
bgavrilMS, line 55: I place private methods at the bottom of the file.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/AndroidBroker.cs
```diff
@@ -218,18 +210,30 @@ internal static void SetBrokerResult(Intent data, int resultCode)
...
+            if (pp != null)
+            {
+                return pp.UseBroker;
+            }
+
+            return false;
         }
     }
+
+    //internal class CallBackHandler : Java.Lang.Object, IAccountManagerCallback
```
bgavrilMS, line 234: Not used, I'll delete it.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -426,23 +426,9 @@ internal AdalResultWrapper LoadFromCacheCommon(CacheQueryData cacheQueryData, Re
...
-                        {
-                            Result = new AdalResult(null, null, DateTimeOffset.MinValue),
-                            RefreshToken = resultEx.RefreshToken,
-                            ResourceInResponse = resultEx.ResourceInResponse
-                        };
-
-                        newResultEx.Result.UpdateTenantAndUserInfo(resultEx.Result.TenantId, resultEx.Result.IdToken,
-                            resultEx.Result.UserInfo);
-                        resultEx = newResultEx;
+                        resultEx = HandleMrrtCase(cacheQueryData, requestContext.Logger, resultEx, cacheKey);
```
bgavrilMS: This is the part of the bug fix - return null in MRRT case when MRRT is not in cache and request is for a different resource than what we have in the cache.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -463,7 +448,7 @@ internal AdalResultWrapper LoadFromCacheCommon(CacheQueryData cacheQueryData, Re
                             (resultEx.Result.ExpiresOn - DateTime.UtcNow).TotalMinutes));
                     }
 
-                    if (resultEx.Result.AccessToken == null && resultEx.RefreshToken == null)
+                    if (resultEx?.Result?.AccessToken == null && resultEx?.RefreshToken == null && tokenIsForSameResource)
```
bgavrilMS, line 477: This is the second part of the fix - do not delete the token for the original resource, it is clearly not expired.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -492,6 +477,47 @@ internal AdalResultWrapper LoadFromCacheCommon(CacheQueryData cacheQueryData, Re
...
+                string.Format(CultureInfo.CurrentCulture,
+                    "Multi resource refresh token for resource '{0}' will be used to acquire token for '{1}'.",
+                    cacheKey.Resource, cacheQueryData.Resource),
+                "Multi resource refresh token will be used to acquire a token");
+
+            bool mrrtIsInCache = resultEx.RefreshToken != null;
+
+            if (mrrtIsInCache)
+            {
+                logger.Info("MRRT will be used to fetch an access token. ");
```
bgavrilMS: This is the original logic -> no change (when MRRT is local)

tests/Test.ADAL.NET.Unit.net45/TokenCacheTests.cs
```diff
@@ -335,6 +336,29 @@ internal static void TokenCacheCrossTenantOperationsTest()
             AdalResultWrapper value = CreateCacheValue(null, "user1");
         }
 
+        internal static void MRRTWithBroker_Test()
```
bgavrilMS: I will delete this.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Broker/IBroker.cs
```diff
@@ -1,4 +1,8 @@
-﻿//----------------------------------------------------------------------
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
```
jennyf19, line 2: these look like they are in the wrong place

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/AndroidBrokerProxy.cs
```diff
@@ -206,48 +151,166 @@ public AdalResultWrapper GetAuthTokenInBackground(AuthenticationRequest request,
...
+        private void VerifyNotOnMainThread()
+        {
+            Looper looper = Looper.MyLooper();
+            if (looper != null && looper == _androidContext.MainLooper)
+            {
+                Exception exception = new AdalException(
+                    "Calling this from your main thread can lead to deadlock");
+                _logger.ErrorPii(exception);
+
+                if (_androidContext.ApplicationInfo.TargetSdkVersion >= BuildVersionCodes.Froyo)
```
jennyf19, line 274: >Froyo [](start = 90, length = 5)

what's this?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/AndroidBrokerProxy.cs
```diff
@@ -268,6 +331,7 @@ private AdalResultWrapper GetResultFromBrokerResponse(Bundle bundleResult)
                 {
                     // Initial request from app to Authenticator needs to launch
                     // prompt. null resultEx means initial request
+                    _logger.Info("BrokerProxy: Initial request - not returning a token");
```
jennyf19, line 343: >not returning a token [](start = 65, length = 21)

this is a little confusing. Maybe just "initial request" or add in part of the comment?

jennyf19:
when do you want to get this out to the customer?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/AndroidBrokerProxy.cs
bgavrilMS: I have no idea. My only changes were to move private methods below public methods and add logging. I did not touch anything else.

jennyf19: :joy: 

---
In reply to: [242632545](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1465#discussion_r242632545) [](ancestors = 242632545)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/AndroidBrokerProxy.cs
bgavrilMS: Good catch. I know it's confusing, but I don't understand what it means. I just added a logging message based on the comment in the code, so that I we can cross reference the log with the code.

bgavrilMS:
@jennyf19 - @trwalke may have found a fix for bug #1353 and he'll add it on top of this branch. We'll then do a private build and test it with some of the customers. Then release a bug fix release I think. Ideal situation - tomorrow :). Most likely after Christmas.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Broker/IBroker.cs
bgavrilMS: Fixed

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -407,32 +407,43 @@ internal AdalResultWrapper LoadFromCacheCommon(CacheQueryData cacheQueryData, Re
...
+                                            DateTime.UtcNow + TimeSpan.FromMinutes(ExpirationMarginInMinutes);
+                    bool tokenExtendedLifeTimeExpired = resultEx.Result.ExtendedExpiresOn <= DateTime.UtcNow;
+                    bool tokenIsForSameResource = cacheKey.ResourceEquals(cacheQueryData.Resource);
+                    bool rtPresent = resultEx.RefreshToken != null;
+
+                    // Handle Broker case - RT is held by the broker and the request is for a different resource
+                    if (!tokenIsForSameResource && !rtPresent)
+                    {
+                        // Can't use the AT returned by the cache - it's for a different resource
+                        // and ADAL does not have the RT to fetch a new AT
```
henrik-me, line 420: Is this true both for Android and iOS?

bgavrilMS: Very good point, I'll try it out on iOS and get back to you.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/AndroidBroker.cs
```diff
@@ -37,63 +37,55 @@
 using Microsoft.Identity.Core.Cache;
 using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.OAuth2;
 using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Helpers;
+using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Broker;
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform
 {
     [Android.Runtime.Preserve(AllMembers = true)]
-    internal class BrokerHelper
+    internal class AndroidBroker : IBroker
```
trwalke, line 45: Much better than having all of the BrokerHelper classes with different implementations. easier to find stuff.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
bgavrilMS: I can't test on iOS because I don't have a physical device and Apple does not allow to install apps on the emulators. I can work with Jenny or Travis to get this tested prior to the release.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -392,7 +392,7 @@ internal static async Task<List<string>> GetOrderedAliasesAsync(string authority
             return aliasedAuthorities;
         }
 
-        internal AdalResultWrapper LoadFromCacheCommon(CacheQueryData cacheQueryData, RequestContext requestContext)
+        internal /* internal for test */ AdalResultWrapper LoadFromCacheCommon(CacheQueryData cacheQueryData, RequestContext requestContext)
```
henrik-me, line 395: >/* internal for test */ [](start = 17, length = 23)

nit: is this comment needed? I'm assuming the test will fail if it isn't internal?

bgavrilMS: It's more so that ADAL developers don't call LoadFromCacheCommon from anywhere but this class. Normally this method should be private, but we need it to be internal so that tests can call it. 

Not the best design in terms of test hooks, but redesigning this class is not appropriate at this time in ADAL's lifecycle.

bgavrilMS:
Merged manually

PR #1468, opened 2018-12-19
Adding UID and package name to broker options/parameters on android 

trwalke:
Adding UID and package name to broker option on anroid to address Silent auth broker issue.
Changes approved by @bgavrilMS


trwalke:
Tested broker flows manually (interactive, silent, etc)

Issue #1462, opened 2018-12-13
Connection pooling appears broken when using AccessTokens with SQL Azure


MatthewSteeples:
**Which Version of ADAL are you using ?**
ADAL 4.4.1 (latest from NuGet)

**Which platform has the issue?**
net472 (Azure App Service)

**What authentication flow has the issue?**
`Task<AuthenticationResult> AcquireTokenAsync(string resource, IClientAssertionCertificate clientCertificate)`

**Repro**
We use this method to populate the AccessToken property of a SqlConnection object to talk to SQL Azure. This was working fine in 3.19.8, but when we updated the package (with no other code changes) to 4.4.1 we found that SQL connections were no longer pooled (so we quickly went up to using over 5000 connections when our baseline is less than 100).

```csharp
var clientId = "GuidFromAzureAD";
var cert = new X509Certificate2("AzureADCert");
ClientAssertionCertificate certCred = new ClientAssertionCertificate(clientId, cert);
AuthenticationContext authContext = new AuthenticationContext(authority, TokenCache.DefaultShared);
var token = await authContext.AcquireTokenAsync(resourceId, certCred);
var sqlConnection = new SqlConnection("SqlAzureConnectionString");
sqlConnection.Token = token.AccessToken;
```

**Expected behavior**
Connection pooling works as it did in 3.19.8

**Actual behavior**
Number of connections created increases dramatically

**Possible Solution**
Haven't looked yet, but the only resolution we came up with was to roll back to 3.19.8

jmprieur:
@MatthewSteeples : I understand what you write, but don't understand how this is related to ADAL.NET / authentication as ADAL.NET does not manage SQL connections.

Can you please tell me if you observe something different from the point of view of tokens between ADAL.NET 3.19.8 and ADAL.NET 4.4.1 ? For instance, do I understand that, when you are getting the token in ADA.NET 3.19.8, and assign it as the `AccessToken` of the sqlConnection, the token the exact same as in a previous call (unless it has to be refreshed) whereas now it's refreshed each time / not cached?

MatthewSteeples:
@jmprieur Thanks for your quick response. I enabled logging to try and narrow it down based on your questions, but it doesn't make much sense. Both versions retrieve the token from the cache (so do not communicate with AD) after the first request and both versions return the exact same AccessToken that is then passed in to SQL Server.

[...]

PR #1469, opened 2018-12-19
WIP - Implement clone method rather than relying on Serialization

MatthewSteeples:
See #1462 for details

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Core/Cache/AdalResult.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Core/Cache/AdalResult.cs
@@ -144,5 +144,18 @@ namespace Microsoft.Identity.Core.Cache
                 this.UserInfo = new AdalUserInfo(userInfo);
             }
         }
+
+        internal AdalResult Clone()
+        {
+            var result = new AdalResult(AccessTokenType, AccessToken, ExpiresOn, ExtendedExpiresOn)
+            {
+                Authority = Authority,
+                ExtendedLifeTimeToken = ExtendedLifeTimeToken,
+            };
+
+            result.UpdateTenantAndUserInfo(TenantId, IdToken, UserInfo);
+
+            return result;
+        }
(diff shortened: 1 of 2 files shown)


bgavrilMS:
The PR looks ok, but I feel that the entire logic of cloning that object is not needed. I've managed to find the original commit that introduced this - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/95e796b4390417f0518909e30109c75e4e3651aa  (in fact you can see that they also used a member-by-member clone originally). 

@MarkZuber - the Clone() is only used in the TokenCache. I only see a string verification and a bunch of unrelated checks. I'm thinking we can remove this Clone logic completely ? Otherwise, Matthew's fix is good.

CCing @jennyf19 as she approved the original PR 2 years ago, so she should remember all the details 😼

MarkZuber:
@bgavrilMS -- Looking at this.  we're this code is returning the AdalResult directly out of the cache, and that object is mutable.  So if it's not cloned, and we hand it out to a caller, they could externally modify the cache since it's the same reference.  Also, it's modified directly after the clone to clear things out if the token is expired and such.

I'm still wondering about the root cause here around how the strings/values in this object matter when all that's being sent into SqlConnection is the AccessToken string, which compares equally.  I feel like we're missing something else that's going on.

bgavrilMS:
@MatthewSteeples - is this still actual ? We are releasing a new version of ADAL soon and were wondering whether to include this tactical fix

MatthewSteeples:
@bgavrilMS The code as part of this pull request should still be usable, as it doesn't change any behaviour of the library (just changes the method of cloning) so it should be safe to include

I do agree with @MarkZuber though, that we're definitely fixing a symptom rather than the underlying cause. I've tried looking at the source code of the framework and stepping through it manually but it all seems to be based on hashcodes (which should be identical for strings with equal contents, even if the strings are referentially different) so I can't determine what is causing this behaviour

In summary, I'm happy for (and would appreciate) this change being included

Issue #1459, opened 2018-12-11
HttpClient is disposed every time


rokondra:
**ADAL 3.19.8**

**Which platform has the issue?**
net45

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [x] OBO, I guess. The one that uses AAD applications and certificates

**Repro**
Inside Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Http.HttpClientWrapper
```csharp
	public async Task<IHttpWebResponse> GetResponseAsync()
	{
		using (HttpClient client = new HttpClient(HttpMessageHandlerFactory.GetMessageHandler(this.UseDefaultCredentials)))
		{

```

**Expected behavior**
It is more like a question, if this behavior is expected or not.
I would expect re-usable single HttpClient instance, so that it won't re-open/re-negotiate everything required everytime you request a new token (specifically in multi-tenant scnearios)

**Actual behavior**
Whenever the HttpClient is disposed, connection is closed. A new connection is required and all the overhead associated with that.

**Possible Solution**
Re-use the HttpClient instance.

**Additional context/ Logs / Screenshots**
Add any other context about the problem here, such as logs and screebshots. Logging is described at https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/wiki/Logging-in-ADAL.Net

MatthewSteeples:
Internet seems to agree with you - https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/

jmprieur:
This was fixed in ADAL 4.x (which is source compatible with 3.19.8)
Also MSAL.NET does even better.

rokondra:
> This was fixed in ADAL 4.x (which is source compatible with 3.19.8)
> Also MSAL.NET does even better.

We took the latest available ADAL 4.4.2.0 and the issue seems still there, are we missing something?
```
var authenticationContext = new AuthenticationContext("https://login.windows.net/tenantid");
var certBase64 = "";

X509Certificate2 cert = new X509Certificate2(Convert.FromBase64String(certBase64), "<password>");
ClientAssertionCertificate clientAssertion = new ClientAssertionCertificate("appId", cert);

[...]

PR #1472, opened 2019-01-02
Allow HttpClient to be provided when creating AuthenticationContext

AaronPatterson:
This is a fix for issue #1459

When interacting with multiple different tenants new instances of the AuthenticationContext need to be created for each tenant.  Under the covers this causes a new instance of an HttpClient to be created.  This usage pattern causes connections to be re-established for each tenant and this can cause problems when working with large number of different tenants.

To resolve this, a new constructor has been added to the AuthenticationContext which allows a HttpClient instance to be provided.  This allows the user of the SDK to control the reuse of the HttpClient in scenarios where this is needed.

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
@@ -26,6 +26,7 @@
 //------------------------------------------------------------------------------
 
 using System;
+using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.Identity.Core;
 using Microsoft.Identity.Core.Cache;
@@ -107,14 +108,35 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
         {
         }
 
+        /// <summary>
+        /// Constructor to create the context with the address of the authority and flag to turn address validation off.
+        /// Using this constructor, address validation can be turned off. Make sure you are aware of the security implication of not validating the address.
+        /// 
+        /// This constructor allows the HttpClient to be provided instead of having a new one created.  This can be useful when working with multiple different tenants.
+        /// </summary>
+        /// <param name="authority">Address of the authority to issue token.</param>
+        /// <param name="validateAuthority">Flag to turn address validation ON or OFF.</param>
+        /// <param name="tokenCache">Token cache used to lookup cached tokens on calls to AcquireToken</param>
+        public AuthenticationContext(string authority, bool validateAuthority, TokenCache tokenCache, HttpClient httpClient)
+            : this(null, authority, validateAuthority ? AuthorityValidationType.True : AuthorityValidationType.False,
+                   tokenCache, httpClient)
+        {
+        }
+
         internal AuthenticationContext(IServiceBundle serviceBundle, string authority, AuthorityValidationType validateAuthority,
-            TokenCache tokenCache)
+            TokenCache tokenCache, HttpClient httpClient = null)
         {
             // If authorityType is not provided (via first constructor), we validate by default (except for ASG and Office tenants).
             Authenticator = new Authenticator(authority, (validateAuthority != AuthorityValidationType.False));
             TokenCache = tokenCache;
 
-            _serviceBundle = serviceBundle ?? ServiceBundle.CreateDefault();
+            IHttpClientFactory httpClientFactory = null;
+            if (httpClient != null)
+            {
+                httpClientFactory = new HttpClientFactory(httpClient);
+            }
+
+            _serviceBundle = serviceBundle ?? new ServiceBundle(httpClientFactory: httpClientFactory);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -107,14 +108,35 @@ public AuthenticationContext(string authority, bool validateAuthority, TokenCach
...
+        /// <summary>
+        /// Constructor to create the context with the address of the authority and flag to turn address validation off.
+        /// Using this constructor, address validation can be turned off. Make sure you are aware of the security implication of not validating the address.
+        /// 
+        /// This constructor allows the HttpClient to be provided instead of having a new one created.  This can be useful when working with multiple different tenants.
+        /// </summary>
+        /// <param name="authority">Address of the authority to issue token.</param>
+        /// <param name="validateAuthority">Flag to turn address validation ON or OFF.</param>
+        /// <param name="tokenCache">Token cache used to lookup cached tokens on calls to AcquireToken</param>
+        public AuthenticationContext(string authority, bool validateAuthority, TokenCache tokenCache, HttpClient httpClient)
```
bgavrilMS, line 120: While I am not opposed to the implementation on functional grounds, adding a new constructor is frawned upon in ADAL because users already find the API hard to use. Still, I understand that not being able to specify a custom HttpClient is preventing some Xamarin users from using ADAL. 

@jmprieur @henrik-me - please advise.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -107,14 +108,35 @@ public AuthenticationContext(string authority, bool validateAuthority, TokenCach
...
+
         internal AuthenticationContext(IServiceBundle serviceBundle, string authority, AuthorityValidationType validateAuthority,
-            TokenCache tokenCache)
+            TokenCache tokenCache, HttpClient httpClient = null)
         {
             // If authorityType is not provided (via first constructor), we validate by default (except for ASG and Office tenants).
             Authenticator = new Authenticator(authority, (validateAuthority != AuthorityValidationType.False));
             TokenCache = tokenCache;
 
-            _serviceBundle = serviceBundle ?? ServiceBundle.CreateDefault();
```
henrik-me, line 117: >_serviceBundle = serviceBundle ?? ServiceBundle.CreateDefault(); [](start = 12, length = 64)

IMO: it would be more clean to call ServiceBundle.CreateWithHttpClient(), as we are handling the object creation in the ServiceBundle class and not here.  Also you have to CreateDefault if httpClient is null, this code can't tell if that is what is happening or not.

henrik-me (changes requested):
:clock1:

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
henrik-me: Overall, given ADAL doesn't have the ConfigObject, I think this is acceptable, though I'm not sure it's enough for all scenarios we would need to support.

---
In reply to: [246794805](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1472#discussion_r246794805) [](ancestors = 246794805)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -107,14 +108,35 @@ public AuthenticationContext(string authority, bool validateAuthority, TokenCach
         {
         }
 
+        /// <summary>
+        /// Constructor to create the context with the address of the authority and flag to turn address validation off.
+        /// Using this constructor, address validation can be turned off. Make sure you are aware of the security implication of not validating the address.
+        /// 
+        /// This constructor allows the HttpClient to be provided instead of having a new one created.  This can be useful when working with multiple different tenants.
```
henrik-me, line 115: >This constructor allows the HttpClient to be provided instead of having a new one created.  This can be useful when working with multiple different tenants. [](start = 12, length = 156)

This should likely be in the remarks section and also we should have an aka.ms link to the wiki explaining this in more detail. Basically this allows for full control over the httpClient passed in including httpHandlers, which can be used for more control, but also for additional error handling and additional telemetry/logging.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -107,14 +108,35 @@ public AuthenticationContext(string authority, bool validateAuthority, TokenCach
         {
         }
 
+        /// <summary>
+        /// Constructor to create the context with the address of the authority and flag to turn address validation off.
+        /// Using this constructor, address validation can be turned off. Make sure you are aware of the security implication of not validating the address.
```
henrik-me, line 113: JM, to help come up with the right text

henrik-me:
Thanks @AaronPatterson for bringing the suggestion forward. This is definitely something we would like to support, however we haven't been able to find a good way to add this as you can also see from @bgavrilMS 's comment

bgavrilMS:
We will accept this as discussed with @jmprieur . Some basic unit tests should be added.

henrik-me:
fyi : @jennyf19 @MarkZuber

bgavrilMS:
Closing this PR as I have made some changes but I can't update this PR because it is based on a fork. New PR is here: https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1482

PR #1473, opened 2019-01-03
set token cache from requestData object in AcquireTokenHandlerBase()

jennyf19:
resharper suggestions/spelling fixes


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -47,18 +47,19 @@ internal abstract class AcquireTokenHandlerBase
...
+        protected readonly BrokerHelper BrokerHelper = new BrokerHelper();
+        private AdalHttpClient _client = null;
+        private readonly TokenCache _tokenCache;
         internal readonly RequestContext RequestContext;
 
         protected AcquireTokenHandlerBase(RequestData requestData)
         {
-            this.Authenticator = requestData.Authenticator;
+            Authenticator = requestData.Authenticator;
+            _tokenCache = requestData.TokenCache;
```
bgavrilMS, line 60: This is the only material change. What does this fix? Is it the incorrect logging?

bgavrilMS (approved):

PR #1476, opened 2019-01-10
remove MSAL from authActivity name to avoid issues w/MSAL authActiivty

jennyf19:
Same issue as MSAL with iOS UniversalViews, but this time w/Android. Since the name was MSAL, I made the change in ADAL instead. Have verified it fixes the issue w/a new customer sample and am waiting for verification from them w/the release build I sent.

Here's the error message: 
image

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/SystemWebview/AuthenticationActivity.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/SystemWebview/AuthenticationActivity.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Identity.Core.UI.SystemWebview
 {
     /// <summary>
     /// </summary>
-    [Activity(Name = "microsoft.identity.client.AuthenticationActivity")]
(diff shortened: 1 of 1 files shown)


henrik-me (approved):
:shipit:

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/android/SystemWebview/AuthenticationActivity.cs
```diff
@@ -41,7 +41,7 @@ namespace Microsoft.Identity.Core.UI.SystemWebview
 {
     /// <summary>
     /// </summary>
-    [Activity(Name = "microsoft.identity.client.AuthenticationActivity")]
+    [Activity(Name = "microsoft.identitymodel.clients.AuthenticationActivity")]
```
henrik-me, line 44: >[Activity(Name = "microsoft.identitymodel.clients.AuthenticationActivity")] [](start = 4, length = 75)

we should have a test that captures stuff like this, both for ADAL and for MSAL

Issue #1489, opened 2019-01-24
Occasional NullReferenceException comming from ADAL Http calls


bgavrilMS:
**Which Version of ADAL are you using ?**
4.5

**Which platform has the issue?**
desktop

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Repro**
1. Perform a standard interactive call. 
2. Ensure the call fails due to a timeout (you can use Fiddler to simulate this)

**Expected behavior**
An AdalServiceException should be thrown

**Actual behavior**
NullReferenceException 

2019-01-24T07:15:54.5446958Z: 7e59e2c4-c704-471b-b685-847c560c0a23 - AdalLoggerBase.cs: Network timeout, Exception message: A task was canceled.
2019-01-24T07:15:55.5599983Z: 7e59e2c4-c704-471b-b685-847c560c0a23 - AdalLoggerBase.cs: Retrying one more time..
2019-01-24T07:16:25.5677407Z: 7e59e2c4-c704-471b-b685-847c560c0a23 - AdalLoggerBase.cs: Network timeout, Exception message: A task was canceled.
2019-01-24T07:16:25.5677407Z: 7e59e2c4-c704-471b-b685-847c560c0a23 - AdalLoggerBase.cs: Retry Failed, Exception message: A task was canceled.
2019-01-24T07:16:25.5677407Z: 7e59e2c4-c704-471b-b685-847c560c0a23 - AdalLoggerBase.cs: System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Http.AdalHttpClient.<GetResponseAsync>d__23`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Http.AdalHttpClient.<GetResponseAsync>d__23`1.MoveNext()
[...]

PR #1491, opened 2019-02-01
Throw a relevant timeout ADALServiceException instead of null ref

bgavrilMS:
Fix for  https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1489


jmprieur (commented):
Awesome!
Do we need to do the same for MSAL.NET ?

bgavrilMS:
@jmprieur - no on MSAL we use a different HTTP Stack and we seem to be handling this exception well https://github.com:443/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/Microsoft.Identity.Client/Http/HttpManager.cs#L175

PR #1512, opened 2019-03-05
SilentWebUI needs a longer timeout

stefangossner:
SilentWebUI needs a longer timeout as the overall navigation time can significantly exceed 2 seconds


jmprieur:
LGTM. @bgavrilMS

Issue #1533, opened 2019-03-13
AcquireTokenByDeviceCodeAsync needs to support CancellationToken


MarkZuber:
**Which Version of ADAL are you using ?**
ADAL 5.0.0-preview

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [x] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

**Repro**

AcquireTokenByDeviceCodeAsync will internally poll for the access token but doesn't support cancellation.  CancellationToken support is needed in this API to allow the developer to cancel the operation as needed.

jmprieur:
how did I miss that ? :)

MarkZuber:
https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1534

PR #1534, opened 2019-03-13
AcquireTokenByDeviceCodeAsync needs CancellationToken [1533]

MarkZuber:
https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1533

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
@@ -27,6 +27,7 @@
 
 using System;
 using System.Net.Http;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Identity.Core;
 using Microsoft.Identity.Core.Cache;
@@ -237,11 +238,28 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 #if iOS
         [Obsolete("This device profile API should only be used on text-only devices, and not on this target platform which is offering an interactive authentication experience. For details please see https://aka.ms/AdalNetConfFlows")]
 #endif
-        public async Task<AuthenticationResult> AcquireTokenByDeviceCodeAsync(DeviceCodeResult deviceCodeResult)
+        public Task<AuthenticationResult> AcquireTokenByDeviceCodeAsync(DeviceCodeResult deviceCodeResult)
+        {
+            return AcquireTokenByDeviceCodeAsync(deviceCodeResult, CancellationToken.None);
+        }
+
+        /// <summary>
+        /// Acquires security token from the authority using an device code previously received.
+        /// This method does not lookup token cache, but stores the result in it, so it can be looked up using other methods such as <see cref="AuthenticationContext.AcquireTokenSilentAsync(string, string, UserIdentifier)"/>.
+        /// </summary>
+        /// <param name="deviceCodeResult">The device code result received from calling AcquireDeviceCodeAsync.</param>
+        /// <param name="cancellationToken">A CancellationToken which can be triggered to cancel the operation in progress.</param>
+        /// <returns>It contains Access Token, its expiration time, user information.</returns>
+#if iOS
+        [Obsolete("This device profile API should only be used on text-only devices, and not on this target platform which is offering an interactive authentication experience. For details please see https://aka.ms/AdalNetConfFlows")]
+#endif
+        public async Task<AuthenticationResult> AcquireTokenByDeviceCodeAsync(
+            DeviceCodeResult deviceCodeResult,
+            CancellationToken cancellationToken)
         {
             if (deviceCodeResult == null)
             {
-                throw new ArgumentNullException("deviceCodeResult");
+                throw new ArgumentNullException(nameof(deviceCodeResult));
             }
 
             RequestData requestData = new RequestData
@@ -253,7 +271,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
                 ClientKey = new ClientKey(deviceCodeResult.ClientId)
             };
 
-            var handler = new AcquireTokenByDeviceCodeHandler(requestData, deviceCodeResult);
+            var handler = new AcquireTokenByDeviceCodeHandler(requestData, deviceCodeResult, cancellationToken);
             return await handler.RunAsync().ConfigureAwait(false);
         }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -237,11 +238,28 @@ public async Task<DeviceCodeResult> AcquireDeviceCodeAsync(string resource, stri
 #if iOS
         [Obsolete("This device profile API should only be used on text-only devices, and not on this target platform which is offering an interactive authentication experience. For details please see https://aka.ms/AdalNetConfFlows")]
 #endif
-        public async Task<AuthenticationResult> AcquireTokenByDeviceCodeAsync(DeviceCodeResult deviceCodeResult)
+        public Task<AuthenticationResult> AcquireTokenByDeviceCodeAsync(DeviceCodeResult deviceCodeResult)
```
henrik-me, line 241: async ? #ByDesign

henrik-me:
no tests to update? add? #Resolved

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
MarkZuber: This doesn't need to be async since we're just returning the task from the other method.  There's no need to do an await here since we're just calling another method that returns a task. #ByDesign

MarkZuber:
test added

---
In reply to: [472635492](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1534#issuecomment-472635492) [](ancestors = 472635492)

jmprieur (approved):
LGTM

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -237,11 +238,28 @@ public async Task<DeviceCodeResult> AcquireDeviceCodeAsync(string resource, stri
...
+
+        /// <summary>
+        /// Acquires security token from the authority using an device code previously received.
+        /// This method does not lookup token cache, but stores the result in it, so it can be looked up using other methods such as <see cref="AuthenticationContext.AcquireTokenSilentAsync(string, string, UserIdentifier)"/>.
+        /// </summary>
+        /// <param name="deviceCodeResult">The device code result received from calling AcquireDeviceCodeAsync.</param>
+        /// <param name="cancellationToken">A CancellationToken which can be triggered to cancel the operation in progress.</param>
+        /// <returns>It contains Access Token, its expiration time, user information.</returns>
+#if iOS
+        [Obsolete("This device profile API should only be used on text-only devices, and not on this target platform which is offering an interactive authentication experience. For details please see https://aka.ms/AdalNetConfFlows")]
```
bgavrilMS, line 254: Not all users will understand "this target platform". Better to replace it with iOS.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenByDeviceCodeHandler.cs
```diff
@@ -34,24 +35,31 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Flows
 {
     internal class AcquireTokenByDeviceCodeHandler : AcquireTokenHandlerBase
     {
-        private readonly DeviceCodeResult deviceCodeResult;
+        private readonly DeviceCodeResult _deviceCodeResult;
+        private readonly CancellationToken _cancellationToken;
```
bgavrilMS, line 39: Wouldn't it be better place in the base? Ah nevermind, this is ADAL :)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenByDeviceCodeHandler.cs
```diff
@@ -34,24 +35,31 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Flows
...
 
         protected internal /* internal for test only */ override async Task<AdalResultWrapper> SendTokenRequestAsync()
         {
-            TimeSpan timeRemaining = deviceCodeResult.ExpiresOn - DateTimeOffset.UtcNow;
+            TimeSpan timeRemaining = _deviceCodeResult.ExpiresOn - DateTimeOffset.UtcNow;
             AdalResultWrapper resultEx = null;
 
             while (timeRemaining.TotalSeconds > 0)
             {
+                if (_cancellationToken.IsCancellationRequested)
```
bgavrilMS: Use _cancellationToken.ThrowIfCancellationRequested() instead.

bgavrilMS (approved):

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
henrik-me: Will it break anyone by removing it?

henrik-me:
Seems like tests were failing as they couldn't get the credentials.

henrik-me (changes requested):
:clock1:

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
MarkZuber: No.  the async keyword isn't actually part of the public API per se, it's a compiler hint for how to thread the call.  The important part is that it still returns Task<AuthenticationResult>

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenByDeviceCodeHandler.cs
MarkZuber: Yup, ADAL. In MSAL, the cancellation token is part of the method call itself as it should be, but i don't want to re-wire all of RunAsync() across the actions for adal.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenByDeviceCodeHandler.cs
MarkZuber: done.

henrik-me (approved):
:shipit:

Issue #1554, opened 2019-03-27
Use HttpClient intialized with NSUrlSessionHandler() for iOS and Mac


jennyf19:
iOS Xamarin team recommends we use NSUrlSessionHandler() for the HttpClient.

See [documentation](https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/http-stack) for more details.

jennyf19:
released in [ADAL 5.0.1-preview](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/releases/tag/ADAL-v5.0.1-preview)

PR #1555, opened 2019-03-27
add NSURLSessionHandler as HttpClient for iOS and MAC

jennyf19:
[issue](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1554)
tested w/iOS device

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Core/Http/HttpClientFactory.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Core/Http/HttpClientFactory.cs
@@ -28,28 +28,34 @@
 using Microsoft.IdentityModel.Clients.ActiveDirectory;
 using System.Net.Http;
 using System.Net.Http.Headers;
+#if iOS || MAC
+using Foundation;
+#endif
 
 namespace Microsoft.Identity.Core.Http
-{ 
+{
     internal class HttpClientFactory : IHttpClientFactory
     {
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Core/Http/HttpClientFactory.cs
```diff
@@ -28,28 +28,38 @@
...
+        public const long MaxResponseContentBufferSizeInBytes = 1024 * 1024;
 
         public HttpClientFactory()
         {
-            if (_httpClient == null)
+#if iOS
+            if (UIDevice.CurrentDevice.CheckSystemVersion(7, 0))
             {
-                _httpClient = new HttpClient(new HttpClientHandler() { UseDefaultCredentials = true })
+                _httpClient = new HttpClient(new NSUrlSessionHandler())
```
bgavrilMS, line 51: Please add a comment with our learnings from Xamarin about this. You might also want to update the wiki.

bgavrilMS (approved):

MarkZuber (approved):
:shipit:

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Core/Http/HttpClientFactory.cs
jennyf19: done. to keep the comment less, i just added in the aka.ms link there.

PR #1556, opened 2019-03-28
Cache fallback to MSAL should respect uniqueID

bgavrilMS:


src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -670,7 +674,7 @@ internal async Task<List<string>> GetOrderedAliasesAsync(string authority, bool
                     CacheFallbackOperations.WriteMsalRefreshToken(TokenCacheAccessor, result, authority, clientId, displayableId,
                         result.Result.UserInfo.GivenName,
                         result.Result.UserInfo.FamilyName, 
-                        idToken?.ObjectId); // should this be UniqueId?!
+                        idToken?.GetUniqueId()); // should this be UniqueId?!
```
MarkZuber: you can probably cleanup the comment since this is now uniqueid

jennyf19 (approved):

PR #1560, opened 2019-03-28
update supported abis to be arm64-v8a

jennyf19:


bgavrilMS (approved):

devApps/XFormsApp.Droid/XFormsApp.Droid.csproj
```diff
@@ -37,7 +37,7 @@
     <EmbedAssembliesIntoApk>True</EmbedAssembliesIntoApk>
     <BundleAssemblies>False</BundleAssemblies>
     <AndroidCreatePackagePerAbi>False</AndroidCreatePackagePerAbi>
-    <AndroidSupportedAbis>armeabi-v7a,x86</AndroidSupportedAbis>
+    <AndroidSupportedAbis>arm64-v8a</AndroidSupportedAbis>
```
MarkZuber, line 40: removing x86 will remove the ability for this to be in the emulator.  may want to keep that one.

jennyf19: @MarkZuber I get this error when I add x86:
Android ABI mismatch. You are deploying an app supporting 'arm64-v8a,x86' ABIs to an incompatible device of ABI 'arm64-v8a;armeabi-v7a;armeabi'. You should either create an emulator matching one of your app's ABIs or add 'arm64-v8a' to the list of ABIs your app builds

MarkZuber (approved):

PR #1566, opened 2019-04-01
httpclient null fix

jennyf19:


src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs
```diff
@@ -123,7 +123,7 @@ public AuthenticationContext(string authority, bool validateAuthority, TokenCach
             : this(null, authority, validateAuthority ? AuthorityValidationType.True : AuthorityValidationType.False,
                    tokenCache, httpClientFactory)
         {
-            if (httpClientFactory == null)
+            if (_serviceBundle.HttpManager == null)
```
bgavrilMS: Why this check ? We want to ensure that the param that the user has passed in is valid, i.e not null and throw a meaningul meassage in case it is null.

bgavrilMS (approved):

Issue #1571, opened 2019-04-03
TokenCache before/after access notifications need to be within the cachelock boundary


MarkZuber:
**Which Version of ADAL are you using ?**
ADAL 5.0.1-preview

TokenCache access before/after notifications need to be within the _cacheLock lock in order for contract with the extensions file cache locking story to work.  Attached is screen shot of deadlock in VS.  Currently the before/after notifications are fired individually within the locks at the AcquireTokenHandlerBase level and encompass possibly more than one access to the cache.

![adalthreadlock](https://user-images.githubusercontent.com/7103178/55508905-b6a0cb00-560f-11e9-8542-e8a1b3586048.png)

PR #1572, opened 2019-04-03
Fix cache notification locking semantics [1571]

MarkZuber:
https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1571


jennyf19 (approved):
LGTM

bgavrilMS (approved):
I don't see the whole code, but as long we don't double notify (i.e. notify during a notification), LGTM

bgavrilMS (approved):

jmprieur:
which problem is it solving?

bgavrilMS:
@jmprieur The current implementation is causing deadlocks if the token cache file is syncronized from the outside. Check out the bug https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1571 to see a deadlock.

PR #1576, opened 2019-04-05
removing unnecessary exclusions from polycheck

trwalke:


trwalke:
previously approved changes
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1043

PR #1584, opened 2019-04-17
android httpclient fix

jennyf19:


bgavrilMS (approved):

bgavrilMS:
A unit test would be good here. You can use our MockHttpClient to respond with valid responses to all the discovery stuff, and then respond with an error response for the OAuthClient. We have a lot of tests like that.

jennyf19:
@bgavrilMS need to get a release out for a customer, so am going to merge and will take your suggestion later. thanks.

PR #1590, opened 2019-05-02
Automate ROPC in ADAL

bgavrilMS:


tests/Test.ADAL.Integration.net45/HeadlessTetss/UsernamePasswordTests.cs
```diff
@@ -0,0 +1,130 @@
+﻿using Microsoft.Identity.Test.LabInfrastructure;
```
jennyf19: Add header info

jennyf19 (approved):
:shipit:

Issue #1599, opened 2019-05-09
CreateFromResourceUrlAsync gone missing in ADAL 5.x preview 1


MattB-msft:
**Which Version of ADAL are you using ?**
ADAL5.0.1-preview => ADAL 5.0.5

**Which platform has the issue?**
,net462
 
**What authentication flow has the issue?**
* Desktop / Mobile
    * [x ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Repro**

```csharp
 AuthenticationParameters authParameters = AuthenticationParameters.CreateFromResourceUrlAsync(requestUri).Result;
```

**Expected behavior**
CreateFromResourceUrlAsync is no longer present as static.. though it appears in GitHub codebase still. 

**Actual behavior**
A clear and concise description of what happens, e.g. exception is thrown, UI freezes  

**Possible Solution**
Looks like it got lost with this PR: 
https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/commit/486c46c9741c6ad06c92bb2c8f50cec8ab18584a#diff-ac95f94d0b5f9e01efcf64255d0e85b1

Looks like this may have been a mistake? 
You cannot create an instance of AuthenticationParameters to get at it.

henrik-me:
Change was intentional, however for certain scenarios makes sense that this is available as a static as well. @bgavrilMS ?

henrik-me:
@MattB-msft : Can you pls. describe the scenario where you need this?

MattB-msft:
This is the common flow for our tools over the last several years to discover the authority and resource components for authentication.  We used this method so as to not have to implement our own web request / fault mechanism.

This has been there since at least the 2.x version.  Why remove it from the static interface and not supply a way to get at it.

MattB

________________________________

bgavrilMS:
Yes, I see the problem. We refactored ADAL to use an abstraction over the http layer and we made these methods non-static to be able to inject an HttpManager. Sadly, the constructor is internal, so there is no way to use the public method. 

Here's what I propose: 

- remove the non-static method. It won't break anyone because it can't be called, as the ctor for the class is internal
- bring back the static

bgavrilMS:
@MattB-msft - you can use the following workaround until a new release is ready: 

[...]

PR #1602, opened 2019-05-15
Fix AuthenticationParamters

bgavrilMS:
Fix for https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1599

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationParameters.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationParameters.cs
@@ -40,7 +40,7 @@ using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform;
 
 namespace Microsoft.IdentityModel.Clients.ActiveDirectory
 {
-    // This is a helper class and its functionality is tied to ADAL. It uses a vanilla HttpClient
+    // This is a helper class and its functionality is not tied to ADAL. It uses a vanilla HttpClient
 
     /// <summary>
     /// Contains authentication parameters based on unauthorized response from resource server.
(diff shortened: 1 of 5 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationParameters.cs
```diff
@@ -154,7 +159,8 @@ private async Task<AuthenticationParameters> CreateFromResourceUrlCommonAsync(Ur
 
             try
             {
-                await _httpManager.SendGetAsync(
+                HttpManager httpManager = new HttpManager();
```
MarkZuber: will this mean that we're getting a different HttpClient than the default one?  I need to dig into the httpmanager constructor to validate.  I"m mainly concerned if someone has decorated their httpclientfactory with proxy information then it might not get through here.

MarkZuber (changes requested):
Want to validate my comment.  If we're all good for httpclient stuff then i'm signed off.

henrik-me:
Can we introduce a factory method instead of restructuring?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationParameters.cs
bgavrilMS: This is fine, because this is a completely separate helper class, has nothing to do with ADAL. You're not even creating an AuthenticationContext, so a custom HttpClient can't even be used. 

If we weren't already in on a stable version, I would propose to remove this API completely.

bgavrilMS: To be honest, an HttpManager isn't even needed here (see how many nulls are used in initialization) -  I could create a simple HttpClient to make a GET call. The only minor advantage of HttpManager is a bit of extra logging.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationParameters.cs
```diff
@@ -53,36 +55,51 @@ public sealed class AuthenticationParameters
...
         /// <summary>
-        /// Creates authentication parameters from address of the resource. This method expects the resource server to return unauthorized response
-        /// with WWW-Authenticate header containing authentication parameters.
+        /// Sends a GET request to the url provided with no Authenticate header. If a 401 Unauthorized is received, this helper will parse the WWW-Authenticate header to 
+        /// retrieve the authority and resource.
         /// </summary>
         /// <param name="resourceUrl">Address of the resource</param>
         /// <returns>AuthenticationParameters object containing authentication parameters</returns>
-        public async Task<AuthenticationParameters> CreateFromResourceUrlAsync(Uri resourceUrl)
+        /// <remarks>Most protected APIs, including those owned by Microsoft, no longer advertise a resource. Authentication should be done using MSAL, which uses scopes. See https://aka.ms/msal-net-migration-adal-msal </remarks>
```
bgavrilMS, line 82: @henrik-me @jmprieur - how about I add these "remarks" to a nice Obsolete message instead? :D

henrik-me: Makes sense to me... except some might be using it who will not be able to if it's marked as obsolete

---
In reply to: [284784777](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1602#discussion_r284784777) [](ancestors = 284784777)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationParameters.cs
```diff
@@ -53,36 +55,51 @@ public sealed class AuthenticationParameters
...
+        public static async Task<AuthenticationParameters> CreateFromResourceUrlAsync(Uri resourceUrl)
         {
-            return await CreateFromResourceUrlCommonAsync(resourceUrl).ConfigureAwait(false);
+            if (resourceUrl == null)
+            {
+                throw new ArgumentNullException("resourceUrl");
+            }
+
+            HttpClient httpClient = new HttpClient();
```
henrik-me, line 90: >httpClient [](start = 23, length = 10)

not sure why we would not want to allow passing in a httpclient, allowing re-use and proxy, etc. ...

Issue #1604, opened 2019-05-22
Null Ref when IdToken is not returned


jennyf19:
**Which Version of ADAL are you using ?**
5.0.5 - v4.

**Repro**
```CSharp
AuthenticationResult res = await new AuthenticationContext(authority)
                                           .AcquireTokenByAuthorizationCodeAsync(arg.Code, new Uri(postLogoutRedirectUri), new ClientCredential(clientId, clientSecret));
           string token = res.AccessToken;
```
ADAL does not send the scope OpenId w/the AuthCode flow, so no idToken is returned. The idToken is used when writing to the MSAL cache. 

In `CacheFallbackOperations.WriteMsalRefreshToken`, we don't check for null in `result.Result.UserInfo.GivenName`, so there is a [null ref thrown here](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs#L693).

**Expected behavior**
1. OpenId scope should be sent w/each request, as we do in [MSAL](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs#L345)
2. Check for null in UserInfo and not throw, later the code will [return if no IdToken is present](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/034f0876ccf4f478ff4eefc462af07628c0497fb/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Core/Cache/CacheFallbackOperations.cs#L52) and not attempt to create an MSAL refresh token and account object.

**Actual behavior**
System.NullReferenceException: Object reference not set to an instance of an object. at Microsoft.IdentityModel.Clients.ActiveDirectory.TokenCache.StoreToCacheCommon(AdalResultWrapper result, String authority, String resource, String clientId, TokenSubjectType subjectType, RequestContext requestContext) in D:\a\1\s\src\Microsoft.IdentityModel.Clients.ActiveDirectory\TokenCache.cs:line 692

jmprieur:
@jennyf19 
In ASP.NET / ASP.NET core apps things are a bit different as ASP.NET request the scopes and starts the authorization flow, and then the code is redeemed by ADAL or MSAL when the code is received
[...]

PR #1605, opened 2019-05-22
Add openId scope to all requests via ATHandlerBase  [do not merge before discussing]

jennyf19:
Check for null values on UserInfo
Add a utest

[null ref issue](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1604)

Also did some manual testing w/all the flows on the desktop app.
If we don't want to take the change of adding openid scope to all requests, we can just fix the null ref instead.

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
@@ -326,7 +326,8 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Flows
         {
             var requestParameters = new DictionaryRequestParameters(Resource, ClientKey)
             {
-                { OAuthParameter.ClientInfo, "1" }
+                { OAuthParameter.ClientInfo, "1" },
+                { OAuthParameter.Scope, OAuthValue.ScopeOpenId }
             };
             AddAdditionalRequestParameters(requestParameters);
             return await SendHttpMessageAsync(requestParameters).ConfigureAwait(false);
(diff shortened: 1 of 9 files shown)


jmprieur (commented):
@jennyf19 . Which problem is it solving? 
Does it make sense for client credentials?

jmprieur:
Assuming here this is attempting to fix #1604 
See my comment in the issue. I think that the problem is with the Web app, not with ADAL.

henrik-me:
@jmprieur @jennyf19 : if I undertstand @jennyf19 correctly, we do this for MSAL already. Seems like something we can also for for ADAL? (even though I believe the right/better approach for web apps is what JM as suggested in the issue). Anyway we should definitely improve the error message and add information to the wiki/ref-doc on this.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Flows/AcquireTokenHandlerBase.cs
```diff
@@ -326,7 +326,8 @@ protected virtual async Task PostTokenRequestAsync(AdalResultWrapper resultEx)
         {
             var requestParameters = new DictionaryRequestParameters(Resource, ClientKey)
             {
-                { OAuthParameter.ClientInfo, "1" }
+                { OAuthParameter.ClientInfo, "1" },
+                { OAuthParameter.Scope, OAuthValue.ScopeOpenId }
```
henrik-me, line 330: you may want to add a comment on why this is added (e.g. needed due to cache requirements on the idt).

jmprieur: do we add openid for client creds? that would be odd ?

PR #1607, opened 2019-05-30
fix null ref in tokenCache when UserInfo is absent

jennyf19:
utest for IdToken and WriteMsalRefreshToken()

Alternative for [issue 1604](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1604)

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
@@ -690,8 +690,8 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
                         IdToken.TryParse(result.Result.IdToken, out IdToken idToken);
 
                         CacheFallbackOperations.WriteMsalRefreshToken(TokenCacheAccessor, result, authority, clientId, displayableId,
-                            result.Result.UserInfo.GivenName,
-                            result.Result.UserInfo.FamilyName,
+                            result.Result.UserInfo?.GivenName,
+                            result.Result.UserInfo?.FamilyName,
                             idToken?.GetUniqueId());
                     }
                 }
--- a/tests/Test.ADAL.NET.Unit.net45/TokenCacheUnitTests.cs
+++ b/tests/Test.ADAL.NET.Unit.net45/TokenCacheUnitTests.cs
@@ -36,6 +36,8 @@ using Microsoft.Identity.Test.Common.Core.Mocks;
 using Microsoft.Identity.Core;
 using Microsoft.Identity.Core.Cache;
 using Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Cache;
+using Test.ADAL.NET.Common.Mocks;
+using System;
 
 namespace Test.ADAL.NET.Unit
 {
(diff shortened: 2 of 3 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
```diff
@@ -690,8 +690,8 @@ internal async Task<List<string>> GetOrderedAliasesAsync(string authority, bool
                         IdToken.TryParse(result.Result.IdToken, out IdToken idToken);
 
                         CacheFallbackOperations.WriteMsalRefreshToken(TokenCacheAccessor, result, authority, clientId, displayableId,
-                            result.Result.UserInfo.GivenName,
-                            result.Result.UserInfo.FamilyName,
+                            result.Result.UserInfo?.GivenName,
+                            result.Result.UserInfo?.FamilyName,
```
jennyf19, line 694: adding ? is the actual fix. the customer was getting a null ref because UserInfo is null here. There will no longer be a null ref here. 
image
 #Closed

henrik-me: Is it then possible to complete the scenario? I believe the MSAL cache will now be empty and thus only when using the ADAL cache will there be any values in the cache... thus depending on how the customer persists the cache (which serializer/deserializer) this may or may not work? #Closed

tests/Test.ADAL.NET.Unit.net45/TokenCacheUnitTests.cs
```diff
@@ -186,11 +188,72 @@ public void ParallelStoreTest()
...
+
+        [TestMethod]
+        [Description("Test to ensure WriteMsalRefreshToken handles presence of IdToken correctly")]
+        [TestCategory("AdalDotNet.Unit")]
+        public void IdTokenReturnedCacheTest()
+        {
+            // arrange
+            TokenCache tokenCache = new TokenCache();
+
+            var result = CreateAdalResultWrapper();
```
henrik-me: nit: wrapper (or similar) #Closed

tests/Test.ADAL.NET.Unit.net45/TokenCacheUnitTests.cs
```diff
@@ -186,11 +188,72 @@ public void ParallelStoreTest()
...
+            return new AdalResultWrapper
+            {
+                RefreshToken = "some-rt",
+                RawClientInfo = MockHelpers.CreateClientInfo(AdalTestConstants.DefaultUniqueId, AdalTestConstants.DefaultUniqueTenantIdentifier),
+                ResourceInResponse = AdalTestConstants.DefaultResource,
+                Result = new AdalResult("Bearer", "some-access-token", DateTimeOffset.UtcNow, (DateTimeOffset.UtcNow + TimeSpan.FromMinutes(180)))
+            };
+        }
+
+        private void WriteMsalRefreshToken(AdalResultWrapper result, ITokenCacheAccessor tokenCacheAccessor)
```
henrik-me: suggestion: WriteMsalRefreshTokenWithNullAsGivenNameAndFamilyName / WriteMsalRefreshTokenWithNullUserInfoValues #Closed

tests/Test.ADAL.NET.Unit.net45/TokenCacheUnitTests.cs
```diff
@@ -186,11 +188,72 @@ public void ParallelStoreTest()
...
+
+            result.Result.IdToken = "some-id-token";
+
+            // act
+            WriteMsalRefreshToken(result, tokenCache.TokenCacheAccessor);
+
+            // assert
+            // IdToken present
+            Assert.AreEqual(tokenCache.TokenCacheAccessor.AccountCount, 1);
+            Assert.AreEqual(tokenCache.TokenCacheAccessor.RefreshTokenCount, 1);
```
henrik-me, line 233: >Assert.AreEqual [](start = 12, length = 15)

can it be tested that the userinfo values are null? #Closed

tests/Test.ADAL.NET.Unit.net45/TokenCacheUnitTests.cs
```diff
@@ -186,11 +188,72 @@ public void ParallelStoreTest()
...
+        [Description("Test to ensure WriteMsalRefreshToken handles presence of IdToken correctly")]
+        [TestCategory("AdalDotNet.Unit")]
+        public void IdTokenReturnedCacheTest()
+        {
+            // arrange
+            TokenCache tokenCache = new TokenCache();
+
+            var result = CreateAdalResultWrapper();
+
+            result.Result.IdToken = "some-id-token";
```
henrik-me: >result.Result.IdToken = "some-id-token"; [](start = 12, length = 40)

perhaps something to be passed into the wrapper? Or another method which does this? seems like magic what you are doing here, this can be better explained with a method names elegantly imo #Resolved

henrik-me:
I believe the added null checks should go in, have a few questions around the change more than actual comments relevance of the fix.  #Resolved

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
jennyf19: you are right. if there is no id_token, then we will exit out of writing items to the MSAL cache...[see here](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Core/Cache/CacheFallbackOperations.cs#L52) however, i'm assuming, since this has been out for awhile, that it hasn't been very impactful, or our customers are not using the cache in that way?

---
In reply to: [288842460](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1607#discussion_r288842460) [](ancestors = 288842460)

jennyf19:
sure. we can sync up tomorrow.

---
In reply to: [497186624](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1607#issuecomment-497186624) [](ancestors = 497186624)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/TokenCache.cs
jennyf19: addressed the comments. lmk if you'd like to discuss anything further. thanks for the utest suggestions and improvements.

---
In reply to: [288855740](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1607#discussion_r288855740) [](ancestors = 288855740,288842460)

henrik-me (approved):
:shipit:

PR #1614, opened 2019-06-07
Jennyf/ios broker fixes

### LINKED ISSUE #1606: iOS Broker Null Ref in TokenResponse (Key missing)
5.0.5

Xamarin iOS + iOS Broker v2

**Repro**
- Register an app in the portal
- Register this redirectUri: `adaliosapp://com.yourcompany.xformsapp`
- Do not include MS Graph as required permissions
- Include the client_id from the app in the Xamarin project (in AppConstants.cs)
- Rebuild

- When the app loads, select the "broker" option under client_id
- Hit "Acquire Token with Broker"

**Expected behavior**
ADAL will return a nice error message that the app is misconfigured based on the broker response (does not have permission to MS Graph)

**Actual behavior**
[A null ref is thrown in the token response because the broker response dictionary does not contain the key "error" in this case](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/OAuth2/TokenResponse.cs#L126). 

**Possible Solution**
If the key "error" is not present, use "error_domain" instead, which is a required value coming back from iOS broker. Only iOS broker is using this code, so Android broker response is not valid in this scenario. 

I don't know if we ever tested this particular scenario, and is not related to iOS broker v2.

jennyf19:
add logging
fix teamId default
fix keychain w/legacy cache persistence
fix [issue with null ref](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1606)

Tested w/iOS device w/broker (v2) and w/ interactive/silent flows and w/ and w/out the SetIosKeychainProperty

build passed [here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=352436)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/OAuth2/TokenResponse.cs
```diff
@@ -119,33 +116,47 @@ internal class TokenResponse
...
                     ErrorDescription = responseDictionary[TokenResponseClaim.ErrorDescription]
                 };
+                if (responseDictionary.ContainsKey(TokenResponseClaim.Error))
+                {
+                    tokenResponse.Error = responseDictionary[TokenResponseClaim.Error];
+                }
+                else
+                {
+                    // error_domain is a required field in a failed iOS broker response
+                    tokenResponse.Error = responseDictionary[TokenResponseClaim.ErrorDomain];
```
bgavrilMS, line 134: Does MSAL have this logic? #Resolved

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/iOSTokenCacheAccessor.cs
```diff
@@ -36,9 +36,24 @@
 
 namespace Microsoft.Identity.Core
 {
-    internal class iOSTokenCacheAccessor : ITokenCacheAccessor
+    internal class iOSTokenCacheAccessor : ITokenCacheAccessor, ILegacyCachePersistence
     {
         public const string CacheKeyDelimiter = "-";
+        private const bool _defaultSyncSetting = false;
+        private const SecAccessible _defaultAccessiblityPolicy = SecAccessible.AfterFirstUnlockThisDeviceOnly;
```
bgavrilMS: const have different style, like the one at line 45 #Resolved

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/iOSTokenCacheAccessor.cs
```diff
@@ -36,9 +36,24 @@
 
 namespace Microsoft.Identity.Core
 {
-    internal class iOSTokenCacheAccessor : ITokenCacheAccessor
+    internal class iOSTokenCacheAccessor : ITokenCacheAccessor, ILegacyCachePersistence
```
bgavrilMS: This is for backwards compat with ADAL v3. I would rather not do this unless there are customers who cannot use MSAL directly. ADAL v3 -> ADAL v5 / MSAL never worked as far as I can tell.  #Resolved

bgavrilMS:
I think the extra logging and the error description changes are fine and should be commited (please do). I would not make changes the to the iOS accessor. #Resolved

jennyf19:
yeah...sure...sounds good. 

---
In reply to: [500560924](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1614#issuecomment-500560924) [](ancestors = 500560924)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/OAuth2/TokenResponse.cs
jennyf19: not sure...will check.

---
In reply to: [292153534](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1614#discussion_r292153534) [](ancestors = 292153534)

bgavrilMS (approved):

PR #1617, opened 2019-06-12
Force usage of System.Net.Http 4.3.4 as 4.3.0 has security flaws

bgavrilMS:
Problem: System.Net.Http 4.3.0 has security issues. ADAL and some test projects use it indirectly (deep in the dependency tree). Cannot update high level components (NSubstitute, Azure.KeyVault.dll because even the latest version still use System.Net.Http 4.3.0), so I chose to force 4.3.3 by adding a direct reference to it. 

ADAL impact: ADAL will use this only for .NET Standard. 
Tests done: smoke testing with a .net core app (as it will exercise the .net standard code)

Passing build: https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=354511&_a=summary

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Microsoft.IdentityModel.Clients.ActiveDirectory.csproj
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Microsoft.IdentityModel.Clients.ActiveDirectory.csproj
@@ -91,13 +91,14 @@
     <Reference Include="System.Core" />
     <Reference Include="System.Drawing" />
     <Reference Include="System.IdentityModel" />
-    <Reference Include="System.Net.Http" />
     <Reference Include="System.Windows.Forms" />
     <Reference Include="System.Xml.Linq" />
     <Reference Include="System.Data.DataSetExtensions" />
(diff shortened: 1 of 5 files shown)


jennyf19 (approved):

jnyrup:
In version 5.1.0 the net45 target now depends on the nuget package of `System.Net.Http`, whereas 5.0.5 does not and would get `System.Net.Http.dll` from the framework assemblies.

Is that intended?
I see that [Microsoft.Azure.KeyVault 3.0.4](https://www.nuget.org/packages/Microsoft.Azure.KeyVault) only forces `System.Net.Http` for .netstandard targets.

bgavrilMS:
Yes, there seems to be security flaw in the 4.3.0 version and this is a way to ensure 4.3.0 is not used. Are you getting conflicts?

jnyrup:
I haven't tried it out, to see if I get conflicts with this particular version of the nuget `System.Net.Http`.
I've had problem with referencing `System.Net.Http` from full framework projects, that's why I'm asking.
Also because `Microsoft.Azure.KeyVault` does not use the nuget version for their full framework targets.

PR #1628, opened 2019-07-11
allow scrolling on the result in dev app

jennyf19:
fix scroll view on the result
few other small app fixes


jennyf19:
[pr test passed here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=368359)

Issue #1627, opened 2019-07-09
5.1.0 Authority forced to port 443


keystroke:
**MSAL is the recommended auth library for use with the Microsoft identity platform**

No new features will be implemented on ADAL. The team's efforts are on improving [MSAL](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet), the next-gen auth library. MSAL's wiki contains a [migration guide](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Adal-to-Msal) from ADAL.

Only regressions, high severity issues and security issues will be fixed on ADAL. Other issues are likely to have already been fixed in MSAL. 

If you think that your issue falls into the above categories, please fill in the form below.

**Which Version of ADAL are you using ?**
Note that to get help, you need to run the latest preview or non-preview version
For MSAL, please log issues to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet
Latest Version (5.1.0)

**Which platform has the issue?**
.NET 4.6.2

**What authentication flow has the issue?**
* Desktop / Mobile
    * [X] Interactive
    * [ ] Integrated Windows Auth
    * [X] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Is this a new or existing app?**
Existing test code.

**Repro**

```csharp
var context = new AuthenticationContext(
                authority: "https://localhost:4433/common",
                validateAuthority: false,
                tokenCache: new TokenCache());

// call acquire token and look at network calls, authority is set to port 443 no matter what input is given.
```

**Expected behavior**
Calls should be made on specified port.

**Actual behavior**
Calls are always made on port 443.

**Possible Solution**
Don't reset to port 443.

**Additional context/ Logs / Screenshots**
Unnecessary.

jmprieur:
@keystroke 
I'm completely confused by what you've trying to do.
The authority is something like https://login.microsoftonline.com, See https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-client-application-configuration#authority
Are you writing/testing your own Security Token Service?
what is your scenario?

keystroke:
[...]

PR #1632, opened 2019-07-25
Enabling custom ports to be used in authorities.

trwalke:
https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1627


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
```diff
@@ -122,10 +125,35 @@ public async Task UpdateFromTemplateAsync(RequestContext requestContext)
                 {
                     ServiceBundle.InstanceDiscovery.AddMetadataEntry(host);
                 }
-                AuthorizationUri = ServiceBundle.InstanceDiscovery.FormatAuthorizeEndpoint(host, tenant);
-                DeviceCodeUri = string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/oauth2/devicecode", host, tenant);
-                TokenUri = string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/oauth2/token", host, tenant);
-                UserRealmUriPrefix = EnsureUrlEndsWithForwardSlash(string.Format(CultureInfo.InvariantCulture, "https://{0}/common/userrealm", host));
+
+                //Check if custom port is set on authority
+                if (authorityUri.Port != 443)
```
keystroke: Did you mean to reference "DefaultHttpsPort" const declared above?

trwalke: Yes. I forgot to push my last commit. Thanks

tests/Test.ADAL.NET.Unit.net45/UnitTests.cs
```diff
@@ -117,6 +119,30 @@ public void AuthorityTypeDetectionTest()
...
+                httpManager.AddInstanceDiscoveryMockHandler();
+                var serviceBundle = ServiceBundle.CreateWithCustomHttpManager(httpManager);
+
+                var context = new AuthenticationContext(
+                              serviceBundle,
+                              CustomPortAuthority,
+                              AuthorityValidationType.False,
+                              new TokenCache());
+
+                var result = context.AcquireTokenAsync(AdalTestConstants.DefaultRedirectUri.ToString(),
```
bgavrilMS, line 142: We run all our tests on .NetFx and .NetCore, however you'll notice that on .net core PlatformParamters does not exist. On ADAL, you can't do acquire token interactive from .net core. So you need to not run this test on .net core.

tests/Test.ADAL.NET.Unit.net45/UnitTests.cs
```diff
@@ -117,6 +119,30 @@ public void AuthorityTypeDetectionTest()
...
+            {
+                var CustomPortAuthority = "https://localhost:5215/common/";
+                httpManager.AddInstanceDiscoveryMockHandler();
+                var serviceBundle = ServiceBundle.CreateWithCustomHttpManager(httpManager);
+
+                var context = new AuthenticationContext(
+                              serviceBundle,
+                              CustomPortAuthority,
+                              AuthorityValidationType.False,
+                              new TokenCache());
```
bgavrilMS, line 137: You should asssert the authority here as well: 

` Assert.AreEqual(CustomPortAuthority, context.Authority);`

tests/Test.ADAL.NET.Unit.net45/UnitTests.cs
```diff
@@ -117,6 +119,30 @@ public void AuthorityTypeDetectionTest()
             Assert.AreEqual(AuthorityType.ADFS, Authenticator.DetectAuthorityType("https://abc.com/adfs/dummy/"));
         }
 
+        [TestMethod]
+        public void AuthorityCustomPortTest()
+        {
+            using (var httpManager = new MockHttpManager())
+            {
+                var CustomPortAuthority = "https://localhost:5215/common/";
```
bgavrilMS: naming - customPortAuthority

tests/Test.ADAL.NET.Unit.net45/UnitTests.cs
```diff
@@ -117,6 +119,30 @@ public void AuthorityTypeDetectionTest()
             Assert.AreEqual(AuthorityType.ADFS, Authenticator.DetectAuthorityType("https://abc.com/adfs/dummy/"));
         }
 
+        [TestMethod]
```
bgavrilMS, line 124: I like to add a comment like "Test for bug XXX and the github issue url". Bugs have a lot of context..

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
```diff
@@ -122,10 +125,35 @@ public async Task UpdateFromTemplateAsync(RequestContext requestContext)
                 {
                     ServiceBundle.InstanceDiscovery.AddMetadataEntry(host);
                 }
-                AuthorizationUri = ServiceBundle.InstanceDiscovery.FormatAuthorizeEndpoint(host, tenant);
-                DeviceCodeUri = string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/oauth2/devicecode", host, tenant);
-                TokenUri = string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/oauth2/token", host, tenant);
-                UserRealmUriPrefix = EnsureUrlEndsWithForwardSlash(string.Format(CultureInfo.InvariantCulture, "https://{0}/common/userrealm", host));
+
+                //Check if custom port is set on authority
+                if (authorityUri.Port != DefaultHttpsPort)
```
bgavrilMS: So the logic here is good but it is too repetitive, i.e. you write the same code 4 times. How about creating a static helper in `CoreHelpers` or `UriBuilderExtensions`

`UriBuilder GetHttpsUriWithOptionalPort(string host, string tenant, string path, int port)`

The advantages of having this are: 

- very easy to test (your existing test is good, but you can write more tests this way)
- very easy to move to MSAL, along with tests

henrik-me (changes requested):
:clock1:

devApps/AdalUAPTestApp/AdalUAPTestApp.csproj
```diff
@@ -17,7 +17,8 @@
     <FileAlignment>512</FileAlignment>
     <ProjectTypeGuids>{A5A43C5B-DE2A-4C0C-9213-0A381AF9435A};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
     <WindowsXamlEnableOverview>true</WindowsXamlEnableOverview>
-    <PackageCertificateKeyFile>UAPTestApp_TemporaryKey.pfx</PackageCertificateKeyFile>
+    <PackageCertificateKeyFile>AdalUAPTestApp_TemporaryKey.pfx</PackageCertificateKeyFile>
```
trwalke, line 20: Our certificate expired and it stopped the ADAL builds from working so I had to recreate it. This should be ok right @bgavrilMS?

LibsAndSamples.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28606.126
+# Visual Studio 15
```
bgavrilMS: Please don't check these changes in.

devApps/AdalUAPTestApp/AdalUAPTestApp.csproj
bgavrilMS: Yes.

bgavrilMS (approved):

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Core/Helpers/UriBuilderExtensions.cs
```diff
@@ -47,5 +50,32 @@ public static void AppendQueryParameters(this UriBuilder builder, string queryPa
...
+        public static string GetHttpsUriWithOptionalPort(string host, string tenant, string path, int port)
+        {
+            var builder = new UriBuilder("https", host);
+            builder.Path = string.Format(CultureInfo.InvariantCulture, "{0}/{1}", tenant, path);
+
+            //No need to set port if it equals 443 as it is the default https port
+            if (port != DefaultHttpsPort)
+            {
+                builder.Port = port;
+            }
```
henrik-me, line 63: I wonder why you pass in host to the UriBuilder as you are overriding the path... thoughts?

henrik-me, line 63: Seems like this could just call into the method below?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
```diff
@@ -107,6 +108,7 @@ public async Task UpdateFromTemplateAsync(RequestContext requestContext)
             {
                 var authorityUri = new Uri(Authority);
                 var host = authorityUri.Host;
+                var authority = authorityUri.Authority;
```
henrik-me, line 111: Seems like authority is not used any more?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Core/Helpers/UriBuilderExtensions.cs
trwalke: when ADAL uses the authority, it strips out the host and appends various paths to it to create the various endpoints. I was simply following that logic. Other than that there is no real concrete reason.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
trwalke: Yes. will remove

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Core/Helpers/UriBuilderExtensions.cs
trwalke: yes, I can modify it to do that

PR #1636, opened 2019-08-05
Set exception on TCS so it is thrown in correct place

crmann1:


jennyf19:
@crmann1 This is the same fix you did in MSAL last week right?

jennyf19 (approved):

crmann1:
Yes it is

Issue #1478, opened 2019-01-10
Adal 4.4 can't acquire token using Integrated authentication on Windows


myurashchyk:
**Which Version of ADAL are you using ?**
ADAL 3.19.8, ADAL 4.4.2

**Which platform has the issue?**
net45

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [x] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Repro**
We have password hash synchronization with Azure AD configured and want acquire token for current user silently. With ADAL 3.19.8 we are getting token as expected, but ADAL 4.4.2 throws exception on the same code.
```csharp
            try
            {
                var context = new AuthenticationContext("https://login.microsoftonline.com/{aad}");
                var result = context.AcquireTokenAsync(resource, client, new UserCredential()).Result;
                Console.WriteLine(result.AccessToken);
            }
            catch (Exception ex)
            {
                Console.WriteLine(ex);
            }
            Console.ReadKey();
```

**Expected behavior**
Token should be acquired without error.

**Actual behavior**
Microsoft.IdentityModel.Clients.ActiveDirectory.AdalException: Integrated Windows Auth is not supported for managed users. See https://aka.ms/adal-iwa for details.

**Possible Solution**

**Additional context/ Logs / Screenshots**
[Information][False]: 2019-01-10T08:43:40.7200168Z: af17f266-f509-4360-8395-8a38
b0a044a1 - AdalLoggerBase.cs: ADAL PCL.Desktop with assembly version '4.4.2.0',
file version '4.4.2.0' and informational version '4.4.2' is running...
[Information][True]: 2019-01-10T08:43:40.7200168Z: af17f266-f509-4360-8395-8a38b
0a044a1 - AdalLoggerBase.cs: === Token Acquisition started:
        Authority: *
        Resource: *
        ClientId: *
        CacheType: null
        Authentication Target: User

[Information][True]: 2019-01-10T08:43:41.0012785Z: af17f266-f509-4360-8395-8a38b
0a044a1 - AdalLoggerBase.cs: Logged in user detected with user name 'User*@domain.com'
[Verbose][False]: 2019-01-10T08:43:41.0012785Z: af17f266-f509-4360-8395-8a38b0a0
[...]

PR #1640, opened 2019-08-21
Enabling IWA in Seamless SSO for hybrid scenarios

trwalke:
Adding product info headers to user realm discovery request. This is required according to the seamless sso/ws trust spec:

```
This is the list of headers whose existence (At least one) will trigger WS Trust with Seamless SSO: 
-       x-client-Ver Valentin: should work for c++. We always send this. 
                   ADAL sends this header to indicate its client version. 
-       x-client-SKU 
                   ADAL sends this header to indicate its SKU (i.e. the platform it's running on). 
-       x-client-OS 
                   ADAL sends this header to indicate os it's running on. 
```
Fix for bug here https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1478
This was tested manually in our lab test environment


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Core/WsTrust/WsTrustWebRequestManager.cs
```diff
@@ -141,7 +142,7 @@ public async Task<MexDocument> GetMexDocumentAsync(string federationMetadataUrl,
 
             var uri = new UriBuilder(userRealmUriPrefix + userName + "?api-version=1.0").Uri;
 
-            var httpResponse = await _httpManager.SendGetAsync(uri, null, requestContext).ConfigureAwait(false);
+            var httpResponse = await _httpManager.SendGetAsync(uri, (Dictionary<string, string>)AdalIdHelper.GetAdalIdParameters(), requestContext).ConfigureAwait(false);
```
bgavrilMS: Can you modify `SendGetAsync` to accept an `IDictionary<string, string>` instead of `Dictonary<string, string>` ? It is a trivial change. 

In general methods should take in interfaces to collections and not implementations. In this case, would it matter if I'd passed in a SortedDictionary<String, string> instead of a Dictionary<string, string> ? No - the HttpManager just iterates through it to get the keys and the values out.

bgavrilMS (approved):
See comment inline, but otherwise looks good.

PR #1679, opened 2019-11-27
fix null ref in non broker calls when source app is null

jennyf19:
now w/ios 13, sourceApplication can be null, and if null, we are throwing a null ref later in the code for non broker responses, as the SemaphoreSlim was never created during the creation of the ios broker object. have tested w/customer repro and our own broker app. see issue for more details.

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationContinuationHelper.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationContinuationHelper.cs
@@ -72,13 +72,16 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory
         /// <param name="url"></param>
         public static void SetBrokerContinuationEventArgs(NSUrl url)
         {
-            if(url == null)
+            // url should always be present
+            // if coming from broker, will be the broker response
+            // check to prevent null ref later
+            if (url == null) 
             {
                 return;
             }
 
             string urlString = string.Format(CultureInfo.CurrentCulture, url.ToString());
-            if(urlString.Contains(BrokerConstants.Broker))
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/AuthenticationContinuationHelper.cs
```diff
@@ -71,7 +72,16 @@ public static bool IsBrokerResponse(string sourceApplication)
         /// <param name="url"></param>
         public static void SetBrokerContinuationEventArgs(NSUrl url)
         {
-            iOSBroker.SetBrokerResponse(url);
+            if(url == null)
```
bgavrilMS: When can this happen? Please add a comment or a remark  #Resolved

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Platforms/iOS/BrokerConstants.cs
```diff
@@ -51,5 +51,7 @@ internal static class BrokerConstants
         public const string ApplicationToken = "application_token"; // sent in request and response with iOS broker v3
         public const string AttemptToSaveBrokerApplicationToken = "Attempt to save iOS broker application token resulted in: ";
         public const string SecStatusCodeFromTryGetBrokerApplicationToken = "The SecStatusCode from trying to get the broker application token is: ";
+
+        public const string Broker = "broker";
```
bgavrilMS: This constant needs a better name :) #Resolved

bgavrilMS (approved):
See minor comments

jennyf19:
[PR build passed here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=437548)
[UI build passed here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=437558)

PR #1682, opened 2019-12-10
throw instance discovery timeout exception

jennyf19:


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
```diff
@@ -187,6 +187,11 @@ private async Task DiscoverAsync(Uri authority, bool validateAuthority, RequestC
                 // All other unexpected exceptions will still bubble up, as always.
                 if (validateAuthority)
                 {
+                    if (ex.ErrorCode == ErrorCodes.UnknownError && ex.StatusCode == 502)
```
bgavrilMS: This is a bit tricky because HTTP error 502 is not the only one that represents a timeout. There are several others that represent the same thing. 

In my mind, we should just re-throw all exceptions except of when ex.ErrorCode == "invalid_instance". Neither ADAL nor MSAL will work without instance discovery anyway, so we are not introducing a breaking change.

So I would do smth like 

if error is "invalid_instance" throw AdalError.AuthorityNotInValidList
else rethrow #Resolved

bgavrilMS (changes requested):
Let's consider re-throwing all but "invalid_authority" exception

jennyf19:
Let's consider re-throwing all but "invalid_authority" exception

---
In reply to: [330036997](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1682#issuecomment-330036997) [](ancestors = 330036997)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
```diff
@@ -187,11 +187,15 @@ private async Task DiscoverAsync(Uri authority, bool validateAuthority, RequestC
                 // All other unexpected exceptions will still bubble up, as always.
                 if (validateAuthority)
                 {
-                    // hard stop here
-                    throw new AdalException(
-                        (ex.ErrorCode == "invalid_instance")
-                            ? AdalError.AuthorityNotInValidList
-                            : AdalError.AuthorityValidationFailed, ex);
+                    if (ex.ErrorCode == ErrorCodes.InvalidInstance)
```
bgavrilMS: This works, but please use `string.Equals` with an ignore case culture. #Resolved

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
```diff
@@ -187,11 +187,15 @@ private async Task DiscoverAsync(Uri authority, bool validateAuthority, RequestC
...
+                    {
+                        throw new AdalException(
+                            (ex.ErrorCode == "invalid_instance")
```
bgavrilMS: Well, the ErrorCode is "invalid_instance" because you just checked :) #Resolved

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
```diff
@@ -187,11 +187,15 @@ private async Task DiscoverAsync(Uri authority, bool validateAuthority, RequestC
...
+                                ? AdalError.AuthorityNotInValidList
+                                : AdalError.AuthorityValidationFailed, ex);
+                    }
+
+                    throw new AdalException(ex.Response.StatusCode.ToString(), ex.InnerException.Message);
```
bgavrilMS: You are already in a `catch` block where you caught the relevant exception. So instead of creating a new exception, just let the old one propagate further. There are 2 ways of doing it: 

`throw;` // preserves the original stack trace
`throw ex;` // resets the stack trace

It's not very important, but you should preserve the original stack trace. 

 #Resolved

bgavrilMS (changes requested):
No need to create a new exception, just rethrow the old one.

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
jennyf19: sounds good. thanks

---
In reply to: [356229220](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1682#discussion_r356229220) [](ancestors = 356229220)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
```diff
@@ -187,11 +187,15 @@ private async Task DiscoverAsync(Uri authority, bool validateAuthority, RequestC
                 // All other unexpected exceptions will still bubble up, as always.
                 if (validateAuthority)
                 {
-                    // hard stop here
-                    throw new AdalException(
-                        (ex.ErrorCode == "invalid_instance")
-                            ? AdalError.AuthorityNotInValidList
-                            : AdalError.AuthorityValidationFailed, ex);
+                    if (Equals(ex.ErrorCode == ErrorCodes.InvalidInstance, StringComparison.OrdinalIgnoreCase))
```
bgavrilMS: Please use `string.Equals(ex.ErrorCode, ErrorCodes.InvalidInstance, StringComparison.OrdinalIngnoreCase)`  #Resolved

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
```diff
@@ -187,11 +187,15 @@ private async Task DiscoverAsync(Uri authority, bool validateAuthority, RequestC
...
+                    {
+                        throw new AdalException(
+                            AdalError.AuthorityNotInValidList, 
+                            AdalError.AuthorityValidationFailed,
```
bgavrilMS: Is this (AuthorityValidationFailed) an actual error message?

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
```diff
@@ -187,11 +187,15 @@ private async Task DiscoverAsync(Uri authority, bool validateAuthority, RequestC
...
-                            : AdalError.AuthorityValidationFailed, ex);
+                    if (Equals(ex.ErrorCode == ErrorCodes.InvalidInstance, StringComparison.OrdinalIgnoreCase))
+                    {
+                        throw new AdalException(
+                            AdalError.AuthorityNotInValidList, 
+                            AdalError.AuthorityValidationFailed, 
+                            ex);
+                    }
+                    throw;
+                    //throw new AdalException(ex.Response.StatusCode.ToString(), ex.InnerException.Message);
```
bgavrilMS: remove dead code. #Resolved

bgavrilMS (approved):
Please see comments

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
jennyf19: we turn the invalid_instance into authority validation failed...is that what you mean?

---
In reply to: [356264823](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1682#discussion_r356264823) [](ancestors = 356264823)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/InstanceDiscovery.cs
jennyf19: oops. yes. sorry about that.

---
In reply to: [356264930](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1682#discussion_r356264930) [](ancestors = 356264930)

jennyf19:
i think i've got everything now.

---
In reply to: [330128996](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1682#issuecomment-330128996) [](ancestors = 330128996)

PR #1683, opened 2019-12-10
fix lab resource url

jennyf19:


jmprieur (approved):
LGTM.
You might want to add a comment, in the PR, to explain why you removed the InstanceCache.Clear() instruction

jennyf19:
i removed it because in the test initializer, we clear the instance cache, but here, we're clearing it again and in azure dev ops, we were getting an issue that the mocks were not correct, probably because it was clearing the instance cache twice...removing it has fixed the issue, unless the tests ran in a different order and the issue will surface again at another time.

---
In reply to: [330318470](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1683#issuecomment-330318470) [](ancestors = 330318470)

Issue #1687, opened 2020-01-10
Android Broker fails to refresh silently due to incorrect UID


trwalke:
**Which Version of ADAL are you using ?**
latest (5.2.5)

**Which platform has the issue?**
xamarin android

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [x] Silent
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

**Repro**
Make an interactive call using broker on android for one resource and then try to acquire a token silently for a different resource. 

**Expected behavior**
Broker should be able to successfully refresh the token
**Actual behavior**
The app is forced to sign in interactively again and broker throws the following error
"Could not locate a cache for this app"

**Possible Solution**
We need to send the UID instead of the PID that we send today because the UID will not change for the user on the device. The broker uses the UID to cache responses so if the UID changes for any reason the cache lookup will fail.

trwalke:
Fixed in ADAL 5.2.6 (https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/releases/tag/5.2.6)

PR #1688, opened 2020-01-10
Fix for silent broker flow on android

trwalke:
Fix for https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1687


trwalke:
This fix has been E2E tested by the customer

henrik-me:
What tests do we have for validating? Also has it been validated we don't have the same or a similar issue in MSAL?

---
In reply to: [573249367](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1688#issuecomment-573249367) [](ancestors = 573249367)

jennyf19:
Agree with henrik...is there a way to do whatever the internal customer was doing to validate on our end as well and have a test for it?

---
In reply to: [573251223](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1688#issuecomment-573251223) [](ancestors = 573251223,573249367)

jennyf19 (approved):
:shipit:

Issue #1653, opened 2019-09-17
ADAL incorrectly forms AuthorityUri, ADFS returns 404 when authenticating KeyVaultclient


mancyc:
If you think that your issue falls into the above categories, please fill in the form below.

**Which Version of ADAL are you using ?**
ADAL 4.5.1

**Which platform has the issue?**
 net45

**What authentication flow has the issue?**
ADFS

Other? - please describe;
ADAL changed the way they form authority URI, 
 Earlier: string tenant = path.Substring(0, path.IndexOf("/", StringComparison.Ordinal));
 Now: string tenant = authorityUri.Segments[authorityUri.Segments.Length - 1].TrimEnd('/');

This causes a 404 from ADFS when KeyVault client tries to authenticate. KeyVault advertises “/adfs/<tenantId>” in it's WWW-Authenticate header. This works using older version of ADAL.NET (~< 3.5) because it had the behavior to “ignore” additional segments of the URI provided after the first. So KV would provide an authority value of “/adfs/<tenantId>” but ADAL.NET would only take the first section (/adfs) and ignore the rest, thus resulting in a working solution. The later versions of ADAL.NET have changed above logic; it instead takes the “final” segment from the input authority and use that as the tenantId; so in the case of KV, it takes the input of “/adfs/<tenantId>” and trim it down to just “/<tenantId>” whereas ADFS expects /adfs

**Is this a new or existing app?**
a. The app is in production, and I have upgraded to a new version of ADAL

**Expected behavior**
 string tenant = path.Substring(0, path.IndexOf("/", StringComparison.Ordinal));
in Authenticator.cs

**Actual behavior**
404 from ADFS since authorityUri is wrongly formed

**Possible Solution**
 string tenant = path.Substring(0, path.IndexOf("/", StringComparison.Ordinal));

jmprieur:
Thanks @mancyc for this heads-up.

@henrik-me @trwalke @MarkZuber @bgavrilMS @jennyf19 
I'd like to take this one. See also #1410 and #1455

-- reality is stuborn ....

bgavrilMS:
@mancyc - can you give me an example of the ADFS url that KV advertises (not necessarily your own server)

mancyc:
Example of ADFS URL that KV advertises:

[...]

PR #1694, opened 2020-02-12
Jennyf/adfx fix

### LINKED ISSUE #1693: ADAL 5.4.2 Client package is dropping /adfs/ literal from the passed authority URI
**MSAL is the recommended auth library for use with the Microsoft identity platform**

No new features will be implemented on ADAL. The team's efforts are on improving [MSAL](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet), the next-gen auth library. MSAL's wiki contains a [migration guide](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Adal-to-Msal) from ADAL.

Only regressions, high severity issues and security issues will be fixed on ADAL. Other issues are likely to have already been fixed in MSAL. 

If you think that your issue falls into the above categories, please fill in the form below.

Note that to get help, you need to run the latest preview or non-preview version
For MSAL, please log issues to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet

**ADAL 5.2.4**

**net452**

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [X] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Is this a new or existing app?**

**Azure Key vault** - 
The app is in production, and I have upgraded to a new version of ADAL from 4.19.X/4.5.1 to 5.4.2

**Azure Key vault VSTS Repo**

```csharp
var your = (code) => here;
```

**Expected behavior**
ADAL Client library should continue to goto the specified Authority URL rather than dropping literals from it. 

**Actual behavior**
Azure Key Vault is looking to update the ADAL client library nuget primarily from 3.19.X/4.5.1 to 5.4.2. We need this updated version to support our upcoming features. However, while testing we hit an issue with respect to the Get Token API. In case the callback Authority has the literal /adfs/, the ADAL client library for some reason is dropping it. 
Hence, in case the Authority URL is: https://AuthEndpointBase/adfs/TenantId, the ADAL client is dropping this literal and actually sending the request to https://AuthEndpointBase/TenantId. 

**Possible Solution**

**Additional context/ Logs / Screenshots**
Add any other context about the problem here, such as logs and screebshots. Logging is described at https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/wiki/Logging-in-ADAL.Net. Don't post logs containing PII on GitHub!

jennyf19:
#1693

### DIFF
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
@@ -109,13 +109,15 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Instance
                 var authorityUri = new Uri(Authority);
                 var host = authorityUri.Host;
 
-                // The authority could be https://{AzureAD host name}/{tenantid} OR https://{Dsts host name}/dstsv2/{tenantid}
+                // The authority could be https://{AzureAD host name}/{tenantid} OR https://{Dsts host name}/dstsv2/{tenantid} OR
+                // https://{adfs host name}/adfs/{tenantid}
                 // Detecting the tenantId using the last segment of the url
-                string path = authorityUri.AbsolutePath.Substring(1);
-                string tenant = path.Substring(0, path.IndexOf("/", StringComparison.Ordinal));
+                string tenant;                
                
                 if (AuthorityType == AuthorityType.AAD)
                 {
+                    tenant = authorityUri.Segments[authorityUri.Segments.Length - 1].TrimEnd('/');
+
                     var metadata = await ServiceBundle.InstanceDiscovery.GetMetadataEntryAsync(authorityUri, ValidateAuthority, requestContext).ConfigureAwait(false);
                     host = metadata.PreferredNetwork;
                     // All the endpoints will use this updated host, and it affects future network calls, as desired.
@@ -123,6 +125,8 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Instance
                 }
                 else
                 {
+                    string path = authorityUri.AbsolutePath.Substring(1);
+                    tenant = path.Substring(0, path.IndexOf("/", StringComparison.Ordinal));
                     ServiceBundle.InstanceDiscovery.AddMetadataEntry(host);
                 }
 
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
@@ -111,7 +111,9 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Instance
 
                 // The authority could be https://{AzureAD host name}/{tenantid} OR https://{Dsts host name}/dstsv2/{tenantid}
                 // Detecting the tenantId using the last segment of the url
-                string tenant = authorityUri.Segments[authorityUri.Segments.Length - 1].TrimEnd('/');
+                string path = authorityUri.AbsolutePath.Substring(1);
+                string tenant = path.Substring(0, path.IndexOf("/", StringComparison.Ordinal));
+               
                 if (AuthorityType == AuthorityType.AAD)
                 {
                     var metadata = await ServiceBundle.InstanceDiscovery.GetMetadataEntryAsync(authorityUri, ValidateAuthority, requestContext).ConfigureAwait(false);
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
@@ -109,8 +109,7 @@ namespace Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Instance
                 var authorityUri = new Uri(Authority);
                 var host = authorityUri.Host;
 
-                // The authority could be https://{AzureAD host name}/{tenantid} OR https://{Dsts host name}/dstsv2/{tenantid} OR
-                // https://{adfs host name}/adfs/{tenantid}
+                // The authority could be https://{AzureAD host name}/{tenantid} OR https://{Dsts host name}/dstsv2/{tenantid}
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
```diff
@@ -109,18 +109,24 @@ public async Task UpdateFromTemplateAsync(RequestContext requestContext)
                 var authorityUri = new Uri(Authority);
                 var host = authorityUri.Host;
 
-                // The authority could be https://{AzureAD host name}/{tenantid} OR https://{Dsts host name}/dstsv2/{tenantid}
+                // The authority could be https://{AzureAD host name}/{tenantid} OR https://{Dsts host name}/dstsv2/{tenantid} OR
+                // https://{adfs host name}/adfs/{tenantid}
```
henrik-me: >https://{adfs host name}/adfs/{tenantid} [](start = 19, length = 40)

I thought ADFS wouldn't necessarily have a tenantid. we probably have to support both? #Resolved

tests/Test.ADAL.NET.Unit.net45/InstanceDiscoveryTests.cs
```diff
@@ -173,7 +173,7 @@ public async Task TestInstanceDiscovery_WhenAuthorityIsAdfs_ShouldNotDoInstanceD
             using (var httpManager = new MockHttpManager())
             {
                 var serviceBundle = ServiceBundle.CreateWithCustomHttpManager(httpManager);
-                var authenticator = new Authenticator(serviceBundle, "https://login.contoso.com/adfs", false);
+                var authenticator = new Authenticator(serviceBundle, "https://login.contoso.com/adfs/" + AdalTestConstants.SomeTenantId, false);
```
henrik-me: >/ [](start = 100, length = 1)

does this mean our code doesn't work if there is no / at the end?

tests/Test.ADAL.NET.Unit.net45/InstanceDiscoveryTests.cs
```diff
@@ -173,7 +173,7 @@ public async Task TestInstanceDiscovery_WhenAuthorityIsAdfs_ShouldNotDoInstanceD
             using (var httpManager = new MockHttpManager())
             {
                 var serviceBundle = ServiceBundle.CreateWithCustomHttpManager(httpManager);
-                var authenticator = new Authenticator(serviceBundle, "https://login.contoso.com/adfs", false);
```
henrik-me: >https://login.contoso.com/adfs [](start = 70, length = 30)

i believe this should still be possible #Resolved

henrik-me (changes requested):
:clock1:

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
keystroke: ADFS will never have a tenantId. ARM etc. will advertise and use /adfs for the OIDC endpoint which is correct and works. KeyVault has code patterns from AAD where they take a "base" authority URI in their code and add a tenantId as a suffix to that, so they end-up using /adfs/tenantId in the endpoints. They can't fix the code to directly support the ADFS authority URI so the ask is for ADAL to support passing-in the invalid authority URI and correcting it under-the-hood. Note that this same ask will be levied against MSAL as well. It seems there is a lot of this code from Azure Services that isn't going to be precise about the OIDC protocol and how authority URIs work, so it may be reasonable to have ADAL and MSAL "tolerate" such code and internally look for "/adfs" on the first segment and make the necessary transformations or w/e for ADFS. In #1653 we discussed this a bit more and how the MSAL design for .WithAdfsAuthority is also problematic for these same kinds of code patterns. Web applications that integrate with identity need to be built supporting both identity providers, but are we going to make end-users parse the authority URI again themselves in their code and branch to call .WithAuthority or .WithAdfsAuthority? This means ADAL / MSAL internally understanding authority URIs that begin with /adfs and "knowing" that it is an ADFS-backed IDP and whatever that needs to mean to make things work. #Resolved

jennyf19: sounds good. thanks for clarifying. I will remove the {tenantid} here, but it doesn't impact any of the changes.

---
In reply to: [378542753](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1694#discussion_r378542753) [](ancestors = 378542753)

tests/Test.ADAL.NET.Unit.net45/InstanceDiscoveryTests.cs
jennyf19: yes it is. i removed the tenantId. 

---
In reply to: [378037207](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1694#discussion_r378037207) [](ancestors = 378037207)

jennyf19:
ok...addressed the comments.

---
In reply to: [357184636](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1694#issuecomment-357184636) [](ancestors = 357184636)

src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/Instance/Authenticator.cs
henrik-me: nice, as discussed. we should have tests for both scenarios.

jennyf19: yes...will add.

henrik-me (approved):
:shipit:

jennyf19:
PR build passed here: https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=473498&view=results

jkapil:
The dSTS V2 and ADFS have fixed tenant Id. In the context of ADFS tenantId is not adfs , it is a GUID. However when ADAL needs to create URL for getting tokens, it needs to ignore tenantId for ADFS (until ADFS fix comes - which could be never) and hard-code the outbound URL to {host}/adfs. 

Current commit to revert the regression scoped to ADFS scenario will also work.

PR #1730, opened 2021-09-19
Fixed typo in bug_report.md

abatishchev:


jmprieur (approved):
Thanks @abatishchev 
Please note that ADAL is in maintenance mode.

PR #1732, opened 2021-09-19
Added to AuthenticationParameters sync overload CreateFromUnauthorizedResponse()

### LINKED ISSUE #1731: AuthenticationParameters.CreateFromUnauthorizedResponseAsync() is async but doesn't need to

```xml
<PackageReference Include="Microsoft.IdentityModel.Clients.ActiveDirectory" Version="5.2.9" />
```

.NET Standard 2.1, .NET Core 3.1

**What authentication flow has the issue?**

- [x] Other?

**Short description**

`AuthenticationParameters.CreateFromUnauthorizedResponse()` is `async` because it calls `async` helper method but doesn't need to, the same outcome can be achieved without doing that.

**Long description**

Here's the code execution flow:

1. await AuthenticationParameters.CreateFromUnauthorizedResponseAsync(responseMessage); 
2. await OAuthClient.CreateResponseAsync(responseMessage);
3. AuthenticationParameters.CreateFromUnauthorizedResponseCommon()

So the method is marked `async` because it calls `OAuthClient.CreateResponseAsync` which is async to read the response content. Then it passes the resulting `IHttpWebResponse` to the private helper. The issue is that the said helper doesn't use the content so the whole thing can be omited.

Another issue is that there are no tests that would cover the method and/or help to prove the safety of the change.

**Possible Solution**

1. Add sync overload `CreateFromUnauthorizedResponse()` which doesn't call the async helper
2. Mark the async overload as `[Obsolete]`

abatishchev:
Resolves #1731.

Changes:

- Added `CreateFromUnauthorizedResponse()`
- Marked `CreateFromUnauthorizedResponseAsync()` as `[Obsolete]`
- Made `CreateFromUnauthorizedResponseCommon()` to accept `HttpResponseMessage` instead
- Updated the changelog

Pending changes:

- Bump version. I couldn't find the way to do that.
- Update tests. Done in #1733 which needs to go in first.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,9 @@
+Version 5.2.10
+=============
+Enhancements:
+- **Added sync overload for AuthenticationParameters.CreateFromUnauthorizedResponse().**
+- **Marked async overload AuthenticationParameters.CreateFromUnauthorizedResponseAsync() as obsolete.**
+
 Version 5.2.9
 =============
 Enhancements:
--- a/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationParameters.cs
+++ b/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationParameters.cs
@@ -1,4 +1,4 @@
-﻿//----------------------------------------------------------------------
+//----------------------------------------------------------------------
 //
 // Copyright (c) Microsoft Corporation.
 // All rights reserved.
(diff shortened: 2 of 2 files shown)


jmprieur (commented):
Thanks @abatishchev 

ADAL is now in maintenance mode and we don't intend to release a new version. 
See [Update your applications to use Microsoft Authentication Library and Microsoft Graph API](https://techcommunity.microsoft.com/t5/azure-active-directory-identity/update-your-applications-to-use-microsoft-authentication-library/ba-p/1257363)

abatishchev:
Hi Jean-Marc,

> ADAL is now in maintenance mode and we don't intend to release a new version.

Yes, I know. Is this functionality available in MSAL though? When I googled, all samples point to ADAL. And that's what we're referencing in our Resource Provider, this helper is an important piece when calling ARM with a first-party AAD app.

jmprieur:
@abatishchev : yes it's available in MSAL now. See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/WWW-Authenticate-parameters

bgavrilMS:
Not taking it in ADAL.

PR #1734, opened 2022-01-27
Disable ADFSv3 tests.

pmaytak:
Break because the lab API to retrieve the user is wrong.


bgavrilMS (approved):

tests/Test.ADAL.NET.UIAutomation/XamarinAndroidTests.cs
```diff
@@ -100,22 +99,22 @@ public void AcquireTokenInteractiveADFSv4NonFederatedTest()
         /// <summary>
         /// Runs through the standard acquire token flow
         /// </summary>
-        [Test]
+        //[Test] // Disabled, breaks because cannot get a correct test lab user because the lab API query is wrong.
```
bgavrilMS: nit: [Ignore]

PR #1741, opened 2022-09-29
Fix for ADFS 3.0 with net45

billvolt:
Addresses issue 1354 opened 10/25/2018 by rodro75: ADAL 4.3.0 with ADFS 3.0: The server returned an invalid response.

A change first made to ADAL in version 4.3.0 seems to expect that the authorization result should come from the response body.  Previously in 3.19.8, the result was expected in the form of a redirect URL.  The 4.3.0 change breaks the previously working process by discarding the result in the redirect URL and substituting form fields submitted to ADFS, including password, for the authentication result.  

The change submitted here maintains the new code for collecting the result.  If that fails, the old code is used instead.

This was tested against an ADFS server that was migrated from Server 2012 R2 (ADFS 3.0) to Server 2016 without raising the Farm Behavior Level (FBL) from 1 to 3.  I suspect other FBL-1 ADFS servers may also require this fix.


bgavrilMS:
ADAL has been deprecated and we only accept security fixes. Please upgrade ADFS and use MSAL or federate it to AAD.

PR #1743, opened 2022-11-04
Update to uap10.0.17763, fix upgrade issues.

pmaytak:
Uap10 doesn't exist on the current build agent. Workaround is to upgrade to newer uap version. 
Also fixes some breaks due to target upgrade.


jmprieur:
Given we are releasing a new version of ADAL.NET, don't we want to add an obsolete attribute on `AuthenticationContext`?
with an aka.ms link to the migration 3P docs?
https://aka.ms/adal-net-to-msal-net

@bgavrilMS @jennyf19 what do you think?

bgavrilMS (approved):

bgavrilMS:
@jmprieur - we weren't planning on releasing a new ADAL lib. This change is being made because our nightly builds are broken for some time. This is just to allow us to release in case of some emergency. 

Do note that if you try to install ADAL via nuget you see this: 

image

That said, I don't oppose to releasing again with Obsolete attribute.

jmprieur:
Thanks @bgavrilMS
I know.
I thought having warnings in the built code could be a little more incentive

PR #1744, opened 2022-11-07
Bogavril/deprecate resource

bgavrilMS:


src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationParameters.cs
```diff
@@ -60,6 +59,7 @@ public sealed class AuthenticationParameters
         /// <summary>
         /// Gets or sets the identifier of the target resource that is the recipient of the requested token.
         /// </summary>
+        [Obsolete("The client apps must know in advance which resource it requests tokens for.", true)]
```
bgavrilMS: @jmprieur for warning

jennyf19 (approved):
:shipit:

